PR #10, opened 2015-02-11
refactor(pre): use `gitHead` information from registry

### LINKED ISSUE #8: Use `gitHead` information from registry
In the `prepublish` step the git commit ref of the latest version is retrieved by complex git command execution.

However when retrieving the info from the npm registry about the last published version we also get the git ref of the last version published for free by npm :)

This simplifies the command and also removes the need to manually `git fetch --tags` for local preview.

christophwitzko:
This PR closes #8.


christophwitzko:
@boennemann i am on it!

boennemann:
@christophwitzko While I like the code I dislike the structure; there are now three libs that are almost one and the same thing (`latestRelease.js` (if anything this should be `latest-release.js`), `version.js` and `registry.js`)

I'd prefer to have something like `npm-info.js` that provides the info from the registry and handles all errors regarding that in this location.
The info is then just passed onto the type analyzer and `semver.inc` (or `npm version` see #11), which would also make the code sync for both of these tasks.

boennemann:
Thank you so much for your patience @christophwitzko :heartbeat:

christophwitzko:
Thank you! :ok_hand:

PR #36, opened 2015-06-01
Print a nicer error message, when there is no repo URL defined in package.json

janl:


src/setup.js
```diff
@@ -26,6 +26,9 @@ module.exports = function () {
   // set up repository
   if (!pkg.repository || !pkg.repository.url) {
     var config = ini.decode(fs.readFileSync('./.git/config') + '')
+    if (!config['remote "origin"'] || config['remote "origin"'].url) {
+      throw new Error('your package.json does not define a repo URL');
```
boennemann:
```
standard: Use JavaScript Standard Style (https://github.com/feross/standard)
  /home/travis/build/boennemann/semantic-release/src/setup.js:30:70: Extra semicolon.
```
;)

janl: fixed in latest

boennemann:
Thank you @janl. This will be addressed by future versions of 4.x

PR #39, opened 2015-06-10
refactor: use babel and add coverage reports

boennemann:


christophwitzko:
LGTM :+1:

src/lib/commits.js
```diff
@@ -0,0 +1,23 @@
+import { exec } from 'child_process'
+
+import { efh } from './error'
```
rmehner: FYI: In my recent Babel project I got back to use the regular `require` again, since module loading was recently kicked out of the spec (https://github.com/lukehoban/es6features/issues/75#issuecomment-97763850) and I wanted to keep the code as standard compatible as possible.

boennemann: Oh nice to know thanks. I'll backport this then after I'm done with the tests.
Argh this PR is so huge already :grimacing:

PR #53, opened 2015-08-06
set npm registry from publishConfig.registry if existent

gr2m:
I've a setup with npm enterprise. `semantic-release` currently sets the registry to the default, as I have a registry set for a scope only. This is what my .npmrc file looks like:

```
@myscope:registry=https://myregistry.example.com/
```

So I think we should do both of these things:

1. check if the current package as a scope. If it does, check for `conf.get('@scope/registry')`
2. If `pkg.publishConfig.registry` and use that if it exists.

My change fixed the issue for me. Happy to help with the tests or anything else, but I'll need guidance

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -42,7 +42,7 @@ npmconf.load({}, (err, conf) => {
       token: env.NPM_TOKEN
     },
     loglevel: conf.get('loglevel'),
-    registry: conf.get('registry'),
+    registry: (pkg.publishConfig && pkg.publishConfig.registry) || conf.get('registry'),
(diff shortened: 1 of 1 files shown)


gr2m:
discussed this quickly with @boennemann. Plan is as follows

1. Make a `lib/get-registry` file
2. If package.json name has a scope, look for `conf.get('@'+scopeName+'/registry')`
3. If it returns undefined, check if pkg.publishConfig.registry is set
3. If that returns undefined, use `conf.get('registry')`

PR #58, opened 2015-08-18
Document the getLastRelease plugin

tlvince:


boennemann:
Awesome! Noticed that I forgot it and now I've got the PR in before I could fix it :sparkling_heart:

PR #60, opened 2015-08-19
feat(commit-analyzer): Change to new commit-analyzer API

### LINKED ISSUE #41: add `[release skip]` commit message command

ariporad:
Support the new commit-analyzer API. With the new API, we call the plugin once
for each commit, and keep track of the changes ourselfs.
This also adds support for the 'rskip' release type, which skips the release.

This revealed a major problem with the pre.js unit tests (test/specs/pre.js),
child process was not actually being mocked, and was simply returning an empty
array. With the old system, the mocked commit-analyzer was still called and
could still trigger a release (this is actually a flaw with the whole system.
With the new system, this is no longer true, so the unit tests fail. I did not
fix this both because it's not related to this branch/pull request, and because
it would mean a major revamp, which I would need to think about more. (this
means that travis will fail).

Fixes #41
Refs semantic-release/commit-analyzer#1, semantic-release/commit-analyzer#2
BREAKING CHANGE:
	The commit-analyzer is now called multiple times, and with a single commit
	(passed in as `commit`, vs `commits` as an array). The plugin also no longer
	has to deal with keeping track of the release type between commits.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "@semantic-release/error": "^1.0.0",
     "@semantic-release/last-release-npm": "^1.1.1",
     "@semantic-release/release-notes-generator": "^2.0.0",
+    "async": "^1.4.2",
     "git-head": "^1.2.1",
     "github": "^0.2.4",
     "lodash": "^3.9.3",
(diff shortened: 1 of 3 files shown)


boennemann:
Sorry this was auto closed because I renamed the default branch, can you send this again please?

ariporad:
Apparently not: http://ariporad.link/1hCvf1H (screenshot). Can you just reopen this?

PR #62, opened 2015-08-20
docs(readme): Update semantic-release setup cmd

kentcdodds:
The image needs to be updated as well because that's not what happens anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ _Note:_ The current release/tag implementation is tied to GitHub, but could be o
 npm install -g semantic-release-cli
 
 cd your-module
-semantic-release-cli setup
+semantic-release setup
 ```
 
 ![dialogue](https://cloud.githubusercontent.com/assets/908178/8766357/f3eadaca-2e34-11e5-8ebb-d40b9ae613d7.png)


boennemann:
The cli exposes itself as `semantic-release-cli`, to avoid confusion with the semantic-release core package, which is available as `semantic-release`. Unless I'm missing something the docs are correct the way they are :) But you're right new CLI images needs an update, w/o the info logs, and w/o passwords :)

kentcdodds:
When I `npm install -g semantic-release-cli` the command I then execute is `semantic-release setup` not `semantic-release-cli setup`. `semantic-release-cli` command is not found.

boennemann:
The semantic-release core used to have a setup command in older versions. I seems like you're running this old version globally installed. We've split that out, to keep the core as small as possible.

The CLI definitely exposes itself as `semantic-release-cli` https://github.com/semantic-release/cli/blob/master/package.json#L6
I'm confused as to why `npm install -g semantic-release-cli` doesn't give you that executable.

Can you run `which semantic-release`, or uninstall `semantic-release` globally and then only install `npm i -g semantic-release-cli` again. Also what's `npm -v`?

Thanks for your help!

kentcdodds:
Huh, that was odd. It's working now:

screen shot 2015-08-20 at 8 09 43 am

boennemann:
@kentcdodds Thanks for clarifying. To me this looks like the first time you didn't `npm install -g semantic-release-cli`, but `npm install -g semantic-release`. You then ended up with the "old" semantic-release as a global install, which also has a setup command. I say "old", but it's still the "latest" semantic-release, because I want to address some more things in the current "next" branch before moving them over.

Glad it's working now :+1:

boennemann:
Just to document this here: 4.x is now on the "latest" dist-tag, which means this should no longer happen in the future :)

PR #80, opened 2015-09-06
feat(post): add prerelease option for GitHub releases

eliias:
Tag a release as "pre-release" on GitHub releases

### DIFF
--- a/src/post.js
+++ b/src/post.js
@@ -29,6 +29,7 @@ module.exports = function (config, cb) {
         tag_name: `v${pkg.version}`,
         target_commitish: hash,
         draft: !!options.debug,
+        prerelease: !!options.prerelease,
         body: log
       }
 


boennemann:
Hey @eliias,

very neat solution. I just want to make sure the name isn't confusing before merging this in.
Is it obvious enough that the global "prelease" option only affects the GitHub Release part?

Also as we were thinking of making the whole github-release part a plugin anyways would it make sense to create this plugin step now and pass "prerelease" to that plugin instead.

Best,
Stephan

eliias:
Yep. A plugin does make total (more) sense than adding arbitrary options for an unspecified repository host. I was not ware of any github-release plugin plans. Beside that, it would allow to add all the other release options to the plugin too. Ping me if you need a hand.

PR #111, opened 2015-10-08
feat(cli): added a next-version command to CLI

olalonde:
- Moved index.js to cli.js
- Created a new index.js that exposes library functions
- Moved configuration initialization logic to its own file
- Added a new 'next-version' command that returns the result of the
  pre() function. This could be used in other scripts. e.g.:

  `npm version $(semantic-release next-version --type)`
  `semantic-release next-version)`
  `semantic-release next-version --type)`
  `semantic-release next-version --version)`


olalonde:
Any chance to get feedback on this PR?

olalonde:
The reason I need this is I only want to use the part of semantic-release which guesses what should be the next package version as part of a different publish flow (without all the other stuff).

boennemann:
Hey @olalonde,

thanks for taking the time to send a pull request and sorry for taking so long to answer.

The philosophy of this project is to **not** bother anyone with versions or types, which is why we haven't and won't expose them.

I see that you want to reuse this with another script and don't want to use the rest of semantic-release. I'm happy to tell you that there is dedicated module that does exactly what you want: [conventional-recommended-bump](https://github.com/stevemao/conventional-recommended-bump)

Thanks a lot for your patience and time. If you have ideas for improvement in the future please reach out in the Gitter chat first, so we can avoid unnecessary work.

Best,
Stephan

olalonde:
Oh great, thanks. Wish I had found out about this earlier!

PR #148, opened 2015-12-19
fix(commits.js): Allow for long commit logs

wcandillon:


hbetts:
I don't really understand why this is failing. The `exec` method takes an options object, and the transpiled code looks fine.

Issue #114, opened 2015-10-12
An error on CI install


mrself:
I have such an error when setup cli:
```
semantic-release-cli setup
? Are you using GitHub Enterprise? No
? What is your npm registry? https://registry.npmjs.org/
? What is your npm username? mrself
? What is your npm email? yaruk.my@gmail.com
? What is your GitHub username? mrself
? What CI are you using? Travis CI
C:\Users\Lightman\AppData\Roaming\npm\node_modules\semantic-release-cli\dist\lib\travis.js:113
    travis.repos(info.ghrepo.slug[0], info.ghrepo.slug[1]).get(function (err,
                            ^
TypeError: Cannot read property 'slug' of undefined
    at C:\Users\Lightman\AppData\Roaming\npm\node_modules\semantic-release-cli\dist\lib\travis.js:113:29
    at C:\Users\Lightman\AppData\Roaming\npm\node_modules\semantic-release-cli\dist\lib\travis.js:50:5
    at C:\Users\Lightman\AppData\Roaming\npm\node_modules\semantic-release-cli\node_modules\travis-ci\lib\travis-ci.js:51:21
    at Request._callback (C:\Users\Lightman\AppData\Roaming\npm\node_modules\semantic-release-cli\node_modules\travis-ci\lib\travis-http.js:51:13)
    at Request.self.callback (C:\Users\Lightman\AppData\Roaming\npm\node_modules\semantic-release-cli\node_modules\travis-ci\node_modules\request\index.js:148:22)
    at Request.emit (events.js:110:17)
    at Request.<anonymous> (C:\Users\Lightman\AppData\Roaming\npm\node_modules\semantic-release-cli\node_modules\travis-ci\node_modules\request\index.js:886:14)
    at Request.emit (events.js:129:20)
    at IncomingMessage.<anonymous> (C:\Users\Lightman\AppData\Roaming\npm\node_modules\semantic-release-cli\node_modules\travis-ci\node_modules\request\index.js:837:12)

    at IncomingMessage.emit (events.js:129:20)
```

mrself:
My github repo was `git+git@github.com:mrself/` instead of `git+https://github.com/mrself/`. I think I did not notice it when run `npm init`

bahmutov:
Same problem, the repo in package.json
```
"repository": {
     "type": "git",
     "url": "git@github.com:bahmutov/obind.git"
}
```

bahmutov:
[...]

PR #77, opened 2015-08-31
gitlab support

johann-sonntagbauer:
Provide support for gitlab based repositories. To enable the support the github based verification checks are refactord into own verificationConditions plugin. Also the plugin mechanism is modified a little so that all plugins can be called with Array notation.

The configuration for gitlab scenarios is as follows:
package.json

``` javascript
{
 "release": {
    "debug": false,
    "verifyConditions": "semantic-release/dist/lib/plugin-noop"
  },
  scripts: {
    "semantic-release": "semantic-release pre && npm publish"
  }
}
```

PR #166, opened 2016-01-07
feat(gitlab): added support for parsing hosted gitlab repo urls

bahmutov:
Replaced parse github url with updated version with more git cases supported, including gitlab domains.
Might be a good complement when/before/after https://github.com/semantic-release/semantic-release/pull/77 lands

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "path": "node_modules/cz-conventional-changelog/"
   },
   "dependencies": {
+    "@bahmutov/parse-github-repo-url": "^0.1.0",
     "@semantic-release/commit-analyzer": "^2.0.0",
     "@semantic-release/condition-travis": "^5.0.0",
     "@semantic-release/error": "^1.0.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -13,6 +13,7 @@
     "path": "node_modules/cz-conventional-changelog/"
   },
   "dependencies": {
+    "@bahmutov/parse-github-repo-url": "^0.1.0",
```
hbetts: Perhaps a silly question, but why another parse library?

Just searching Google, there are quite a few out there.

Also, if it's not github specific (your README mentions GItLab), why not remove the `github` reference from the package name?

bahmutov: The idea was that library was already used, just not maintained, so it was missing a few cases. I just kept adding use cases, so it is pretty simple replacement.

boennemann:
Thanks for this @bahmutov. Just to clarify, merging this would also fix #114, right?

bahmutov:
Yes forgot about that open issue, should fix it

package.json
hbetts: Thanks @bahmutov

hbetts:
:+1:

boennemann:
Thanks @bahmutov! Landed here https://github.com/semantic-release/semantic-release/commit/66454e7e46d32b00bb597a4eeac05036cad8dda8

PR #173, opened 2016-01-19
Make config verification a plugin

ArnaudRinquin:
At the moment, `semantic-release` is very package oriented. Using it for end-of-the-chain app release, which don't get released on npm, isn't really possible.

In our scenario, we just run `semantic-release pre && npm semantic-release post` in order to tag and create a release Github release of the app.

The main blocker is the hard check on npm token config.

While this check is valuable when actually running `npm publish`, it's useless if we omit that step.

That PR makes it so what was in `verify.js` is now used as the `verifyConfig` plugin. I had to make asynchronous and respect regular plugin signature.

That will allow use to write our custom `verifyConfig` that doesn't check for `npm token`.

I hope this is clear enough, please shoot questions!

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,6 +212,11 @@ This plugin is responsible for determining the type of the next release. It addi
 
 This plugin is responsible for generating release notes. Call the callback with the notes as a string. Have a look at the [default implementation](https://github.com/semantic-release/release-notes-generator/).
 
+### `verifyConfig`
+
+This plugins is responsible for verifying that the environment contains all the necessary configuration. By default it checks that `package.json` contains a `name` and `repository`. It also checks that
+a GitHub token is provided as weel as a npm one.
+
 ### `verifyConditions`
 
 This plugins is responsible for verifying that a release should happen in the first place. For example, the [default implementation](https://github.com/semantic-release/condition-travis/) verifies that the publish is happening on Travis, that it’s the right branch, and that all other build jobs succeeded. There are more use cases for this, e.g. verifying that test coverage is above a certain threshold or that there are no [vulnerabilities](https://nodesecurity.io/) in your dependencies. Be creative.
--- a/bin/semantic-release.js
+++ b/bin/semantic-release.js
@@ -22,6 +22,7 @@ var knownOptions = {
   'analyze-commits': [path, String],
   'generate-notes': [path, String],
   'verify-conditions': [path, String],
+  'verify-config': [path, String],
   'verify-release': [path, String]
 }
 var options = _.defaults(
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -212,6 +212,11 @@ This plugin is responsible for determining the type of the next release. It addi
 
 This plugin is responsible for generating release notes. Call the callback with the notes as a string. Have a look at the [default implementation](https://github.com/semantic-release/release-notes-generator/).
 
+### `verifyConfig`
+
+This plugins is responsible for verifying that the environment contains all the necessary configuration. By default it checks that `package.json` contains a `name` and `repository`. It also checks that
+a GitHub token is provided as weel as a npm one.
```
hbetts: `weel` => `well`

bin/semantic-release.js
```diff
@@ -75,89 +76,91 @@ npmconf.load({}, function (err, conf) {
...
-          } catch (e) {
-            log.silly('pre', 'Couldn\'t find npm-shrinkwrap.json.')
-          }
-
-          fs.writeFileSync('./package.json', JSON.stringify(_.assign(pkg, {
-            version: release.version
-          }), null, 2))
-
-          log.verbose('pre', 'Wrote version ' + release.version + ' to package.json.')
+  plugins.verifyConfig(config, function (errors) {
```
hbetts: Can you this be broken out into a separate file, like `src/lib/plugin-verify-npm.js`.

Actually, that brings up another point. Perhaps this needs to be a very specific type of verify config plugin. One that only checks that npm is setup correctly in the environment.

hbetts: Actually I'm not sure I understand this code at all. The `verifyConditions` plugin only gets called if `verifyConfig` passes?

What if I want to ignore the npm token verification, but want to keep the GItHub verification and the `verifyConditions` plugin? Will I need to re-write the `verifyConfig` plugin, minus the `npm` check code?

test/specs/plugins.js
```diff
@@ -3,7 +3,7 @@ var test = require('tap').test
 var plugins = require('../../src/lib/plugins')
 
 test('export plugins', function (t) {
-  t.plan(5)
+  t.plan(6)
```
hbetts: Because this test block does not run multiple actions asynchronously, I would just remove the `t.plan` call, and call `t.end()` after line 15. Much less flaky (As you don't have to remember to increment the plan count each time you add a test.

src/lib/plugins.js
```diff
@@ -5,7 +5,8 @@ var exports = module.exports = function (options) {
   var plugins = {
     analyzeCommits: exports.normalize(options.analyzeCommits, '@semantic-release/commit-analyzer'),
     generateNotes: exports.normalize(options.generateNotes, '@semantic-release/release-notes-generator'),
-    getLastRelease: exports.normalize(options.getLastRelease, '@semantic-release/last-release-npm')
+    getLastRelease: exports.normalize(options.getLastRelease, '@semantic-release/last-release-npm'),
+    verifyConfig: exports.normalize(options.verifyConfig, './verify.js')
```
hbetts: Since verify is now a plugin, I would probably move it into the `lib` directory and prepend the name with `plugin-`.

hbetts:
@ArnaudRinquin you may also want to check out #77.

mdreizin:
@boennemann Did you have a chance to look at this PR?

hbetts:
As this pull request has been sitting here for a while, I'm going to go ahead and close it out.

ArnaudRinquin:
@mdreizin I'm not working on it anymore really.

pvdlg:
This should be solved by #515 which we just released in `semantic-release@10.0.0` on the `next` dist-tag.
The verifications are now part of https://github.com/semantic-release/github and https://github.com/semantic-release/npm.

PR #177, opened 2016-01-23
refactor(pre): Use npm.config instead of npmconf package.

### LINKED ISSUE #176: deprecated npmconf?
during `npm install` of a package which with semantic-release in its dependency, I see this warning:

```
WARN deprecated npmconf@2.1.2: this package has been reintegrated into npm and is now out of date with respect to npm
```

not sure if there is anything you can do about it, I find the deprecation rather confusing

hbetts:
Closes #176

### DIFF
--- a/bin/semantic-release.js
+++ b/bin/semantic-release.js
@@ -7,7 +7,7 @@ var url = require('url')
 var _ = require('lodash')
 var log = require('npmlog')
 var nopt = require('nopt')
-var npmconf = require('npmconf')
+var npm = require('npm')
 var normalizeData = require('normalize-package-data')
 
 log.heading = 'semantic-release'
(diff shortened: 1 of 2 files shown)


bin/semantic-release.js
```diff
@@ -41,32 +41,32 @@ var options = _.defaults(
...
 
-  var npm = {
+  var npmConf = {
     auth: {
       token: env.NPM_TOKEN
     },
-    loglevel: conf.get('loglevel'),
-    registry: require('../src/lib/get-registry')(pkg, conf),
-    tag: (pkg.publishConfig || {}).tag || conf.get('tag') || 'latest'
+    loglevel: npm.config.get('loglevel'),
```
hbetts: Fun fact: Calling `npm.get` will return the config value, but it also console logs the value (which creates a lot of noise on the console). However, `npm.config.get` does not console log the value.

bin/semantic-release.js
```diff
@@ -91,59 +91,55 @@ npmconf.load({}, function (err, conf) {
...
-      var nerfDart = require('nerf-dart')(npm.registry)
+      var nerfDart = require('nerf-dart')(npmConf.registry)
       var wroteNpmRc = false
 
       if (env.NPM_OLD_TOKEN && env.NPM_EMAIL) {
         // Using the old auth token format is not considered part of the public API
         // This might go away anytime (i.e. once we have a better testing strategy)
-        conf.set('_auth', '${NPM_OLD_TOKEN}', 'project')
-        conf.set('email', '${NPM_EMAIL}', 'project')
+        npm.set('_auth', '${NPM_OLD_TOKEN}', 'project')
```
hbetts: Another fun fact: `npm.config.set`, does not write the config value to disk. It only writes it into memory within the context of the `npm.config` instance. However, `npm.set` will write the value to the user's `.npmrc` file.

boennemann:
Thanks @hbetts,

do you know what exactly changed since npmconf was merged back into npm?
I know it's not too cool that npmconf is deprecated, but the programmatic npm API is unsupported as well.

This means on top of requiring the entire npm client, we would have to pin it down and review every version update – and they happen about every week.

I'm a bit puzzled, and I don't know which of theses two unsatisfying ways is the better one :/

Best,
Stephan

hbetts:
@boennemann I don't know what's changed since npm switched to handling config within their own package. There's very little documentation for npm's programmatic config interface so I don't have anything to compare against. Some of their documentation is actually incorrect (such as console logging for npm.config.get)

I concur that requiring `npm` is about as bad as it goes for refactoring. It's a massive dep that, as you pointed out, changes quite often, and in my opinion, not always in a stable manner.

Though, I'm probably not _as_ worried about pinning `npm` as perhaps you are.

hbetts:
As this pull request has been sitting here for a while, I'm going to go ahead and close it out.

jhabdas:
@hbetts I don't know what's going on with this tool. Has it been totally abandoned. Is there a fork I should be using? Seeing stuff like this makes me uneasy semantic release is a part of my deployment process.

hbetts:
For anyone that comes across this closed pull request, you may be interested in @gr2m's [comment](https://github.com/semantic-release/semantic-release/issues/176#issuecomment-304652381).

jhabdas:
@hbetts thanks. I see I went around in circles on this a couple months back. Earlier comment redacted.

PR #204, opened 2016-03-17
fix: only ever change version field in published package.json

### LINKED ISSUE #203: No README data found! on npm
I was just looking into why https://www.npmjs.com/package/@hoodie/log has a `No README data found!` message on npm after being published with `semantic-release`

I found out that  [normalize-package-data](https://github.com/npm/normalize-package-data) is adding `"readme": "ERROR: No README data found!"`  to the `pkg`  on this line: https://github.com/semantic-release/semantic-release/blob/caribou/bin/semantic-release.js#L16

A fix would be to just remove `normalize-package-data`. It doesn't seem to me like it's needed there.

boennemann:
Closes #203 

@finnp


finnp:
Awesome, LGTM :+1:

PR #209, opened 2016-03-30
docs(README.md): discourage use of pre-patch, pre-minor and pre-release

ryan-kimber:
In my initial use of semantic-release, I spent a considerable amount of time mucking about and trying to determine how to best make use of 'pre' tags, only to later realize that I shouldn't be trying to use them at all (and that dist-tags are a better fit for a CI/CD process).

I hope that, by explicitly discouraging their use, future users can avoid accidentally going down the same path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,7 +206,11 @@ module.exports = function (pluginConfig, config, callback) {}
 
 ### `analyzeCommits`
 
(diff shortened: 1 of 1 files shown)


christophwitzko:
Hi @ryan-kimber, thank you for your contribution. I have rebased the commit here: https://github.com/semantic-release/semantic-release/commit/6a172e172035432b2e063fad3d088beb46332c92

PR #227, opened 2016-05-23
Expose cafile from npm config

aquavitae:
Exposing this allows semantic-release to be used in environments with self-signed certificates.  See also: https://github.com/semantic-release/last-release-npm/pull/75


boennemann:
Thank you! Looks good.

PR #331, opened 2016-11-21
fix(commits): Use `--no-color` for `git branch`-command

PAkerstrand:
Running `semantic-release` fails locally if/when `git branch` produces colored output. This PR simply adds `--no-color` to the executed command. With this fix I get the next version correctly, while previously I would get an `ENOTINHISTORY`-error, since the check does not anticipate color codes in the branch names.

| Before | After |
| ------ | ----- |
| 1____project_stylizr__bash_ | 1____project_stylizr__bash_ |

### DIFF
--- a/src/lib/commits.js
+++ b/src/lib/commits.js
@@ -13,7 +13,7 @@ module.exports = function (config, cb) {
 
   if (!from) return extract()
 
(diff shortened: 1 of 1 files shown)


boennemann:
Great catch, thank you!

Issue #286, opened 2016-08-30
maxBuffer exceeded in git log command


joscha:
When running this in a repository with a lot of commits and long commit messages, e.g. https://bitbucket.org/atlassian/atlaskit/ 

`src/lib/commits.js` errors with:

```console
{ Error: stdout maxBuffer exceeded
    at Socket.<anonymous> (child_process.js:255:14)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at readableAddChunk (_stream_readable.js:177:18)
    at Socket.Readable.push (_stream_readable.js:135:10)
    at Pipe.onread (net.js:542:20)
  cmd: 'git log -E --format=%H==SPLIT==%B==END== bf9db6394efa66bb3afa71609511d4a1e70824b5..HEAD' }
```

(bf9db6394efa66bb3afa71609511d4a1e70824b5 is from https://bitbucket.org/atlassian/atlaskit/)

This is due to the fact that exec by default does not pipe, but buffer and the default buffer is 200KB (https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback)

The output produced by the command above is roughly 320KB

Increasing the buffer size seems the obvious choice for a fast fix, however this problem will potentially come up again for someone with more commits/longer messages.

bengummer:
I've raised a PR to solve this using `child_process.spawn()` as we're hitting the `maxBuffer` error again even with the limit set to 1MB.

Can this issue be re-opened @boennemann ?

PR #332, opened 2016-11-24
fix: use spawn instead of exec for 'git log' to avoid maxBuffer err

bengummer:
Related to https://github.com/semantic-release/semantic-release/issues/286

Uses `child_process.spawn` instead of `child_process.exec`, as `exec` has a fixed `maxBuffer`.

Only converted `git log` to use the new `spawn` method to reduce PR complexity and `git log` seems like the main culprit for `maxBuffer` errors.


boennemann:
Thank you 🎉

PR #348, opened 2016-12-23
Fix 'home' link in package.json

swashcap:
I noticed your (hilarious) branch names! And also that the `home` link is old and points to a non-existent branch. ✨


boennemann:
Thank you!

PR #368, opened 2017-03-06
Add create-npmrc option

nanovazquez:
While trying to integrate `semantic-release` with [Jenkins Multibranch Pipeline](https://jenkins.io/blog/2015/12/03/pipeline-as-code-with-multibranch-workflows-in-jenkins/), we noticed that the script always tries to create the **.npmrc** file before publishing. In our scenario, we already have a **.npmrc** file in place since we publish our libraries to our internal npm private registry (and thus we need this file to run `npm install`). The current implementation is trying to overwrite this file, something we want to avoid.

This PR adds a new option for `semantic-release` to skip the creation of the **.npmrc** file during the `pre` command. 

## Usage

***via cli***
```bash
semantic-release pre --no-create-npmrc && npm publish && semantic-release post
```

**via package.json**
```js
"release": {
  "createNpmrc": false,
  ...
}
```

## Implementation details

* Added new `options.createNpmrc` config, set to `true` by default (see [here](https://github.com/nanovazquez/semantic-release/blob/7b025f6993186b3277cac53438f47eed1faf210a/bin/semantic-release.js#L21)).
* If this value is explicitly set to `false`, we skip the creation of the **.npmrc** file (see [here](https://github.com/nanovazquez/semantic-release/blob/7b025f6993186b3277cac53438f47eed1faf210a/bin/semantic-release.js#L102)).
* A [lib/define-version.js](https://github.com/nanovazquez/semantic-release/blob/7b025f6993186b3277cac53438f47eed1faf210a/src/lib/define-version.js) file was created to extract the logic that defines the version to publish. 
It was copied _as is_ to avoid extra noise in this PR, but there are some improvement opportunities that could be considered that would also require a bigger refactor of the original code (like moving all the logging logic back to the [bin/semantic-release.js](https://github.com/nanovazquez/semantic-release/blob/7b025f6993186b3277cac53438f47eed1faf210a/bin/semantic-release.js) file).
* Updated [verify.js](https://github.com/nanovazquez/semantic-release/blob/7b025f6993186b3277cac53438f47eed1faf210a/src/lib/verify.js) to skip the validation of the **NPM_TOKEN** environment variable if `options.createNpmrc` is set to `false` (it's not needed in the flow).

### DIFF
--- a/bin/semantic-release.js
+++ b/bin/semantic-release.js
@@ -18,6 +18,7 @@ normalizeData(pkg)
 var knownOptions = {
   branch: String,
   debug: Boolean,
+  'create-npmrc': Boolean,
   'github-token': String,
   'github-url': String,
   'analyze-commits': [path, String],
@@ -36,6 +37,7 @@ var options = _.defaults(
       next: 'latest'
     },
     debug: !env.CI,
+    createNpmrc: true,
     githubToken: env.GH_TOKEN || env.GITHUB_TOKEN,
     githubUrl: env.GH_URL
   }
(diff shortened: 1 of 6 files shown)


durancristhian:
I'd love this option for the same use case 💪 

Hope it gets merged soon!

bin/semantic-release.js
```diff
@@ -96,57 +98,30 @@ npmconf.load({}, function (err, conf) {
...
+        conf.save('project', function (err) {
+          if (err) return log.error('pre', 'Failed to save npm config.', err)
+
+          if (wroteNpmRc) log.verbose('pre', 'Wrote authToken to .npmrc.')
+
+          defineVersion(config, options, originalPkg, npm)
         })
-      })
+      } else {
+        log.verbose('pre', 'Skipped creation of .npmrc.')
```
gr2m: can you make this a `log.info` please?

gr2m:
@nanovazquez @durancristhian Would it be an option to set your custom registry in the `package.json` file? You can set `"registry"` in the `"publishConfig"` setting, see https://docs.npmjs.com/misc/registry#i-dont-want-my-package-published-in-the-official-registry-its-private

bin/semantic-release.js
nanovazquez: Sure. I'll make the change.

nanovazquez:
> _ @gr2m said:_ @nanovazquez @durancristhian Would it be an option to set your custom registry in the package.json file? You can set "registry" in the "publishConfig" setting.

We are currently doing it. And that's why we need an `.npmrc` in place. The problem is that we already have one in place, and we want to use that one (not the one `semantic-release` tries to create by default).

gr2m:
I’m still finding my ways around semantic-release & npm details myself. I would prefer to find a way to make semantic-release work without adding another option. 

Can you tell me how your existing `.npmrc` file looks like and how the one created by semantic-release looks like? Maybe we can merge the two if one already exists so it would work for you?

nanovazquez:
> _Can you tell me how your existing .npmrc file looks like and how the one created by semantic-release looks like? Maybe we can merge the two if one already exists so it would work for you?_

**.npmrc file**
```
@<private-scope>:registry=<registry-url>
//<registry-url>/:_authToken=<registry-token>
```
@gr2m: The biggest problem is that `semantic-release` uses the deprecated [npmconf](https://github.com/npm/npmconf) library. And, when this tool tries to save the **.npmrc** at the project's level (not sure why, because we already have one inside the user's folder), the process fails because the [Jenkins Multibranch Pipeline](https://jenkins.io/blog/2015/12/03/pipeline-as-code-with-multibranch-workflows-in-jenkins/) doesn't allow it.

From my point of view, the problem is that `semantic-release` **always** tries to create an `.npmrc` file, which is not required in some cases (this one, for instance). I can try to refactor the code to avoid creating this file when is not needed. But that might the replacement of the [npmconf](https://github.com/npm/npmconf) tool and/or big structural changes. With this PR, we are avoiding the refactor and we maintain backwards compatibility. Thoughts?

bin/semantic-release.js
```diff
@@ -96,57 +98,30 @@ npmconf.load({}, function (err, conf) {
...
-            log.silly('pre', 'Couldn\'t find npm-shrinkwrap.json.')
-          }
-
-          fs.writeFileSync('./package.json', JSON.stringify(_.assign(originalPkg, {
-            version: release.version
-          }), null, 2))
-
-          log.verbose('pre', 'Wrote version ' + release.version + ' to package.json.')
+      var defineVersion = require('../src/lib/define-version')
+      if (options.createNpmrc) {
```
gr2m: would it  be enough to check here if `.npmrc` file exists? And if it does, skip the code below? That way we don’t need to introduce another configuration. But I’m not 100% sure that this wouldn’t break someone else’s setup

nanovazquez:
It will be enough, but as you said, we can't guarantee that we are introducing a breaking change in someone else's setup. 

That's why I suggested the use of a configuration value. But, if you agree, I can update the code, introduce a breaking change and then release it as a major.

gr2m:
how about we check if the `.npmrc` file exists. If it does, we read out its contents and add it back in after `semantic-release` is done overriding it. That should work and not break anything? I guess your file with have the URL of your private registry used for a specific `@scope` and the auth token for it?

nanovazquez:
@gr2m: I'm not sure if that will work because semantic release always tries to save the .npmrc file at the project level, and in this case the .npmrc is stored in the user's folder. The user performing the action might not have access to even create a file inside the project's folder (this particular case). And for sure, it probably won't have access to create a file inside the user's folder.

gr2m:
Okay thanks for all your insights on this. I'll reach out to some friends who know much more about npm internals. I’ll get back to you :)

pvdlg:
I'm working on another PR and I will probably remove the deprecated npmconf.
Would that work if we create the `.npmrc` file is it doesn't exists and append to it if it does exist ?
You mentioned "the process fails because the Jenkins Multibranch Pipeline doesn't allow it", can you explain a bit more in detail what the problem exactly is? Is it Jenkins that disallow to modify the `.npmrc` ?

pvdlg:
We released `semantic-release@10.0.0` on the `next` dist-tag that uses the plugin https://github.com/semantic-release/npm.
We now update the `.npmrc` only if it doesn't contain an authentication setting for the configured registry. Also we append to the file, therefore we don't erase any configuration that might be there already.

Feel free to re-open if that doesn't completely solve your issue.

PR #387, opened 2017-05-11
chore(package): add `preferGlobal` key

luftywiranda13:
If we prefer this package to be installed globally,
then we better add the key

https://docs.npmjs.com/files/package.json#preferglobal


luftywiranda13:
screen shot 2017-05-11 at 07 41 02

it'll make it look like that in npmjs

Issue #256, opened 2016-07-09
Bump minor version when not needed


tleunen:
I'm seeing the issue for a while now on a specific project, here: [tleunen/eslint-import-resolver-babel-module-alias](https://github.com/tleunen/eslint-import-resolver-babel-module-alias)

Any reason why it does that?

```
> eslint-import-resolver-babel-module-alias@0.0.0-semantically-released semantic-release /home/ubuntu/eslint-import-resolver-babel-module-alias
> semantic-release pre && npm publish && semantic-release post

+ eslint-import-resolver-babel-module-alias@1.4.0
Generating changelog from v1.3.0 to HEAD...
Parsed 0 commits.
```

motiz88:
Seeing the same behavior here: [motiz88/astring-flow](https://github.com/motiz88/astring-flow). I'm also using CircleCI, maybe that's a factor? I have several other modules building on Travis with semantic-release, and that has always worked fine.

motiz88:
**tl;dr: try running the attached `git`/`bash` one-liner prior to `npm publish`**

Got it! `semantic-release` is not the culprit (though it could possibly help - see 1 below). This happens when the "latest" version on npm is missing the `gitHead` field of `package.json`. In this case, `semantic-release` analyzes the _entire commit history_ each time, generating spurious version bumps.

This arises from in `npm publish`'s ~~brittle~~ _broken_ way of getting at that Git hash in the first place (PR [here](https://github.com/npm/read-package-json/pull/67)): It silently fails on perfectly valid Git repos if they use [packed refs](https://git-scm.com/docs/git-pack-refs) a certain way.

Curiously, _that_ somehow happens on CircleCI and not on, say, Travis (last time I checked). Haven't looked into why that is, since again, the situation is perfectly valid from a `git` standpoint - It's `npm` that's depending on an internal detail and thus breaking.

Now, the workaround: I added the following Bash command to `circle.yml` right before `npm run semantic-release`. It reconstructs the "unpacked refs" for all the branch heads that exist only as packed refs, so `npm` can read them where it expects them. Thus `npm publish` will include the correct `gitHead` reference and thus the _next_ `semantic-release` run will see the correct, truncated change list.

```bash
git show-ref --head --heads | while IFS=' ' read -r hash name; do test ! -e "${GIT_DIR:-.git}/$name" && echo $hash > "${GIT_DIR:-.git}/$name"; done
```

[...]

PR #393, opened 2017-06-07
Populate gitHead, and use Git tag if it's not there

### LINKED ISSUE #280: Warn user about nonexistent gitHead returned by npm-registry-client
I have spent a day on trying to fix weird behavior of the semantic-release (eg empty releases, releasing on "chore" etc). The thing I finally figured out is that the NPM Registry API doesn't return `gitHead` [for some of versions](http://registry.npmjs.org/matreshka) and [last-release-npm](https://github.com/semantic-release/last-release-npm/blob/next/src/index.js#L42) does not check its existence (this is general issue so I decided to create an issue there).

The reason was the way I publish my library. I compile everything to npm/ folder and publish it from there.   NPM doesn't know about git repository. As dirty solution for my problem I copy .git folder from root to npm/.git before publish.

artemv:
This PR does 2 things:
* let `semantic-release pre` pre-write gitHead property together with version
* if `semantic-release pre` sees that the last released version doesn't have gitHead property, try to use the relevant vX.Y.Z git tag

NPM package's gitHead property is some undocumented thing, not a part of NPM contract (see e.g. npm/npm#3363) and they mostly write it on publish but sometimes not. One non-trivial case I found is after I've split `semantic-release pre && npm publish && semantic-release post` pre-set command to 2 separate CircleCI commands:
```
- semantic-release pre
# ... smth else
- npm publish && semantic-release post
```
In case CircleCI command starts with `npm publish` it doesn't write gitHead to published package.json even if it runs from proper git repo dir - maybe because CircleCI runs each comand in a new shell session. I reproduced it with NPM 3.10.8 on CircleCI (Ubuntu 14.04) but couln't reproduce on MacOS - maybe shell-dependent.

fixes #280 (Warn user about nonexistent gitHead returned by npm-registry-client)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -7,6 +7,7 @@ var log = require('npmlog')
 var nopt = require('nopt')
 var npmconf = require('npmconf')
 var normalizeData = require('normalize-package-data')
+var gitHead = require('git-head')
 
 log.heading = 'semantic-release'
 var env = process.env
@@ -138,11 +139,21 @@ npmconf.load({}, function (err, conf) {
             log.silly('pre', 'Couldn\'t find npm-shrinkwrap.json.')
           }
 
-          fs.writeFileSync('./package.json', JSON.stringify(_.assign(originalPkg, {
-            version: release.version
-          }), null, 2))
+          gitHead(function (err, hash) {
+            var add = {
+              version: release.version
+            }
+            var msg = 'Wrote version ' + release.version
+            if (err) {
+              log.error("Couldn't determine gitHead", hash, err)
+            } else {
+              add.gitHead = hash
+              msg += ', gitHead ' + add.gitHead
+            }
+            fs.writeFileSync('./package.json', JSON.stringify(_.assign(originalPkg, add), null, 2))
 
-          log.verbose('pre', 'Wrote version ' + release.version + ' to package.json.')
+            log.verbose('pre', msg + ' to package.json.')
+          })
         })
       })
     })
--- a/src/lib/commits.js
+++ b/src/lib/commits.js
@@ -11,8 +11,14 @@ module.exports = function (config, cb) {
   var from = lastRelease.gitHead
   if (!from) {
     from = config.lastRelease.version ? 'v' + config.lastRelease.version : false
-    console.warn('NPM registry does not contain "gitHead" in latest package version data. It\'s probably because ' +
-      'the publish happened outside of repository folder. Will try last version\'s Git tag instead: "' + from + '"')
+    var msg = 'NPM registry does not contain "gitHead" in latest package version data. It\'s probably because ' +
+      'the publish happened outside of repository folder.'
+    if (from) {
+      msg = msg + ' Will try last version\'s Git tag instead: "' + from + '".'
+    } else {
+      msg = msg + ' Using all the commits history to HEAD then.'
+    }
+    log.warn(msg)
   }
   var range = (from ? from + '..' : '') + 'HEAD'
 
--- a/src/lib/commits.js
+++ b/src/lib/commits.js
@@ -9,6 +9,11 @@ module.exports = function (config, cb) {
   var options = config.options
   var branch = options.branch
   var from = lastRelease.gitHead
+  if (!from) {
+    from = config.lastRelease.version ? 'v' + config.lastRelease.version : false
+    console.warn('NPM registry does not contain "gitHead" in latest package version data. It\'s probably because ' +
+      'the publish happened outside of repository folder. Will try last version\'s Git tag instead: "' + from + '"')
+  }
   var range = (from ? from + '..' : '') + 'HEAD'
 
   if (!from) return extract()


vanduynslagerp:
Thanks for looking at this issue !

I would have a few questions/remarks. As mentioned in npm/npm#3363 "gitHead" is not part of the spec so it might not be the best idea to add it in the package.json before publishing. Because it's not specified in the spec, it could change anytime and `npm publish` could change to store the 'gitHead' differently in the registry, overwriting it before publishing, or completely ignore it.

For now `npm publish` not publishing the `gitHead` seems to be a bug, as in most cases it's published. Did you manage to find more information about it or a reproduction case ?
So we could open a bug report with `npm` (or check if one exist already).

Regarding using the tag as a fallback, by default Travis doesn't pull tags, so they won't be available on the CI. I'm not sure about the other CI. So we should probably check if the project is a github repo and call the github API or do a `git fetch --tags`.

I think it's important to continue to publish package exactly the way`npm publish` would in a normal case (ie having the version and name in the package.json).
It's probably better for us to only workaround the consequences of a potential `npm publish` (by finding an alternative way to get the `gitHead` like tags) and not workaround the bug itself as we'll encounter it anyway on version not published by `semantic-release`.

artemv:
@vanduynslagerp yeah the gitHead concerns make sense. Looks like the best thing is to make `semantic-release pre` write some semantic-release specific property to package.json instead to make it independent of this undocumented npm thing? E.g. 
```
  "release": {
    "verifyConditions": "condition-circle",
    "gitHead": "bfc1d5eadb48039c038bc02a88a5c48fda1d28ab"
  },
```
and then fallback to looking at tags can be just removed.

vanduynslagerp:
Actually I was proposing the opposite :)

We have to deal with package published on npm without a git head, because that can happen if a package is published from a non git root directory or from outside the git directory. There will always be a lot of package published without `semantic-release` and we have to be able to deal with those.
And the best (only?) workarround seems to be to use the git tags.
We also have a request to allow to customize the tag in #431. We won't be able to implement both  as we would need the tag to match the version published on npm.

Regarding `npm` that doesn't publish the `gitHead` at that point we have no evidence it's a bug in `npm` and it seems to be related to something specific on your environment or on Circle CI.
You mentioned:
> In case CircleCI command starts with npm publish it doesn't write gitHead to published package.json even if it runs from proper git repo dir - maybe because CircleCI runs each command in a new shell session.

To be clear `npm publish` read the `gitHead` and include it by itself, it's not a `semantic-release` feature. So running `semantic-release pre && npm publish` or just `npm publish` doesn't affect the behavior of  `npm publish` regarding `gitHead`.
So most likely in your situation the problem is not `semantic-release pre && npm publish` vs `npm publish` but something else. For some reasons, in your configuration when Circle CI runs `npm publish` it doesn't have access to the git info.

=> We should analyze that and find the root cause of the problem before trying to implement a workaround in `semantic-release`.

Here is a few guesses about the potential root cause of your issue:
- When you run `npm publish` the current directory is not the root of your repo
- When you run `npm publish` the repo directory doesn't contains the `.git` directory

Could you check that ? Or could you provide a link to your repo and the Circle CI jobs so we can analyze more in details ?

vanduynslagerp:
For information when `npm publish` runs is reads the package.json [here](https://github.com/npm/npm/blob/latest/lib/publish.js#L82) and reading the package.json adds the gitHead [there](https://github.com/npm/read-package-json/blob/master/read-json.js#L327).
So we would need to figure out why in your situation [that function](https://github.com/npm/read-package-json/blob/master/read-json.js#L327) doesn't returns the `gitHead`.

artemv:
> Or could you provide a link to your repo and the Circle CI jobs so we can analyze more in details ?

Here I've set up a demo: https://github.com/artemv/semantic-test. The only thing needed to showcase the issue in CircleCI env is to split semantic-release into 2 parts: `semantic-release pre` runs after dependencies are installed, and then `npm publish && npm show && semantic-release post` publishes to NPM, shows the published version of package.json and publishes git tag. 

The commits history has BREAKING CHANGE so because of the gitHead issue it always publishes new major: https://circleci.com/gh/artemv/semantic-test/25 publishes version 6.0.0 and then https://circleci.com/gh/artemv/semantic-test/26 publishes 7.0.0, though the difference between them should be patch increment.

I also showcased the use of this PR's version of semantic release: when I switch to it it 
* warns that gitHead was missing in `semantic-release pre` - see `npm run semantic-pre` step in https://circleci.com/gh/artemv/semantic-test/22
* writes gitHead to package.json `semantic-release pre` so that for use in next CI build - it's then seen inside the output of `npm show` inside `npm run semantic-release` step of https://circleci.com/gh/artemv/semantic-test/22

So when https://circleci.com/gh/artemv/semantic-test/23 runs the gitHead is in place and all works as usual (except `semantic-release pre` writes gitHead to package.json again to make sure it's published).

vanduynslagerp:
Thanks so much for for the reproduction environment ! It's very helpful !
I created this [repo](https://github.com/vanduynslagerp/test-circleci) based on yours and I managed to reproduce the problem with a reduced test case.
It seem to be a Circle CI issue that drop the git refs informations in the deployment phase.

See https://circleci.com/gh/vanduynslagerp/test-circleci/33

In the dependencies step of the build we have a normal git repo with the head reference of the current branch:
```
$ cat .git/refs/heads/master
7e6b47bbb9c467808250a586df39996cd8e4300f
 ```

But in the deployment step:
```
$ cat .git/refs/heads/master
cat: .git/refs/heads/master: No such file or directory
cat .git/refs/heads/master returned exit code 1
```

For some reason, on the deployment step CircleCI clear the `.git/refs/heads` directory. So when `npm publish` runs it cannot get the gitHead with this [function](https://github.com/npm/read-package-json/blob/master/read-json.js#L327).

So the problem is not running `semantic-release pre && npm publish && semantic-release post` in one command or in two. It's related to which phase of the Circle CI build it runs.
So if you would run `semantic-release pre && npm publish && semantic-release post` in the deployment the problem would be the same because files are missing in the `.git` directory.
Both `semantic-release` and `npm publish` have to run on a valid git repository.

I did a bit of research and I couldn't figure out why Circle CI drops the content of `.git/refs/heads` on deployment phase... Seems to be a bug in Circle CI. I also figured out the the problem is similar during the test step.
I opened a support ticket with CircleCI here: https://discuss.circleci.com/t/git-refs-heads-master-missing-on-test-and-deployment-step/16865

Maybe CircleCI 2.0 solve the problem...

**Comment by artemv** [comment]
> So the problem is not running semantic-release pre && npm publish && semantic-release post in one command or in two. It's related to which phase of the Circle CI build it runs.

That's interesting because I remember we started to have this issue on one of my projects exactly when I split the command into two. So I tested your idea here - no, joined version does publishes gitHead: https://circleci.com/gh/artemv/semantic-test/27

**Comment by vanduynslagerp** [comment]
After more analysis I figured out why.
It's because [condition-circle](https://github.com/bahmutov/condition-circle) runs a special script to fix this Circle CI issue. See [refs.sh](https://github.com/bahmutov/condition-circle/blob/master/refs.sh).
The script fixes the problem by restoring the missing `.git/refs/heads/master`.
Apparently someone ran into this problem before....

`condition-circle` is ran by `semantic-release pre`, this is why running it just before `npm publish` fix the problem.

So if you want to run `semantic-release pre` on dependencies step and `npm publish && semantic-release post` on deployment you should run this script just before:
```yml
machine:
  node:
    version: 8.0.0
dependencies:
  post:
    - npm run semantic-pre
deployment:
  master:
    branch: master
    commands:
      - node_modules/condition-circle/refs.sh
      - npm run semantic-release
```

artemv:
Yeah, the mystery unveils, thank you! Found more context behind condition-circle's workaround [here](https://github.com/semantic-release/semantic-release/issues/256#issuecomment-239678516): 

> This arises from npm publish's brittle broken way of getting at that Git hash in the first place (PR [here](https://github.com/npm/read-package-json/pull/67)): It silently fails on perfectly valid Git repos if they use [packed refs](https://git-scm.com/docs/git-pack-refs) a certain way.

So, an NPM issue yeah, the one you've mentioned already. Well, at least semantic-release could do better at identifying the issue on `semantic-release pre` stage - it says nothing about missing gitHead at e.g. https://circleci.com/gh/artemv/semantic-test/27. I think I'll make another PR )

artemv:
Here: https://github.com/semantic-release/semantic-release/pull/447 - I kept the using tags thing there.

vanduynslagerp:
Yes I think we should improve the error messages and the documentation regarding those things.
Using the git tags as a backup when gitHead is not in the npm package is still a good idea as this situation can happen for many reasons.
I'm going to work on that soon. I'm also currently working on improving the logs to be more explicit and easier to read.

Maybe we could open an issue with [read-package-json](https://github.com/npm/read-package-json) suggesting a more robust way to get the git head. Or at least clarify if the `gitHead` is actually part of the `package.json` specification or not (if it is then we could set it in `semantic-release`).

For information we are also trying to figure out a way to have only one step for semantic-release that would handle pre, publish and post. But that's not so easy as we want to publish exactly the way `npm` does.

vanduynslagerp:
Nevermind the PR is already there: https://github.com/npm/read-package-json/pull/67

artemv:
> For information we are also trying to figure out a way to have only one step for semantic-release that would handle pre, publish and post. But that's not so easy as we want to publish exactly the way npm does.

Btw I have an input to this idea - the reason why we've split the semantic-release to two parts at some point is that it can be useful to have the package version before it's published - e.g. then we can console.log it from the JS code being run, useful for browser libs.

vanduynslagerp:
> Btw I have an input to this idea - the reason why we've split the semantic-release to two parts at some point is that it can be useful to have the package version before it's published - e.g. then we can console.log it from the JS code being run, useful for browser libs.

Agreed. I have the same problem on some libs.

My thoughts so far (nothing is settle has it hasn't been discussed) would be something like:
`npm install -g semantic-release`. Then just `semantic-release`.

And in addition:
`semantic-release --dry-run` that would print the new version and the changelog in the console
`semantic-release --update-version` that could be ran before the tests (including on a PR) to just bump the version in the `package.json`.

But there is a lot of side effect and edges case to iron out. So I can't tell what direction thing will end up going nor when.
I joined the maintainer team recently. And what I can tell for sure is that @gr2m and myself are quite motivated to move things forward, and hopefully people have already noticed some small improvements here and there.

artemv:
> And what I can tell for sure is that @gr2m and myself are quite motivated to move things forward, and hopefully people have already noticed some small improvements here and there.

Sure, nice to see that the project is alive again, thank you guys!

PR #399, opened 2017-06-28
Add troubleshooting guide to README

PatrickHeneise:
Starting a library of common issues with semantic-release and how to recover.

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,36 +292,6 @@ Use this in one of your projects? Include one of these badges in your README.md 
 [![semantic-release](https://img.shields.io/badge/%20%20%F0%9F%93%A6%F0%9F%9A%80-semantic--release-e10079.svg?style=flat-square)](https://github.com/semantic-release/semantic-release)
 ```
 
-## Troubleshooting
-
-### ENOTINHISTORY Commit not in history
-
-```
-semantic-release ERR! commits The commit the last release of this package was derived from is not in the direct history of the "master" branch.
-semantic-release ERR! commits This means semantic-release can not extract the commits between now and then.
-semantic-release ERR! commits This is usually caused by force pushing, releasing from an unrelated branch, or using an already existing package name.
-semantic-release ERR! commits You can recover from this error by publishing manually or restoring the commit "123".
-semantic-release ERR! pre Failed to determine new version.
-semantic-release ERR! pre ENOTINHISTORY Commit not in history
-```
-
-To restore semantic-release, follow these steps:
-
-```
-git pull
-git reset --hard origin/master
-npm version x.y.z
-git checkout -B chore/release
-git commit -am 'chore: release'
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -286,6 +286,36 @@ Use this in one of your projects? Include one of these badges in your README.md
...
+git reset --hard origin/master
+npm version x.y.z
+git checkout -B chore/release
+git commit -am 'chore: release'
+git push
+— merge (not squash-merge) on github (this is important before running git push)
+git checkout master
+git pull
+git push --tags
+npm publish
```
gr2m: I would suggest to leave out the pull-request flow and suggest to directly commit to master. If people care, they will know how to do it with pull requests.

I would suggest something like this

1. replace "0.0.0-development" in package.json with the next patch version of your `package.json` file. Find the current version with `npm info <package name> version`
2. run `npm publish`
3. revert the change to `package.json`
4. create a new tag with the new version, prefixed with `v`. So if the new version is `1.2.3`, run `git tag v1.2.3`
5. git push --tags
6. If you like, create a new release with the latest changes, as semantic release failed for the release. But it will work again from now on

gr2m: oh my, I forgot the most important thing: thank you so much for putting time into this and sorry it took so long to get back to it. I’ll help to maintain semantic-release now, you can follow news at https://github.com/semantic-release/semantic-release/issues/411

PatrickHeneise: Thanks for taking this over! In most of our projects we have the master branch protected, direct commits/pushes are not accepted, therefore the pull-request workflow.

I find `npm version x.y.z` easier than replacing the version tag in package.json and running git tag (also as this isn't updated by semantic-release, we don't usually have a `version` tag in our package.json).

gr2m: Yeah I have the same protection enabled for master on most of my repositories, too. But for the troubleshooting section, I’d describe the most simple case. When people don’t want or can’t push to master, they will figure it out because they know what they are doing. For newbies who just work directly on master, the simplified version is easier. 

We can leave the `npm version x.y.z` step, I thought to leave it out to avoid an extra commit, but then again, I can work around this myself, but for newbies it will be simpler this way, you are right 👍

PatrickHeneise: Ok, I’ll work on it tomorrow. I’d like to have both versions for the commit then, it really took me a little while to get this right (it really only worked in that order), the more there’s documented the easier it is for everyone (not just beginners) to figure things out without wasting time. Can also add a new file TROUBLESHOOTING.md instead of putting this in the readme if you’re concerned about the simplicity of the README?

PatrickHeneise:
@gr2m sorry this took so long. Moved it to a separate file and added the PR workflow comments. I'm sure beginners and experts alike will be able to fix the `ENOTINHISTORY` issue with this guide.

PR #408, opened 2017-08-08
feat(commits): check remote branches to determine relevant history

zetaron:
Depending on how the git repository got initialized on a CI system there
might occur issues when not
checking remote branches for the `from` commit.

For example Bitrise initializes it's git repositories as follows:
```
git "init"
git "remote" "add" "origin"
"git@github.com:zetaron/condition-bitrise.git"
git "fetch"
git "checkout" "52ab116ffe04c0f03e24e566b861cc1bde70024b"
git "submodule" "update" "--init" "--recursive"
```

This results in semantic-release being unable to detect which branch
contains the '52ab116ffe04c0f03e24e566b861cc1bde70024b' commit, as git
is in a detatched HEAD mode.
There will be no release as semantic-release breaks with a message just
like the following:
```
$ semantic-release pre && npm publish && semantic-release post
semantic-release ERR! commits The commit the last release of this
package was derived from is not in the direct history of the "master"
branch.
semantic-release ERR! commits This means semantic-release can not
extract the commits between now and then.
semantic-release ERR! commits This is usually caused by force pushing,
releasing from an unrelated branch, or using an already existing package
name.
semantic-release ERR! commits You can recover from this error by
publishing manually or restoring the commit
"52ab116ffe04c0f03e24e566b861cc1bde70024b".
semantic-release ERR! commits Here is a list of branches that still
contain the commit in question:
semantic-release ERR! commits  * * (HEAD detached at 52ab116)
semantic-release ERR! pre ENOTINHISTORY Commit not in history
error Command failed with exit code 1.
```

This commit introduces a small change to the comamnd used to detect the
commits branch, by adding the `--remote` option and normalizing remote
branch names to be matchable against the `options.branch` (eg. master)

### DIFF
--- a/src/lib/commits.js
+++ b/src/lib/commits.js
@@ -13,14 +13,14 @@ module.exports = function (config, cb) {
 
   if (!from) return extract()
 
-  childProcess.exec('git branch --no-color --contains ' + from, function (err, stdout) {
+  childProcess.exec('git branch --no-color --remote --contains ' + from, function (err, stdout) {
     var inHistory = false
     var branches
 
     if (!err && stdout) {
       branches = stdout.split('\n')
       .map(function (result) {
-        if (branch === result.replace('*', '').trim()) {
+        if (branch === result.replace('*', '').replace('origin/', '').trim()) {
           inHistory = true
           return null
         }


gr2m:
Hey @zetaron, I’m just familiarising myself with semantic-release as I joined it as maintainer today. 

Thanks a lot for putting the time into the pull-request, I’ll do my best to get this merge as I understand that it’s a blocking issue for you right now.

As this issue didn’t come up before (I think), can you check with Bitrise if they maybe might want to change their setup to align with the other CI services? 

If that is not an option, could I please ask you for two more things
1. add tests for the change
2. add a comment to explain the additional `.replace('origin/','')`

Thanks 💐

zetaron:
Hey @gr2m, that's great news :) 

I'll try to reach out to Bitrise and see if I can get some feedback on changing the way git initializes its repositories on their platform. 

For 1 and 2 I'd wait until I've talked to Bitrise but in case they're not willing to change their setup I'll take these on. 

Just a small request: Cloud you point me to how to setup the test environment locally? Unit tests are not the problem but there seem to be integration tests with a dependency on some sort of fake registry that requires a database?

zetaron:
I'll see if I can get around to the requested changes 1 and 2 on the coming weekend.

zetaron:
@gr2m I'm actually having trouble finding the spec file where I'd extend the test suite in respect to the change this PR provides.
Maybe you could describe a little bit how the suites are structured and eg. on which directories they operate, for example I initially thought to extend the `test/specs/commits.js` suite but then was unable to understand how the current tests work and how I'd add my own.

pvdlg:
@zetaron thanks for the contribution!

Checking the remote branch might not completely work though... We check if the last release `gitHead` is present in the local branch in order to make sure:
- The source code we currently have on the CI contains all the changes between the last release and the last commit we have in the repo
- We have all the commits between the last release and now

Checking that the last release `gitHead` is present on the remote branch will not guaranties it's present in the current repo on the CI, as we can't verify how far back commit history in the local clone goes or from which branch the detached head was created.

Maybe a better solution would be to use a different command that would be able to check if the commit is present in the history of the branch or the detached head indifferently.

I didn't test much but maybe something along those line would do: `git cat-file -t <sha>^\{commit\}` or `git rev-parse --verify <sha>^{commit}` ?
I'm not really sure if that would work in every cases, or it would have some blind spot or false positive in some cases. So more test would needed.

Or do you know another way to make that check ?

PR #414, opened 2017-08-14
Deprecate all Node versions below 8

boennemann:
[v7.0.0](https://github.com/semantic-release/semantic-release/releases/tag/v7.0.0) just removed support for Node < 4 and with this PR im proposing to be even more radical.

Here is my reasoning (quoted from the FAQ entry I added).

> Being able to write code for just the most recent node versions greatly simplifies development. More language features are available, no transpilation is required, less test builds are to be run, awaited and debugged.
> For a special purpose tool like `semantic-release`, that's only meant to be used in controlled CI environments, we think it's okay to have such a high version requirement. As `semantic-release` handles package publishing we expect almost every project to have at least one build job running node 8 already – and that's all it takes. Even if that's not that case `semantic-release` can still be executed with the help of [npx](https://www.npmjs.com/package/npx) (`npx -p node@8 npm run semantic-release`).

Changes:

- Only test node 8 on Travis
- Early exit with helpful error message if process.version < 8
- New FAQ entry

Future changes:

download

<a name="progress"></a>
### Progress

- [x] release breaking version of `@semantic-release/travis-deploy-once` with `@next` dist-tag. ([done](https://github.com/semantic-release/travis-deploy-once/releases/tag/v2.0.0))
- [x] release breaking version of `@semantic-release/conditication-travis` with `@next` dist tag updating to latest `@semantic-release/travis-deploy-once`
- [x] adapt error message at https://github.com/semantic-release/semantic-release/blob/node-8-only/bin/semantic-release.js#L12-L14
- [x] release breaking version of `semantic-release` with `@next` dist tag updating to latest `@semantic-release/conditication-travis`
- [ ] `@next` version announcement and testing
- [ ] bump `@latest` dist tag for all 3 

closes https://github.com/semantic-release/semantic-release/pull/418


package.json
```diff
@@ -47,7 +47,7 @@
     "tap": "^10.0.1"
   },
   "engines": {
-    "node": ">=0.10",
+    "node": ">=8",
     "npm": ">=2"
```
gr2m: shall we bump to npm 5, too?

gr2m:
I think we can even work around the problem that the build leader on travis must be running on Node 8. When getting a build with [travis-ci]() library using `travis.builds(id).get(callback)`, the response includes a jobs array. Here is the same response using curl (the header is significant)

```
$ curl -H'Accept: application/vnd.travis-ci.2+json, */*; q=0.01' https://api.travis-ci.org/builds/264152577 
```

The response looks like this (excerpt

```js
{
  "build": {
    "id": 264152577,
    "repository_id": 5110092,
    // ...
    "config": {
      "node_js": [
        6,
        8
      ],
      // ...
    },
  },
  "jobs": [
    {
      "id": 264152578,
      "number": "58.1",
      // ...
      "config": {
        "node_js": 6,
         // ...
      }
    },
    {
      "id": 264152579,
      "number": "58.2",
      // ...
      "config": {
        "node_js": 8,
         // ...
      }
    }
  ]
}
```

Instead of declaring "58.1" the job leader just because its job number ends with ".1" (like it is implemented today), we could find the first job with a node version >= 8 and declare this job the job leader, no matter the version.

bin/semantic-release.js
```diff
@@ -1,165 +1,31 @@
...
+if (semver.lt(process.version, '8.0.0')) {
+  console.error(
+`semantic-release: node version >= 8 is required. Found ${process.version}.
 
-    plugins.verifyConditions(config, function (err) {
-      if (err) {
-        log[options.debug ? 'warn' : 'error']('pre', err.message)
-        if (!options.debug) process.exit(1)
-      }
+Don't fret. It's not that big of a deal.
```
gr2m: I know people will have strong opinions about. I know this is meant to sound fun & human, but it just provoke heated emotions. I would suggest the information super dry here, the argument is solid, no need to tell anyone not to fret, I’d just go straight into a simple, clear explanation.

bin/semantic-release.js
```diff
@@ -1,165 +1,31 @@
...
+`)
+  process.exit(1)
+}
 
-    require('../src/post')(config, function (err, published, release) {
-      if (err) {
-        log.error('post', 'Failed to publish release notes.', err)
-        process.exit(1)
-      }
+// Let's enter the beatiful lands of node 8+
```
gr2m: similar to my comment above, I’d suggest to leave it out.

vanduynslagerp:
> I think we can even work around the problem that the build leader on travis must be running on Node 8.

Travis released [Build stages](https://docs.travis-ci.com/user/build-stages/) in beta a few months ago.
Using that would allow to greatly simplify the verification:
- No need to verify for build leader
- No need for `travis-deploy-once`

gr2m:
@vanduynslagerp I tried that, but run into several problems with it. Among others, there is no way from the script being run to tell that it’s run as a stage, at least not from environment variables alone. 

But thanks for bringing it up, it’s out of scope for this pull request, but we sure will give it more thought

package.json
boennemann: I'm all for it, if we have a reason. E.g. for security we could enforce sha512 hashes when publishing.

But this package.json entry doesn't really do much, so we'd have to write some additional code first. I'd do all of this in a separate step.

boennemann:
@vanduynslagerp @gr2m I still conceptually like travis-deploy-once better, because it's zero config on Travis (Pro). We can however think about making it possible to use build stages for more complex setups in the future.

@gr2m Updated the comments, thanks for the hints. I suggest we leave the comments about "build leader" in until we can merge a travis-deploy-once that's capable of auto choosing the leader.

boennemann:
Alrighty, I went ahead and added node 8 detection into travis-deploy-once https://github.com/semantic-release/travis-deploy-once/pull/7

Once we have that released we can add it here, and make this change even less likely to break end-users.

package.json
boennemann: Btw, by enforcing node8 npm5 becomes the de facto min npm version – unless people explicitly downgrade.

gr2m:
@boennemann the `npx` trick didn’t work:

https://travis-ci.org/hoodiehq/hoodie/builds/276387784#L2843

```
$ npm install --global npx
$ npx -p node@8 npm run semantic-release
> hoodie@0.0.0-semantically-released semantic-release /home/travis/build/hoodiehq/hoodie
> semantic-release pre && npm publish && semantic-release post
semantic-release: node version >= 8 is required. Found v4.8.4.
```

Any thoughts?

gr2m:
`npx -p node@8 node ./node_modules/.bin/semantic-release` should work, I’m currently testing it via https://travis-ci.org/hoodiehq/hoodie/builds/279660236

gr2m:
> npx -p node@8 node ./node_modules/.bin/semantic-release should work

🤦‍♂️ no of course it does not, because the `semantic-release` npm task is more than just running the binary. Testing `npx -p node@8 semantic-release` now

gr2m:
okay that also didn’t work, but `npx -p node@8 -c "semantic-release pre && npm publish && semantic-release post"` will I think. Not very pretty but if it works, it works :D

gr2m:
that worked, we can just update the error / documentation to recommend `npx -p node@8 -c "semantic-release pre && npm publish && semantic-release post"`: 
https://travis-ci.org/hoodiehq/hoodie/builds/279667094#L2846

Issue #97, opened 2015-09-27
Release a beta version


ericdouglas:
Is possible to release a beta version with semantic-release?

boennemann:
If by "beta" you mean sth. like `1.0.0-beta.1 the default `analyzeCommit` plugin doesn't support them out of the box. If you find some heuristics on whether to publish those you can implement this though: https://github.com/semantic-release/semantic-release#analyzecommits

Even though this possible with your own plugin I'd strongly discourage you from using "beta versions". semantic-release has support for release-channels (https://github.com/semantic-release/semantic-release#how-does-it-work 3rd paragraph) by adding `"publishConfig": {"tag": "next"}` to your package.json. This is way better mechanism to communicate stability.


Best,
Stephan

ericdouglas:
Great, @boennemann ! Many thanks :smile:

Apidcloud:
Is there any way to release on GitHub under the same dist-tag? as mentioned in https://github.com/kentcdodds/ama/issues/318

Thanks!

gr2m:
`semantic-release` does not support that, it only creates tags with the versions released on npm. 

I don’t see the use case for it. Can you explain why you’d want that? Would that be in addition to the version tags like `v1.2.3`, or as a substitute?

Apidcloud:
At least as a substitute just like it happens over npm. Despite using a dist tag such as **beta**, GitHub keeps using a number as the version.

gr2m:
yes that is intended behavior :) At the end, both npm & GitHub have releases with version numbers, but only npm gets the beta/next/custom tag, that is correct. Nobody asked for that to be added to GitHub yet. I wouldn’t be opposed adding it though. You’d also need to reset it if it already exists. Pull requests welcome 👍

PR #431, opened 2017-09-15
feat: support custom release tag in GitHub

Apidcloud:
It's just a simple change, but as mentioned in https://github.com/semantic-release/semantic-release/issues/97, this PR should make GitHub releases use a custom tag whenever defined - just as it happens with npm release.

Does it need anything else? @gr2m said something about resetting the tag. I wonder if that's really needed. 

I have noticed the tests (even before the changes) fail though. Seems to be an issue in plugins?
```
test/specs/plugins.js ................................. 4/8
  export plugins
  not ok Unexpected token function
```

### DIFF
--- a/src/post.js
+++ b/src/post.js
@@ -24,17 +24,30 @@ module.exports = function (config, cb) {
       if (err) return cb(err)
 
       var ghRepo = parseSlug(pkg.repository.url)
-      var tag = {
+      var versionName = 'v' + pkg.version
+      var customTagName = (pkg.publishConfig || {}).tag
+
+      var refPart = 'tags'
+
+      var versionTag = {
         owner: ghRepo[0],
         repo: ghRepo[1],
-        ref: 'refs/tags/v' + pkg.version,
+        ref: 'refs/' + refPart + '/' + versionName,
         sha: hash
       }
+
+      var customTag = customTagName ? {
+        owner: ghRepo[0],
+        repo: ghRepo[1],
+        ref: 'refs/' + refPart + '/' + customTagName,
+        sha: hash
+      } : null
+
       var release = {
         owner: ghRepo[0],
         repo: ghRepo[1],
-        tag_name: 'v' + pkg.version,
-        name: 'v' + pkg.version,
+        tag_name: versionName,
+        name: versionName,
         target_commitish: options.branch,
         draft: !!options.debug,
         body: log
@@ -57,8 +70,37 @@ module.exports = function (config, cb) {
         })
       }
 
-      github.gitdata.createReference(tag, function (err) {
+      github.gitdata.createReference(versionTag, function (err) {
         if (err) return cb(err)
+
+        // create reference based on the custom tag
+        var customReference = {
+          owner: ghRepo[0],
+          repo: ghRepo[1],
+          ref: refPart + '/' + customTagName
+        }
+
+        // try to retrive it
+        github.gitdata.getReference(customReference, function (err, ref) {
+          // if it doesn't exist,
+          if (err) {
+            // try to create it
+            github.gitdata.createReference(customTag, function (err) {
+              if (err) return cb(err)
+            })
+          } else if (ref) {
+            // if reference exists, try to delete it
+            github.gitdata.deleteReference(customReference, function (err, msg) {
+              if (err) return cb(err)
+              else if (msg) {
+                // if delete was successful, try to create again
+                github.gitdata.createReference(customTag, function (err) {
+                  if (err) return cb(err)
+                })
+              }
+            })
+          }
+        })
 
         github.repos.createRelease(release, function (err) {
           if (err) return cb(err)
--- a/test/specs/post.js
+++ b/test/specs/post.js
@@ -12,6 +12,14 @@ var pkg = {
   repository: {url: 'http://github.com/whats/up.git'}
 }
 
+var customTagPkg = {
+  version: '1.0.0',
+  repository: {url: 'http://github.com/whats/up.git'},
+  publishConfig: {
+    tag: 'beta'
+  }
+}
+
 var plugins = {
   generateNotes: function (pkg, cb) {
     cb(null, 'the log')
@@ -66,6 +74,19 @@ test('full post run', function (t) {
       tt.is(published, true)
       tt.match(release, defaultRelease)
 
+      tt.end()
+    })
+  })
+
+  t.test('production with custom tag', function (tt) {
+    post({
+      options: {githubToken: 'yo', branch: 'master'},
+      pkg: customTagPkg,
+      plugins: plugins
+    }, function (err, published, release) {
+      tt.error(err)
+      tt.is(published, true)
+      tt.match(release, defaultRelease)
       tt.end()
     })
   })
--- a/test/mocks/github.js
+++ b/test/mocks/github.js
@@ -6,6 +6,12 @@ module.exports = function () {
     gitdata: {
       createReference: function (release, cb) {
         cb(null)
+      },
+      getReference: function (reference, cb) {
+        cb(null)
+      },
+      deleteReference: function (reference, cb) {
+        cb(null)
       }
     },
     repos: {


src/post.js
```diff
@@ -24,17 +24,19 @@ module.exports = function (config, cb) {
       if (err) return cb(err)
 
       var ghRepo = parseSlug(pkg.repository.url)
+      var tagName = (pkg.publishConfig || {}).tag || 'v' + pkg.version
+
       var tag = {
         owner: ghRepo[0],
         repo: ghRepo[1],
-        ref: 'refs/tags/v' + pkg.version,
```
gr2m: now you are replacing creating the `v1.2.3` tags. We will not remove that feature. But we could additionally create a tag if a custom tag is set in `publishConfig`.

test/specs/post.js
```diff
@@ -9,6 +9,9 @@ var post = proxyquire('../../src/post', {
 
 var pkg = {
   version: '1.0.0',
+  publishConfig: {
+    tag: 'beta'
+  },
```
gr2m: Don’t change an existing test, please create a new on

src/post.js
Apidcloud: Not sure I understand the first part. Shouldn't it match with release.tag_name and release.name?

test/specs/post.js
Apidcloud: I was about to, but it raises the error I mentioned even without the changes. What do you recommend to fix that?

src/post.js
gr2m: instead of creating a single tag, we know would create 2 tags at once:

1. the version tag in the form of `v1.2.3`
2. the release tag which matches the name of the tag used to publish the package on npm

Does that make sense?

Apidcloud: ah, so this line would be kept in v1.2.3 format and only the release one changes if publishConfig is present?

test/specs/post.js
gr2m: the tests should pass, they just did about 19h ago on travis:
https://travis-ci.org/semantic-release/semantic-release/builds/275739048

The tests that currently fail are related to your changes.

Before the changes, what errors do you see?

Apidcloud: Hmmm that's weird. I forked and just ran npm install and npm test. Do I need anything else?

Apidcloud: Log: https://pastebin.com/HARMnXg5

gr2m: try to run in node 8, we recently deprecated support for lower versions via https://github.com/semantic-release/semantic-release/pull/414

I think you also need to setup CouchDB locally for the integration tests to work: http://couchdb.apache.org/

Apidcloud: Alright, just updated to node 8.5 and npm to 5.3 and it seems to be the cause. It still raises the following error though: https://pastebin.com/PPDfSnz3

gr2m: have you installed CouchDB? Can you run `couchdb` in your terminal?

Apidcloud: I've installed it, and it's running in http://localhost:5984/_utils/ but I'm not sure what else I need to set it up for the tests

gr2m: The whole setup is not ideal right now, I don’t even know if it’s working on Windows :( 

as a workaround, can you try to adapt these lines for local testing:

https://github.com/semantic-release/semantic-release/blob/88b5895f800f608eca3f1f4b435b34891d00082d/test/registry/start.sh#L22-L23

Change the port number in L22 to to `5984` and comment out the second. Make sure you created an admin account with username "admin" and password "password" at http://localhost:5984/_utils/

Apidcloud: Still doesn't seem to work. I will try with a different couchdb version tomorrow, as running couchdb is producing some weird output (kernel-poll not supported).

gr2m: I’m sorry for the trouble, thanks for your patience! If it doesn’t work, you can still push and have travis run the integration tests.

Apidcloud: It seems to be a problem with privileges in Windows... Still trying to figure out a solution.

src/post.js
gr2m: nothing changes, if we release a new version `1.2.3` to npm, we will still create a git tag `v1.2.3`. But if `publishConfig.tag` is to set `"next"`  then we can create an _additional_ git tag called `next`. 

What I’m not exactly sure about is what happens if a git tag `next` already exists from an earlier release. We would probably check if it does, if yes delete it, then create it again.

Apidcloud: So I just need to create an additional reference (github.gitdata.createReference) when publishConfig.tag is present, right?

gr2m: yes :)

Apidcloud:
Please check the changes 👍 I'm still not sure how to test it thoroughly though.

gr2m:
I also don’t know yet, but I’ll have a look. I’ll be busy with work until Saturday, I'll have a look then.

What you can do until then is to setup a test repository for a test npm package, set `"semantic-release": "Apidcloud/semantic-release#GitHubCustomTag"` in your package.json so your fork gets installed, and then see if it works, if there are any problems left you’ll catch it that way

Apidcloud:
I have been trying to delete or update the custom tag reference for a few hours (yeah...) without success.

When I try to create a reference, it says it already exists. But when I try to delete or update the said reference, it says it doesn't exist. It's not making any sense 😢 

Any ideas?

pvdlg:
@Apidcloud thanks for the contribution!

We now use the git tags as a fallback to determine the commit that was used to make the last release. And we expect the tag associated with the release to be named according npm recommendation (or at to the way `npm version` creates it).

In addition, it's not possible to have a version released multiple time on different tags.
So if you release `2.0.0` on the `@next` diest-tag you won't be able to release `2.0.0` on the `latest` dist-tag. You would have to run `npm dist-tag add <pkg>@2.0.0 latest` to add the `latest` dist-tag to version `2.0.0`. See [dist-tag](https://docs.npmjs.com/cli/dist-tag).

That means that when you release `2.0.0` with the `next` dist-tag and create the git tag `tags/v2.0.0` (as it's done in this PR, or maybe you intended to do `next/v2.0.0` ?) at some point `v2.0.0` will be made available on the `latest` dist-tag and the git tag won't make much sense anymore. It will just reflect that originally `v2.0.0` was released as a beta, even though it's now available on `latest`.
Not all the versions you release on `next` are meant to become available on `latest` as some might have bugs, but some are, and they will have a inconsistent tag name.

What would be the objective of your proposal? Could you describe which scenario you'd like to achieve? Thanks !

src/post.js
```diff
@@ -57,9 +70,38 @@ module.exports = function (config, cb) {
...
+              if (err) return cb(err)
+              else if (msg) {
+                // if delete was successful, try to create again
+                github.gitdata.createReference(customTag, function (err) {
+                  if (err) return cb(err)
+                })
+              }
+            })
+          }
+        })
```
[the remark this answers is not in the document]
Apidcloud: This part makes sure the reference (tag) doesn't exist when we create it, otherwise, it throws an error. But on a first glance, I believe I forgot to check whether the custom tag was defined in the first place.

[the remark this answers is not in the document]
pvdlg: Isn't the github API would already throw an error when you create a tag that already exist?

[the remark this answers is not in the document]
Apidcloud: Exactly, and thus deleting it when that's the case.

test/specs/post.js
```diff
@@ -70,5 +78,18 @@ test('full post run', function (t) {
...
 
+  t.test('production with custom tag', function (tt) {
+    post({
+      options: {githubToken: 'yo', branch: 'master'},
+      pkg: customTagPkg,
+      plugins: plugins
+    }, function (err, published, release) {
+      tt.error(err)
+      tt.is(published, true)
+      tt.match(release, defaultRelease)
```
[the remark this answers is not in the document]
Apidcloud: Yeah, I wasn't sure how to test it. Can give it another try though.

pvdlg:
> The first idea was to mirror what happens over npm when releasing under a custom tag (say alpha).

I might be wrong but I think npm create a tag like `v2.0.0` when you do `npm version <patch|minor|major>` and doesn't create a tag when doing `npm publish --tag next`.
So it seems that what happen with npm is that the git tag is not related in any way to the dist-tag. 

> But as the GitHub behaviour is rather different, the purpose of this PR is to create a 2nd tag (when defined) that points to the same release. Doesn't have to be next.

It didn't occur to me that the code was creating a second tag, but I checked fairly quickly :)
What's the point of creating a second tag that points to the same version? Can you describe your workflow or how that would be useful ?

src/post.js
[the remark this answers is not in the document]
pvdlg: What I mean is:
- We probably don't want to delete tags. We have no way to check if the original tag is intentional or not. We don't want to take any destructive action without user consent.
- If the objective is to test if the tag exist before creating it, the "create tag" API call is already doing that.

Apidcloud:
Ok. I'll try to explain it based on what I wanted (and still want) to achieve in the first place.
I'm developing some packages that aren't ready for production use. With the current semantic-release version, when it comes to npm releases, I'm able to do it under a beta or alpha tag. 

In GitHub the behaviour is different. There's just a numerical version (e.g., v0.4.0) as explained in the issue https://github.com/semantic-release/semantic-release/issues/97 by @gr2m:
> but only npm gets the beta/next/custom tag

What I originally wanted was to mirror what happens in npm, so we thought about creating a 2nd tag (custom tag, only when defined) reference that points to the same numerical release.

Does that make sense?

src/post.js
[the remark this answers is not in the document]
Apidcloud: I agree with you. Originally I thought the same tag could have multiple versions within, but I believe that's not possible and thus the need of replacing it.

Would `next/v2.0.0` as you mentioned earlier make more sense?

pvdlg:
I understand. But as I was trying to explain in this [comment](https://github.com/semantic-release/semantic-release/pull/431#issuecomment-333700674) the notion of git tag and npm dist-tag are quite different.

A git tag flags or mark as a milestone a commit in the history and is not meant to be changed.
An npm dist-tag is distribution channel (Arguably the term dist-channel would be a lot less confusion). It works like the Chrome dist channel:
- Version `2.0.0` is release (it's the only version `2.0.0` that will ever be released and the commit associated with `2.0.0` will never change)
- This version is cutting edge, has many new features and might have bugs, so it's made available only on the distribution channel `next` (or dist-tag in npm world). Therefore only early adopters (aka adventurous users) will use it because they opt in by setting the dependency as 'my-package@next'.
- If everything goes well and it turns out nobody report bugs on `v2.0.0` for a month or so, we decide to also make it available to regular users by doing `npm dist-tag add my-package@2.0.0 latest`. Version `2.0.0` will now be distributed to everyone. But `2.0.0` is still the same, point to the same commit and the same git tag.

In that scenario there is no version `2.0.0-beta` and version `2.0.0-stable`. There is only `2.0.0` distributed to a subset of users for a month then distributed to everyone.

So it doesn't make much sense to have a git tag, meant to represent a milestone in the evolution of the package, to be named after it's temporary distribution channel.

The way github handle that is with pre-release. See [creating-releases](https://help.github.com/articles/creating-releases/), especially:
> If the release is unstable, select This is a pre-release to notify users that it's not ready for production.

What would make more sense would be to create github release flagged as `pre-release` at first (when your package is distributed only on `next` channel/dist-tag).
And when/if you are ready to make that release available to everyone you would do `npm dist-tag add my-package@2.0.0 latest` and go to the github release and uncheck the `pre-release`.

In terms of semantic-release you would configure your `package.json` like this:
```json
 "publishConfig": {
    "tag": "next"
 },
"release": {
  "pre-release": true //New option we would have to add
}
```
Each time `semantic-release` would make a release it would be:
- Available only on the npm `next` distribution channel
- Flagged as pre-release on Github

If the version is never deemed good enough for general availability, it would stay like that. But if it is deemed good enough for general availability you can move it to the npm `lastest` distribution channel and remove the pre-release flag on github.

Does that make sense? Would that work for you?

pvdlg:
Here is a good blog post on the subject: https://blog.greenkeeper.io/one-simple-trick-for-javascript-package-maintainers-to-avoid-breaking-their-user-s-software-and-to-6edf06dc5617

pvdlg:
In addition, it's worth noting that npm does support the prereleases but as mentioned [here](https://github.com/semantic-release/semantic-release#analyzecommits) `semantic-release` doesn't recommends to use them and recommends to use dist-tags instead, fitting the scenario I explained above.

Apidcloud:
Thank you for the wonderful explanation and link. I think I understand it better now.

Would the migration to latest dist channel be done manually though? And should this PR focus on creating a pre release in GitHub when the dist tag is next and the pre release is set to true? (Is this flag even needed? Shouldn’t it happen by default as the tag is next?)

Thanks again! I’ll try to take a look tomorrow.

pvdlg:
> Would the migration to latest dist channel be done manually though?

I think so, because `semantic-release` wouldn't have a way to determine which release can be moved to the `latest` channel nor when to do so.

> And should this PR focus on creating a pre release in GitHub when the dist tag is next and the pre release is set to true?

If that fit your need and you would like to make a PR for this feature, yes. But maybe it would be better to start a new PR (referencing this on in the comment). Also it's probably better to do that independently of the release branch configured in `"release": {"branch": "branch-to-release-from"}`.

> Is this flag even needed? Shouldn’t it happen by default as the tag is next?

Some users might name their beta/unstable/preview/notProdReadyYet branch other than `next`. In addition some users might want to not use the Github pre-release feature at all independently of the distribution channel they choose. So it's probably better to keep it configurable.

@gr2m do you have some feedback / concerns on this ?

gr2m:
> And should this PR focus on creating a pre release in GitHub when the dist tag is next and the pre release is set to true?

The only problem I see with this is that when you set the `@latest` dist-tag on npm to the new version, you should have to manually update the release on GitHub, too, as it no longer is a pre release.

> Is this flag even needed? Shouldn’t it happen by default as the tag is next?

I’m undecided. I’d be okay flagging a GitHub release as "pre-release" if the dist tag is set to anything else than `latest`, but we should inform folks about that in the docs and tell them that they have to remove the flag once they set the `@latest` dist-tag on npm.

pvdlg:
> The only problem I see with this is that when you set the @latest dist-tag on npm to the new version, you should have to manually update the release on GitHub, too, as it no longer is a pre release.

Yes indeed. I'm fiddling with what we can do regarding the refactoring of the cli (the main `semantic-release` not the setup one) to have only one command (instead of pre > publish > post and the reliance on 1 exit code). Maybe we could add the possibility to run manually `semantic-release --latest <version>` that would do `npm dist-tag add <package>@<version> latest` and remove the pre-release flag on github at the same time.

> I’m undecided. I’d be okay flagging a GitHub release as "pre-release" if the dist tag is set to anything else than latest, but we should inform folks about that in the docs and tell them that they have to remove the flag once they set the @latest dist-tag on npm.

We also have to handle situations where folks publish on a `legacy` dist channel in order to do bug fixes on old versions. For example one can maintain a legacy version that retain node 4 compatibility, and release only bug fixes there. So the package users could reference `<package>@legacy` and receive only the latest bug fixes on the node 4 compatible legacy version.
That's probably a very rare situation though. But maybe if we do a Github pre-release by default on any publish that is not on `latest` dist-tag, we should have a way to override that behavior for some specific dist-tags.

gr2m:
>  Maybe we could add the possibility to run manually semantic-release --latest <version> that would do npm dist-tag add <package>@<version> latest and remove the pre-release flag on github at the same time.

That’d be super nice!

I’m also thinking about the GitHub app that could do such things, but that’s out of scope and would only work for GitHub. Love the idea to make it work with the CLI!

gr2m:
> We also have to handle situations where folks publish on a legacy dist channel in order to do bug fixes on old versions. For example one can maintain a legacy version that retain node 4 compatibility, and release only bug fixes there

Good point ... we could also add another option the package.json publishConfig ... or be smart about it and see if the released version is higher than the currently `@latest`, and only then set the GitHub release to be a pre-release?

pvdlg:
> I’m also thinking about the GitHub app that could do such things, but that’s out of scope and would only work for GitHub. Love the idea to make it work with the CLI!

I was thinking the same :) But that would require a user action (as the decision to make a release available on the `latest` channel can only be made by a human). Therefore a UI. Therefore quite some work (at least for someone with no design talent like me :-))

pvdlg:
> Good point ... we could also add another option the package.json publishConfig ... or be smart about it and see if the released version is higher than the currently @latest, and only then set the GitHub release to be a pre-release?

Yes, but we are touching a complex subject. I've been thinking about that and hoe deal with or prevent situations of "non consistent" release across dist-tag.

For example if folks have a `legacy`, `latest` and `next` dist-tags.
Last release on `legacy` is `1.1.0`, last release on `latest` is `2.0.0`, last release on `next` is `2.1.0`.
What if you make a breaking change commit on the `legacy` channel ? => Fails because `2.0.0` already exists.
What if you make a breaking change commit on `latest`? Release `3.0.0`, but the last release on `latest` is higher than last release on `next`. I think npm would handle that without any issue but seems weird. Or not. I'm not sure.

I have a few ideas on how to handle those cases but the deeper I dig the more edges case I find and the more unlikely they get. My best shot so far would be to have the `last-release-npm` plugin to also report the versions already released in other tags or unpublished (as npm doesn't allow to publish a version on an unpublished one) and to skip those versions when we determine the next release version number. 

Anyway that's not really the subject of this PR. But maybe it's safer to keep it configurable, and try later to improve the algorithm to automatically determine the appropriate config?

Apidcloud:
It might be better to create another PR then. And just to make sure, it would address the following:
* A custom tag other than `latest `+ `pre-release flag` mark the GitHub release as pre-release.

Does that sound ok?

pvdlg:
> A custom tag other than latest+ pre-release flag mark the GitHub release as pre-release.
Does that sound ok?

The idea is to mark the github release as a pre-release if the dist-tag in `publishConfig` is not `latest` by default. But to add a possibility to override that with a config that would allow to define other branches on which we want a non pre-release release.

Maybe an option with that can accept a string or an array of string that defines the branches the branches on which we don't do a non pre-release. If this config is not present, then we consider it's `['latest']` by default.

Keep in mind that in order to accept the PR we would ask to have unit tests that cover all the cases introduced by the change. I refactored all the unit test recently so they should be easier to understand, but feel free to ask any questions :)

Apidcloud:
Ok. Where would that branch config be placed? Inside `publishConfig`?

pvdlg:
I'd say in `release` as it's `semantic-release` specific and `publishConfig` is npm specific.

gr2m:
> What if you make a breaking change commit on the legacy channel ? => Fails because 2.0.0 already exists.

Well you don’t, this would be like an LTS version, you keep maintaining it, but you can’t do breaking changes. We had discussion on this somewhere else. The idea is that when I release 3.0.0 I’d create a branch 2.x.x and in that branch I could set a base version to "2" so releases from that version would release as the next version within 2.x.x, but it would not allow breaking changes at all. Not sure what the npm tag would be, we could default to `v2` or we could require a dist-tag setting that isn’t "latest".

pvdlg:
> Well you don’t, this would be like an LTS version, you keep maintaining it, but you can’t do breaking changes.

Yep. Enforcing this kind of things is not something we do currently. But we probably should. We would need to know the precedence of dist tags though (`next` > `latest` > `legacy`). The `verifyRelease` plugin might be a good candidate for that. And a GitHub app would allow us to handle the situation better by opening an issue or even better, setting the status of the PR that would introduce a non allowed breaking change.

Apidcloud:
> Maybe an option with that can accept a string or an array of string that defines the branches the branches on which we don't do a non pre-release. If this config is not present, then we consider it's ['latest'] by default.

Shouldn't it be `['master']` by default? Also, what do you think of naming the flag `no-pre-release` instead?

pvdlg:
My mistake, I said branch but I meant dist-tag. We are checking the dist-tag not the branch to determine if we do a github pre-release or not. So the default should be `['latest']`.

Ok for `no-pre-release`, I can't find a better name.

Apidcloud:
I've just created another PR. I believe this one might be closed.

pvdlg:
Thanks so much ! I'll review shortly

PR #432, opened 2017-09-16
fix: Remove loglevel notice error

### LINKED ISSUE #430: Invalid config warning
I repeatedly get this warning: 

```
> semantic-release WARN invalid config loglevel="notice"
```

Where can I find more information on it?

RichardLitt:
`notice` was added as a loglevl in 4.1.0 (See https://github.com/npm/npmlog/commit/77b3b95c3474afdc164eae9caa80749452d42300). Here, I up the `npmlog` version to stop an error I've been getting. Should close #430, but I don't know how to test that.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3009,7 +3009,7 @@
     "npmlog": {
       "version": "4.1.2",
       "resolved": "https://registry.npmjs.org/npmlog/-/npmlog-4.1.2.tgz",
-      "integrity": "sha1-CKfyqL9zRgR3mp76StXMcXq7lUs=",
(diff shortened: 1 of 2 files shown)


gr2m:
hmm shouldn’t the `4.1.2` version be installed automatically as it is inside the range of `^4.0.0`?

vanduynslagerp:
I think it's not installed because of `package-lock.json`

gr2m:
I think it’s already set to 4.1.2 in there, it looks like `@semantic-release/last-release-npm"` is depending on an outdated version though, probably it’s the root of the problem

PR #442, opened 2017-09-21
Set minimum node version to 4

### LINKED ISSUE #433: Yarn fails to install semantic-release on node version < 8
Given that you only support node 8+, how do you suggest configuring `package.json` and `.travis.yml` such that there aren't dependency errors on node versions less than 8?

vanduynslagerp:
Avoid `yarn` to fail when installing on node `<8` with error `error semantic-release@8.0.0: The engine "node" is incompatible with this module. Expected version ">=8".
error Found incompatible module` when ran without `--ignore-engines`.

The library will never have to run on `node` `<8` as `semantic-release` will call it only on when running on `node` `>=8`.

All dependencies should have been modified to to use `"engines": {"node": ">=4"}`

Fix #433


hbetts:
This change seems odd. The whole point of `engines` is to say what version of Node is required for the successful operation of this package.

However, you're saying it works on Node 4, only to then turn around and fail later on.

vanduynslagerp:
Thanks for the feedback @hbetts!

The goal of `engines` is not that clear... According to npm documentation:
> If you specify an "engines" field, then npm will require that "node" be somewhere on that list. If "engines" is omitted, then npm will just assume that it works on node.

But it seems that npm actually doesn't care much and do not "require" anything.
Still in the same documentation:
> Unless the user has set the engine-strict config flag, this field is advisory only will produce warnings when your package is installed as a dependency.

I didn't check but I don't think it produces warnings.

And on the next line:
> engineStrict
This feature was removed in npm 3.0.0

So the `engines` config is:
- For npm: an **indication** on which node version can **run** the package
- For yarn: a **constraint** on which node version the package can be **installed**

A lot of confusion come from the fact that `semantic-release` throw errors for situation that are actually expected and not really errors. This is because we run `semantic-release pre && npm publish && semantic-release post` and the only way for `semantic-release pre` to prevent `npm publish && semantic-release post` to run is to exit the process with `1` which make npm to log an error.
We are currently working on a solution that would allow us to run only one command and would avoid to rely on this mechanism of throwing error to interrupt the command chain. That's not necessarily as easy as it sounds so it might takes some time to be done.

So in conclusion when you run `semantic-release` on node < 8 it throws an error but it's actually meant to be an informative message to the user. And `semantic-release` is designed so it can run on node < 8, the same way it can run Travis job that are not the build leader or on branches that are not `master`. In all these situations it will not trigger a release but it can and should be run on these situations, precisely to verify those conditions.

I think the best approach is:
- Go with `"engines": {"node": ">=4"}` right away to avoid pain for `yarn` users (and avoid them to turn away from `semantic-release`)
- Continue to work on a better solution to avoid throwing those "expected" errors to interrupt the command chain, and avoid the confusion regarding those error that are actually not errors

PR #451, opened 2017-09-29
ES6 migration, repo harminization

vanduynslagerp:
- ES 6 + AVA + test refactoring:
    - Use async/await instead of callbacks
    - Use [execa](https://github.com/sindresorhus/execa) to run command lines
    - Use AVA for tests
    - Add several assertions in existing unit tests
    - Add documentation (comments) in the tests
    - Run tests with a real git repo instead of mocking `child_process` and add test helpers to create repos, commits and checkout
    - Simplify test directory structure
    - Simplify code readability (mostly with async/await)
    - Switch from `standard` to `eslint` + `eslint-config-standard` + [prettier](https://github.com/prettier/prettier)
- Add a License file + badge
- Remove npm and yarn lockfile
- Use a more generic `.gitignore` that would work on Windows, Linux, Mac OS
- Add badges

The test coverage appear to be worse than before because `src/index.js` was not included in the coverage and it is now. The coverage is actually a bit better now.
I'll add more tests in a following PR.

Overall this PR doesn't change functionality, and does not fix bugs. It makes code and unit/integration tests more readable and easier to understand, hopefully making the project more accessible to contributors.

I found several improvement and bugs to fix will rewriting the tests. That will come in following PRs.

Fix #391, Fix #378, Fix #361


src/lib/get-commits.js
```diff
@@ -0,0 +1,57 @@
...
+    }
+  }
+
+  try {
+    ({stdout} = await execa('git', [
+      'log',
+      '-E',
+      '--format=%H==SPLIT==%B==END==',
+      `${lastRelease.gitHead ? lastRelease.gitHead + '..' : ''}HEAD`,
+    ]));
```
gr2m: do we need to wrap this block into `()`?

gr2m: I’ve seen it in a few other places, too

vanduynslagerp: Yes in order to reuse the `stdout` variable.
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Assignment_without_declaration

I could also use different variables.

.gitignore
```diff
@@ -1,9 +1,137 @@
...
+*.msp
+
+# Windows shortcuts
+*.lnk
+
+# End of https://www.gitignore.io/api/macos,windows,linux,node
+
+package-lock.json
+yarn.lock
+
```
luftywiranda13: ignoring lockfiles like this is harmful, because lockfiles are still in the repo (they're just **invisible**)
See https://github.com/luftywiranda13/remove-lockfiles#why

vanduynslagerp: What do you mean 'they're just invisible' ?
If they are not committed to the repo, they are not in the repo.

vanduynslagerp: Also I added [npmrc](https://github.com/semantic-release/semantic-release/pull/451/files#diff-794bdde32142176a26aef68e6770521eR1) and [.yarnrc](https://github.com/semantic-release/semantic-release/pull/451/files#diff-a3e03be55c12fe638fe1e1c5aa684b9aR1) to not regenerate them.

luftywiranda13: i updated my comment, sorry 😆

luftywiranda13:
> Also I added npmrc and .yarnrc to not regenerate them.

2 dotfiles to prevent lockfiles to be generated? I think it's overkill

.gitignore
luftywiranda13: > Also I added npmrc and .yarnrc to not regenerate them.

2 dotfiles to prevent lockfiles to be generated? I think it's overkill

package.json
```diff
@@ -20,38 +20,60 @@
...
+    "eslint-plugin-node": "^5.2.0",
+    "eslint-plugin-prettier": "^2.3.0",
+    "eslint-plugin-promise": "^3.5.0",
+    "eslint-plugin-standard": "^3.0.1",
     "nock": "^9.0.2",
     "npm-registry-couchapp": "^2.6.12",
-    "nyc": "^10.0.0",
+    "nyc": "^11.2.1",
+    "p-map-series": "^1.0.0",
+    "prettier": "^1.7.0",
```
luftywiranda13: Isn't it better to integrate prettier + eslint in `pre-commit hook`? 
So we can make sure prettier + eslint reformat any changes before committing

package.json
```diff
@@ -70,6 +92,23 @@
...
+      "html"
+    ],
+    "all": true
+  },
+  "prettier": {
+    "printWidth": 120,
+    "singleQuote": true,
+    "bracketSpacing": false,
+    "trailingComma": "es5"
+  },
```
luftywiranda13: IMO it's better to put prettier config directly inside eslint config. we're using `eslint-plugin-prettier`, right?

.gitignore
vanduynslagerp: The point of removing the lockfiles is to do the development with latest versions of dependencies. Your solution (unless I'm missing something) just remove the lockfiles before commit even though they wouldn't be committed anyway due to gitignore.

That doesn't solve the problem as the lockfile will be present on the local machine until the contributor do a commit. So a contributor would do all the development and run the test locally with potentially outdated libraries. Therefore what worked on their local machines might not work on the CI.

Also adding 2 extra dependency + a commit hook seem more overkill to me than two one line dotfiles.

package.json
luftywiranda13: Also this approach will reduce the possibility of us having code style discussion in code review
https://github.com/semantic-release/semantic-release/pull/451/files#r141981878

.gitignore
luftywiranda13: > The point of removing the lockfiles is to do the development with latest versions of dependencies.

yes, I know. that's the main reason why I decide to never commit lockfiles in my projects



> Also adding 2 extra dependency + a commit hook seem more overkill to me than two one line dotfiles.


hmmm I think https://github.com/semantic-release/semantic-release/pull/451#discussion_r141984300 is a relevant answer

package.json
vanduynslagerp: We had this discussion about commit hook previously among maintainer and we've decided to not use commit hook for simplicity. Also that could create confusion for contributors to see their code automatically changed on commit.

The chosen solution is to enforce the formating with eslint (with eslint-plugin-prettier).
All the code format pushed is validated in CI with eslint to be sure it's formatted with prettier.

This [comment](https://github.com/semantic-release/semantic-release/pull/451/files#r141981878) is not really about style but more about a not well know javascript feature allowing to reassign `let` variables with destructuring.

package.json
vanduynslagerp: No because prettier config is read by editor tools and by the `prettier` command itself.
Actually it work the other way around, it's `eslint-plugin-prettier` that read the prettier config. See prettier/eslint-plugin-prettier#55

src/lib/get-commits.js
gr2m: ah got it, it’s fine, code is readable as is, I was just curious. And learned something, thanks 👍

gr2m:
do you know what’s up with codecov?

src/lib/get-commits.js
vanduynslagerp: I'm thinking of doing a bit more refactoring in other PR actually. Mostly modularize a bit and splitting thing in different function.
That would make the code clearer, allow us to write more docs in the form of jsdoc.
Also that would simplify a lot the situation like were have to reuse variable.

Most of the case I do that is due to `try` / `catch` to avoid having all the code depending the throwing function be part of the try block.

With more modularization, each function that throws could handle it's errors (ie process the error and rethrow a SemanticRealseError) so in the parent module that calls them we could have only one `try` `catch` block.

vanduynslagerp:
> do you know what’s up with codecov?

Yes, it's because codecov has nothing to compare against to figure out if the coverage reduced as it has never ran on `caribou` branch.
All other PR after this one will work as expected.

But let me check if I can put a config to not report statuses when there is no comparison point.

package.json
luftywiranda13: > We had this discussion about commit hook previously among maintainer...

cool, I respect that 👍 

> The chosen solution is to enforce the formating with eslint (with eslint-plugin-prettier).
> All the code format pushed is validated in CI with eslint to be sure it's formatted with prettier.

yeah, the changes will failed on CI if they're not following our prettier config but then we still have to discuss about the code style things in the PR, right? isn't it not so good in terms of productivity?

src/lib/get-commits.js
gr2m: > I'm thinking of doing a bit more refactoring in other PR actually. Mostly modularize a bit and splitting thing in different function.
That would make the code clearer, allow us to write more docs in the form of jsdoc.
Also that would simplify a lot the situation like were have to reuse variable.

yeah +1 on doing that in follow up PRs, this one is quite big already

> With more modularization, each function that throws could handle it's errors (ie process the error and rethrow a SemanticRealseError) so in the parent module that calls them we could have only one try catch block.

Sounds good to me

package.json
luftywiranda13: > Actually it work the other way around, it's eslint-plugin-prettier that read the prettier config. See prettier/eslint-plugin-prettier#55

yeah, it's added in v2.3.0. but it still can read the config directly if we put the config in `rules` field

package.json
vanduynslagerp: Not really because `eslint` will run on `npm run test`. So a contributor would figure that out before committing.
Sure we could run `npm run test` on each commit or something like that but:
- Most contributor would know they have to at least run unit test before commiting
- For some less experienced contributors, we prefer answer them in the PR and help them as we believe guidance from a human would be more pleasant and helpful than an error message on commit

And there wouldn't be any style discussion, the style is enforced by eslint + prettier. There would be just explanation on how to read the eslint error, how to format properly, and maybe suggestion to use tools etc...Anything helpful for the contributor.

gr2m: they will also fail when you run `npm test` locally

package.json
vanduynslagerp: Prettier config in `prettier` property, Eslint config in `eslintConfig` property seems more clear to me. Not sure what would be the point of moving it around.
But thanks for the feedback :)

vanduynslagerp:
So there is a way to prevent codecov to fail on the first PR that use it: https://github.com/codecov/support/issues/173#issuecomment-216067683

But apparently it can also create situation in which Codecov would report success when coverage may be faulty.
As this would happen only on the first PR that configure Codecov, it's probably better to keep configured the way it is now.

gr2m:
yeah feel free to merge it as is any time, I just don’t do it myself because I’m not sure if you still fine tuning it :)

vanduynslagerp:
I'm fine with it as is. I'm gonna wait a few hours while I work on the following PR locally in case anyone as additional comments.

@luftywiranda13 Thanks for the review 👍 
If you have any more comments/review, feel free to add them :)

gr2m:
🙌

Issue #411, opened 2017-08-12
💌📯 Updates from semantic-release


gr2m:
👋 Hello friends, my name is Gregor and my friend Stephan @boennemann invited me as maintainer to semantic-release to help move it forward. Some of you might now me from [Hoodie](http://hood.ie/), that’s the project that semantic-release was originally developed for before it became it’s own thing. We use it in all our repositories there and I personally use `semantic-release` in _all_ of my personal projects. 

Honestly, without semantic-release, I would not be able to maintain nearly as many Open Source projects as I do today.

Please subscribe to this issue for updates. I will lock the issue to keep it for updates only, but trust me I know most of your pains myself and I hear you.

I will be very busy in August but will have more time starting September as I [transition into freelancing](https://twitter.com/gr2m/status/895690164168572928) :)

**If you like to help** <3 triaging the existing issues and pull-requests and pointing folks to this issue for updates would be great ✨ If you find issues that you think I should look into next, please mention me in them and I’ll try to have a look. 

Thank for your patience with me! Expect some great updates coming your way :)

gr2m:
The fix of the `"<a name"x.x.x"> in release notes"` bug is currently blocked by a problem with npm publish: https://gist.github.com/gr2m/80e7e580556c1f7eb32d7b8af55ff281

boennemann:
Sorry for neglecting semantic-release in the past and thanks @gr2m for stepping in.
Happy to see progress in here 💖

gr2m:
I’ve released `semantic-release@7.0.0` with the next release. You can install it with `npm install --save-dev semantic-release@next` or directly set the version in your projects’ `package.json` files. 

Note that support for no longer maintained Node.js versions has been dropped as part of this release. If you use Node 4 or newer all will be fine :) I’m very sorry if you still have run on Node 0.10 or Node 0.12, but we can’t support these any longer while at the same time making the project easier to maintain and contribute to.

This should resolve the `"<a name"x.x.x"> in release notes"` issue. Please give it a try and let me know if it works as expected, once a few folks tested it we can set the `semantic-release@7.0.0` to `@latest`

gr2m:
I have set `semantic-release@7.0.1` to `@latest`. Happy updating :)

gr2m:
[...]

PR #453, opened 2017-09-30
Retrieve version gitHead from git tags and unshallow the repo if necessary

### LINKED ISSUE #200: Can't parse change on Travis but can on local machine
Hi. If I'm running semantic-release locally it prints the following:

``` shell
 ~/L/P/cursor   master   npm run semantic-release

> kittik-cursor@0.0.0-semantic-release semantic-release /Users/ghaiklor/Library/Projects/cursor
> semantic-release pre && npm publish && semantic-release post

semantic-release WARN pre semantic-release didn’t run on Travis CI and therefore a new version won’t be published.
semantic-release WARN pre You can customize this behavior using "verifyConditions" plugins: git.io/sr-plugins
semantic-release ERR! pre Determined version 3.1.0 as "latest". Not publishing in debug mode. { type: 'minor', version: '3.1.0' }
```

But if running on Travis CI:

``` shell
$ npm run semantic-release
> kittik-cursor@0.0.0-semantic-release semantic-release /home/travis/build/kittikjs/cursor
> semantic-release pre && npm publish && semantic-release post
semantic-release ERR! commits The commit the last release of this package was derived from is not in the direct history of the "master" branch.
semantic-release ERR! commits This means semantic-release can not extract the commits between now and then.
semantic-release ERR! commits This is usually caused by force pushing, releasing from an unrelated branch, or using an already existing package name.
semantic-release ERR! commits You can recover from this error by publishing manually or restoring the commit "ab2b7d1bdee818855569eced566f8d59576a2eb2".
semantic-release ERR! pre ENOTINHISTORY Commit not in history
```

How can it be fixed ?

### LINKED ISSUE #229: Fetching latest release from npm is error prone
I ran into the problem, that the version was always calculated wrong (pretty much returning major in all cases).

I found out that for some reasons some releases did not hava a gitHead attribute set on npm (in my case version 2.2.1 of https://registry.npmjs.org/formulon).

It then fell back to a way earlier version all the time.

As a workaround I used the module [last-release-git-tag](https://github.com/semantic-release/last-release-git-tag) to fetch the latest version.

### LINKED ISSUE #256: Bump minor version when not needed

### LINKED ISSUE #276: Changes are not detected
semantic release does not detect changes for some reason:
https://travis-ci.org/travelperk/case-converter/jobs/153565553

As you can see, the commit which was build contained a `fix`. Still, semantic-release did not find any changes.

### LINKED ISSUE #280: Warn user about nonexistent gitHead returned by npm-registry-client

pvdlg:
Add several fixes and improvements in the identification of the last release gitHead:
- If there is no last release, unshallow the repo in order to retrieve all existing commits
- If git head is not present in last release, try to retrieve it from git tag with format 'v\<version\>' or '\<version\>'
- If the last release git head cannot be determined and  found in commit history, unshallow the repo and try again
- Throw a ENOGITHEAD error if the gitHead for the last release cannot be found in the npm metadata nor in the git tags, preventing to make release based on the all the commits in the repo as before
- Add integration test for the scenario with a packed repo from which `npm republish` fails to read the git head

Fix #447, Fix #393, Fix #280, Fix #276, Fix #256, Fix #229, Fix #200

Note on shallow repository:
On several CI platform the git repository is cloned as a shallow copy, with limit number of commits in the history and no tags. See [Travis](https://docs.travis-ci.com/user/customizing-the-build#Git-Clone-Depth) for example.
This could cause `ENOTINHISTORY` error when the git head of the last release is more than 50 commits behind the current head.
It also prevent to uses tags as they are not retrieved during the cloning.

With this PR, `semantic-release` will do as much as possible with the shallow clone and if it can't find the last release git head on the shallow it will call `git fetch --unshallow` and try again. This way the unshallow is done only when necessary. It's not necessarily a big operation but no reason to use bandwidth is we don't have too :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,10 +13,6 @@ branches:
     - caribou
     - /^greenkeeper.*$/
 
-# Retrieve 999 commits (default is 50) so semantic-release can analyze all commits when there is more than 50 on a PR
-git:
-  depth: 999
-
 # Retry install on fail to avoid failing a build on network/disk/external errors
 install:
   - travis_retry npm install
--- a/src/lib/get-commits.js
+++ b/src/lib/get-commits.js
@@ -1,56 +1,92 @@
 const execa = require('execa');
 const log = require('npmlog');
-const SemanticReleaseError = require('@semantic-release/error');
+const getVersionHead = require('./get-version-head');
 
-module.exports = async ({lastRelease, options}) => {
-  let stdout;
-  if (lastRelease.gitHead) {
+/**
+ * Commit message.
+ * 
+ * @typedef {Object} Commit
+ * @property {string} hash The commit hash.
+ * @property {string} message The commit message.
+ */
+
+/**
+ * Retrieve the list of commits on the current branch since the last released version, or all the commits of the current branch if there is no last released version.
+ * 
+ * The commit correspoding to the last released version is determined as follow:
+ * - Use `lastRelease.gitHead` is defined and present in `config.options.branch` history.
+ * - Search for a tag named `v<version>` or `<version>` and it's associated commit sha if present in `config.options.branch` history.
+ *
+ * If a commit corresponding to the last released is not found, unshallow the repository (as most CI create a shallow clone with limited number of commits and no tags) and try again.
+ *
+ * @param {Object} config
+ * @param {Object} config.lastRelease The lastRelease object obtained from the getLastRelease plugin.
+ * @param {string} [config.lastRelease.version] The version number of the last release.
+ * @param {string} [config.lastRelease.gitHead] The commit sha used to make the last release.
+ * @param {Object} config.options The semantic-relese options.
+ * @param {string} config.options.branch The branch to release from.
+ * 
+ * @return {Promise<Array<Commit>>} The list of commits on the branch `config.options.branch` since the last release.
+ * 
+ * @throws {SemanticReleaseError} with code `ENOTINHISTORY` if `config.lastRelease.gitHead` or the commit sha derived from `config.lastRelease.version` is not in the direct history of `config.options.branch`.
+ * @throws {SemanticReleaseError} with code `ENOGITHEAD` if `config.lastRelease.gitHead` is undefined and no commit sha can be found for the `config.lastRelease.version`.
+ */
+module.exports = async ({lastRelease: {version, gitHead}, options: {branch}}) => {
+  if (gitHead || version) {
     try {
-      ({stdout} = await execa('git', ['branch', '--no-color', '--contains', lastRelease.gitHead]));
+      gitHead = await getVersionHead(version, branch, gitHead);
     } catch (err) {
-      throw notInHistoryError(lastRelease.gitHead, options.branch);
+      // Unshallow the repository if the gitHead cannot be found and the branch for the last release version
+      await execa('git', ['fetch', '--unshallow', '--tags'], {reject: false});
     }
-    const branches = stdout
-      .split('\n')
-      .map(branch => branch.replace('*', '').trim())
-      .filter(branch => !!branch);
 
-    if (!branches.includes(options.branch)) {
-      throw notInHistoryError(lastRelease.gitHead, options.branch, branches);
+    // Try to find the gitHead on the branch again with an unshallowed repository
+    try {
+      gitHead = await getVersionHead(version, branch, gitHead);
+    } catch (err) {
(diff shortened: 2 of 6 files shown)


gr2m:
I tried it on my semantic-release test repo: https://travis-ci.org/gr2m/semantic-release-test#L879

I’ve `git push -f`ed to remove the git from the 2.0.3 release, and I recreated the v2.0.3 tag for the new commit. Shouldn’t that have worked? I must miss something

pvdlg:
That's expected with the current version of the code.
It's because you have a `gitHead` in the metadata of the last release's `package.json`. So I use that and I don't check the git tags.

I use the git tag as a fallback only if there is no `gitHead` in the last release's `package.json`.

Here we have a situation in which:
- `package.json` from npm version `2.0.3` reference commit A
- commit A is not in the history of the branch anymore
- The `v2.0.3` tag reference commit B
- Commit B is in the branch history

So we have 2 version markers (git tag and npm last release `package.json`) that contradict each other. It's probably a case that is quite rare and that is an indicator of a bigger problem in the repo.
So I though in this situation it's better to bail out with the `ENOTINHISTORY` error and let the user decide if they want to make a release manually.

In the end it's a question of who to trust more when we have contradicting infos between a git tag and npm metadata. As the creation of the `gitHead` is done automatically by `npm publish` and a tag can be more easily created by mistake, I thought we should give priority to npm metadata.

That said I can change the code to handle this situation. Do you see a scenario that will result in this situation ? It would imply to make a release for a given commit, remove the commit from the history, make a new commit and tag a version on this commit. It seems very wrong to change a commit that happened before a release.

gr2m:
> Do you see a scenario that will result in this situation ?

My main scenario would be a simpler recovery from the `ENOTINHISTORY` error. Right now I need to do a manual release. With that change, I could change an existing git tag to point to the correct git hash that is in the master branch and then restart the build to recover. 

But we should document that as well if we should support it. 

I get your point though, my use case is out of scope of this pull request, if you agree that it’s a good idea, we can still make a follow up PR for that

So from my side, I’d suggest to merge & release it, notify folks in #411 and ask them to install the newest version by updating their package.json files and comment on this PR if they run into any trouble, before setting the new version to latest?

pvdlg:
> My main scenario would be a simpler recovery from the ENOTINHISTORY error. Right now I need to do a manual release. With that change, I could change an existing git tag to point to the correct git hash that is in the master branch and then restart the build to recover.

I didn't think about that! It's indeed a great idea for a recovery scenario, that is much convenient than doing a manual release.
So in the `ENOTINHISTORY` message instead of suggesting to do a manual release we could suggest to tag the commit that correspond to the last release and restart the build. I'd have to change the algorithm a bit to check the tag in the case a `gitHead` is in the npm metadata but not in the branch history.

> I get your point though, my use case is out of scope of this pull request, if you agree that it’s a good idea, we can still make a follow up PR for that

Yep. Let's take a couple days to think about it and be sure we won't create side issues in some scenario. I don't see any one so far, but it's a complex problem...

> So from my side, I’d suggest to merge & release it, notify folks in #411 and ask them to install the newest version by updating their package.json files and comment on this PR if they run into any trouble, before setting the new version to latest?

I'll do that.

gr2m:
> So in the ENOTINHISTORY message instead of suggesting to do a manual release we could suggest to tag the commit that correspond to the last release and restart the build.

yeah that’d be great

PR #457, opened 2017-10-03
feat: allow pre-releases on GitHub

Apidcloud:
As discussed back in https://github.com/semantic-release/semantic-release/pull/431, this PR should allow pre-release on GitHub whenever `publishConfig tag` is manually set and `release no-pre-release` doesn't contain or isn't equal to that tag.

I tried to unit test all possible combinations. I'm unable to test it on another project as travis seems to be caching the dependencies for some reason.

Please give it a try and let me know.


src/post.js
```diff
@@ -10,12 +10,42 @@ module.exports = async config => {
...
+    repo,
+    tag_name: name,
+    name,
+    target_commitish: branch,
+    draft: !!debug,
+    body,
+    prerelease: githubPreRelease,
+  };
 
   if (debug && !githubToken) {
```
Apidcloud: maybe this should come after deciding whether it's a pre-release or not

Apidcloud: since we are still deciding the value of a release property

src/post.js
```diff
@@ -10,12 +10,42 @@ module.exports = async config => {
...
     return {published: false, release};
   }
 
+  // obtain npm dist tag if any
```
pvdlg: The tests in `post.test.js` are fairly complex (use `nock`, call a bunch of submodules etc), so it would probably better to externalize the determination of the `prerelease` value to a sub-module in `src/lib`. You have seral example there.
This way you could create a test file dedicated to this sub-module that would be simpler and more focused. This way you would test there all the details regarding the determination of `prerelease` (default parameters, parameter with a value, dist-tag with value x, y, z etc) and in `post.test.js` you would just need one test that will only validate that whatever value returned by the sub-module is properly sent to gihub and that `post.js` pass the correct parameter to the sub-module.

src/post.js
```diff
@@ -10,12 +10,42 @@ module.exports = async config => {
...
+  let npmDistTag = (pkg.publishConfig || {}).tag;
```
pvdlg: The tag can also be in the `.npmrc` config.
It actually already determined in [src/index.js#L54](https://github.com/semantic-release/semantic-release/blob/caribou/src/index.js#L54) and passed to `post.js` in `config.npm.tag` so you should reuse that.

src/post.js
```diff
@@ -10,12 +10,42 @@ module.exports = async config => {
...
+
+  // if npmDistTag is set, we can consider github pre release
+  if (npmDistTag != null) {
```
pvdlg: As determined in [src/index.js#L54](https://github.com/semantic-release/semantic-release/blob/caribou/src/index.js#L54) `config.npm.tag` cannot be `null`;

src/post.js
```diff
@@ -10,12 +10,42 @@ module.exports = async config => {
...
+    // obtain release.no-pre-release config
+    const noPreReleaseTags = (pkg.release || {})['no-pre-release'] || ['latest'];
```
pvdlg: You could use Object destructuring to get the value of `pkg.release` in [src/post.js#L8](https://github.com/semantic-release/semantic-release/blob/caribou/src/post.js#L8) like it's done for other input params.

src/post.js
```diff
@@ -10,12 +10,42 @@ module.exports = async config => {
...
+
+    // make sure no-pre-release isn't the same as npm manually set dist tag,
+    // nor includes it when set as an Array
+    if (
+      (typeof noPreReleaseTags === 'string' && noPreReleaseTags !== npmDistTag) ||
+      (Array.isArray(noPreReleaseTags) && noPreReleaseTags.indexOf(npmDistTag) === -1)
```
pvdlg: You can use [includes()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) here.

src/post.js
pvdlg: Yes :) I think made a review comment in that direction.

src/post.js
Apidcloud: How do I check whether it was manually set or not in this case?

pvdlg: Maybe I'm missing, but what difference does it make if the user explicitly set the dist-tag as `latest` in `publishConfig` or in the code in `index.js` set it as it's default value of `latest` ?
If it's `latest` then no Github pre-release, independently of how it was set isn't it?

Apidcloud: Thought you mentioned that it shouldn't pre-release when set to 'latest' **by default**.

pvdlg: I might have express myself in a confusing way, sorry.
The algorithm I think should be something along those line:
- if dist-tag (default or not) is `latest` or in `no-prerelease` property => no pre-release
- Otherwise => pre-release

src/post.js
Apidcloud: Not sure how to do it here. Can you help?

pvdlg: I was thinking something like: `const {pkg: {release} = {}, options: {branch, debug, githubUrl, githubToken, githubApiPathPrefix}, plugins} = config;`
But actually, I just realized `pkg` is used somewhere else so nevermind.

src/post.js
Apidcloud: I'm trying that at the moment. Is there any good way of debugging these tests? Or at least a way of printing something to the console.

pvdlg: `console.log` would print to the console when you run the tests.
You can run `ava -v --watch` so each test of source file you modify will immediately run so that make it easier.
You can also temporarily add `only` to your test (for example `test.only('test title' ... )` or `test.serial.only('test title' ... )`), so only this test will run.
You can find more info on Ava here https://github.com/avajs/ava

Apidcloud:
I think it should be much cleaner now 👍

All there's left is to check whether the module is called correctly within post. How should I do that? I wonder if I can do it with spies. Any suggestions?

test/post.test.js
```diff
@@ -146,6 +182,75 @@ test.serial('Post dry run without github token', async t => {
...
+      name: tagName,
+      target_commitish: branch,
+      draft: debug,
+      body: releaseLog,
+      prerelease: expectPreRelease,
+    },
   });
 });
+
+test.serial.only('Post dry run with pre-release', async t => {
```
pvdlg: You forgot to remove the `only` here.

src/lib/get-pre-release.js
```diff
@@ -0,0 +1,21 @@
+module.exports = config => {
+  const {pkg: {release = {}}, npm: {tag}} = config;
```
pvdlg: You could do directly:
```js
module.exports =({pkg: {release = {}}, npm: {tag}}) => {
```
And remove `const {pkg: {release = {}}, npm: {tag}} = config;`

test/post.test.js
```diff
@@ -38,7 +39,12 @@ test.serial('Post run with github token', async t => {
   const debug = false;
   const tagName = `v${version}`;
   const options = {branch, debug, githubUrl, githubToken, githubApiPathPrefix};
-  const pkg = {version, repository: {url: `git+https://othertesturl.com/${owner}/${repo}.git`}};
+  const release = {};
+  const npm = {
```
pvdlg: It's a small detail, but for readability and consistency, could you format the objects with few properties on one line? In other places as well?

test/post.test.js
```diff
@@ -64,7 +71,16 @@ test.serial('Post run with github token', async t => {
...
-    release: {owner, repo, tag_name: tagName, name: tagName, target_commitish: branch, draft: debug, body: releaseLog},
+    release: {
+      owner,
+      repo,
+      tag_name: tagName,
+      name: tagName,
+      target_commitish: branch,
+      draft: debug,
+      body: releaseLog,
+      prerelease: expectPreRelease,
```
pvdlg: Can you rename `expectPreRelease` into `prerelease` ?
It's an expected value [there](https://github.com/semantic-release/semantic-release/pull/457/files#diff-0febbd8a0bc438ee56826534d7358b1cR57) but it's the "obtain" value here, in the sense it's what the github API client sent. So better to call it just `prerelease`.

pvdlg:
Great job! Especially on the tests !
I added just a few change requests for minor stuff.

The last step would be to add the new option with a short description [here](https://github.com/semantic-release/semantic-release#options)

If you could squash your commits into 1 that would be awesome. But I can do it on my side I think.

pvdlg:
> All there's left is to check whether the module is called correctly within post. How should I do that? I wonder if I can do it with spies. Any suggestions?

It's done with the tests you modified and the one you added on `post.test.js`. 👍

Apidcloud:
I have addressed (all of?) the minor formatting issues you mentioned, added a description of this new option to the documentation and squashed everything into 1 commit.

Were you able to test this on a real repository? I _believe_ it works, but travis starting caching my fork project and was unable to test further.

Let me know if everything's ok 👍

README.md
```diff
@@ -154,6 +154,7 @@ You can pass options either via command line (in [kebab-case](https://lodash.com
 
 These options are currently available:
 - `branch`: The branch on which releases should happen. Default: `'master'`
+- `no-pre-release`: The npm dist tags that shouldn't be pre-released on GitHub. Default: `['latest']`. Can also be set to a single string. Note that if `publishConfig.tag` value is **included within** this option **or set to** `latest` (it is by default) a **pre-release won't happen**
```
pvdlg: I'm afraid that a **pre-release won't happen** might give the idea that the release will not be done.

Maybe something as follow instead ?

`no-pre-release`: By default, releases on an npm dist-tag other than `latest` will produce a [Github release](https://help.github.com/articles/creating-releases) with the pre-release tag. You can add any dist-tag in the `no-pre-release` array in order to produce a regular (non pre-release) on Github for those dist-tag.

pvdlg:
Just a small change in terms of wording. Other than everything looks good ! 👍 
I'll give it a try on a test repo to be sure and I'll let you know.

Apidcloud:
Alright 👍

pvdlg:
Alright, everything looks good to me: https://github.com/pvdlg/test-sr/releases

Thanks for the great work !

Apidcloud:
Can you test going out of pre-release again? That's the one thing I wasn't able to. 

Thanks for the opportunity and advice 👍

pvdlg:
There is some edge cases scenario that doesn't work yet, but it's independent of this PR.

For example:

1. Publish 1.0.0 on next
2. Publish 1.1.0 on next
3. Make 1.0.0 available on latest
4. Publish a minor version on latest => `semantic-release` determines the new version should be `1.1.0` => Fails as the version already exists

That is a bigger issue we have to solve in `semantic-release`. I'm not sure yet what the best solution is.
We can either skip existing version and in step 4 release a 1.1.1 instead of 1.1.0. That avoid the error but create a situation in which the last release on latest is higher than the last version on next. It's not necessarily a problem (in the technical sense of the term) but it probably reveal a mistake in the way the release flow was handled.

A second solution would be to determine what is the last version in a more complex way:
- Use last version on the dist-tag if increasing this version number doesn't create a version higher or equal than a version on a higher dist-tag
- If it does, use the last version of the higher dist-tag that create a conflict

So in this situation on step 4 we would determine that the last release is 1.0.0, increase it to 1.1.0, see that it's higher or equal to the last version on next, therefore consider the last version on next as the last release and increase it to obtain 1.2.0. But that can creates a lot of issues if for example the release 1.1.0 has been done from a different branch and contains code not yet merge in the current branch. We would have 1.2.0 that contains less features than 1.1.0.

A third solution is to prevent a release that would end with a lower dist-tag having a higher last release than a higher dist-tag. So in the case of step 4 we would not do a release at all and display a message like "Doing this release would make the last version on 'latest' higher than the last version on 'next'."
That would work if we accept to limit the type of changes we can make on each dist-tags. For example in a situation with last release on latest being 1.0.0, and last release on next being 2.0.0, we would accept only patch and minor releases on latest.

We don't enforce anything regarding those scenario so far and we rely on the users to make appropriate type of changes in such configuration.

Apidcloud:
Makes sense. The third solution might be easier/better (at least for now) as it gives that responsibility to the users to change from other dist tags to latest before publishing a new version on npm.

README.md
```diff
@@ -154,6 +154,7 @@ You can pass options either via command line (in [kebab-case](https://lodash.com
 
 These options are currently available:
 - `branch`: The branch on which releases should happen. Default: `'master'`
+- `no-pre-release`: By default, the releases with a npm dist-tag other than `latest` will produce a [Github release](https://help.github.com/articles/creating-releases/) with the __pre-release__ tag. You can add any dist-tag in the `no-pre-release` array in order to produce a regular release (non pre-release) on Github for those dist-tag. Default: `['latest']`
```
gr2m: @pvdlg @Apidcloud are you sure we need that flag? What are current use cases for it? I'm reluctant to add more configuration if we can avoid it

src/lib/get-pre-release.js
```diff
@@ -0,0 +1,19 @@
+module.exports = ({pkg: {release = {}}, npm: {tag}}) => {
```
gr2m: I wonder if `is-pre-release.js` would be a better name for the file?

README.md
pvdlg: The use case is to force a regular release on Github, even if the release is done on a dist-tag other than `latest`.
Basically when you release on `v2` but you still want the release on Github to not be tagged `pre-release`.

Might be useful when you release on a legacy/LTS/old version dist-tag.

test/post.test.js
```diff
@@ -38,7 +39,10 @@ test.serial('Post run with github token', async t => {
   const debug = false;
   const tagName = `v${version}`;
   const options = {branch, debug, githubUrl, githubToken, githubApiPathPrefix};
-  const pkg = {version, repository: {url: `git+https://othertesturl.com/${owner}/${repo}.git`}};
+  const release = {};
+  const npm = {tag: 'latest'};
+  const pkg = {version, repository: {url: `git+https://othertesturl.com/${owner}/${repo}.git`}, release};
+  const prerelease = false;
```
gr2m: camelCase

test/post.test.js
```diff
@@ -87,7 +101,10 @@ test.serial('Post dry run with github token', async t => {
   const debug = true;
   const tagName = `v${version}`;
   const options = {branch, debug, githubToken};
-  const pkg = {version, repository: {url: `git+https://othertesturl.com/${owner}/${repo}.git`}};
+  const release = {};
+  const npm = {tag: 'latest'};
+  const pkg = {version, repository: {url: `git+https://othertesturl.com/${owner}/${repo}.git`}, release};
+  const prerelease = false;
```
gr2m: camelCase

test/post.test.js
```diff
@@ -132,11 +159,14 @@ test.serial('Post dry run without github token', async t => {
   const branch = 'master';
   const debug = true;
   const tagName = `v${version}`;
+  const release = {};
+  const npm = {tag: 'latest'};
   const options = {branch, debug};
-  const pkg = {version, repository: {url: `git+https://othertesturl.com/${owner}/${repo}.git`}};
+  const pkg = {version, repository: {url: `git+https://othertesturl.com/${owner}/${repo}.git`}, release};
+  const prerelease = false;
```
gr2m: camelCase

test/post.test.js
```diff
@@ -146,6 +176,73 @@ test.serial('Post dry run without github token', async t => {
...
+  const generateNotes = stub().resolves(releaseLog);
+  const version = '1.0.0';
+  const branch = 'master';
+  const debug = true;
+  const tagName = `v${version}`;
+  const options = {branch, debug, githubToken};
+  const release = {};
+  const npm = {tag: 'tagOtherThanLatest'};
+  const pkg = {version, repository: {url: `git+https://othertesturl.com/${owner}/${repo}.git`}, release};
+  const prerelease = getPreRelease({pkg, npm});
```
gr2m: camelCase

src/post.js
```diff
@@ -3,14 +3,16 @@ const url = require('url');
...
+const getPreRelease = require('./lib/get-pre-release');
 
 module.exports = async config => {
   const {pkg, options: {branch, debug, githubUrl, githubToken, githubApiPathPrefix}, plugins} = config;
   const [owner, repo] = parseSlug(pkg.repository.url);
   const name = `v${pkg.version}`;
   const tag = {owner, repo, ref: `refs/tags/${name}`, sha: await promisify(gitHead)()};
   const body = await promisify(plugins.generateNotes)(config);
-  const release = {owner, repo, tag_name: name, name, target_commitish: branch, draft: !!debug, body};
+  const prerelease = getPreRelease(config);
```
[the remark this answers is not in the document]
pvdlg: I think the `prerelease` is not on camel case so the variable can be used as a short and in `const release`.
`release` is passed to the Github API which expect `prerelease` not `preRelease`. See https://developer.github.com/v3/repos/releases/#create-a-release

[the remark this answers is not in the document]
gr2m: ah got it. It's tricky though, I think I would make it explicit in the argument passed to github using `prerelease: preRelease` to keep our naming of the variables coherent?

README.md
gr2m: > Might be useful when you release on a legacy/LTS/old version dist-tag.

I think this is a separate use case that we should address separately. My understanding is that this PR is all about setting the pre-release flag on releases when publishing to a custom tag?

We don't really support LTS releases yet, do we? That is a valid use case by itself but I think it will need other features like a built-in restriction of not allowing breaking feature releases and using a major version as base, so that even if v3.0.0 is already released, it can still release v2.0.1

gr2m:
I've tested it and it works as advertised 👍 https://github.com/gr2m/semantic-release-cli-test/releases

Once we make pre-releases on GitHub a first-class citizen feature, how would we recommend to keep releases on GitHub in sync with releases on npm? E.g. ideally, the moment I'd remove the pre-release flag, the release should be set to latest on npm? Maybe we could create a GitHub app for that which also gets setup when using `semantic-release-cli setup`? Or we add that as a feature to `semantic-release-cli`, e.g. `semantic-release-cli set-latest 1.2.3`?

README.md
pvdlg: > We don't really support LTS releases yet, do we?

Kind of. You can have a branch configured to release on the dist-tag `lts` for example. In that case you might not want the prerelease tag.
But yes in order to support it _properly_  we would need a to implement a verification to make sure we don't release something out of bound.

So I guess when we do that, we would have everything necessary to figure out it a LTS branch therefore the `prerelease` tag should be `false`.

In that case yes it makes the `no-pre-release` param useless.

Kind of feel bad that I asked @Apidcloud to add it...

src/post.js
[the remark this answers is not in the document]
pvdlg: Ok make sense 👍

pvdlg:
> Maybe we could create a GitHub app for that which also gets setup when using semantic-release-cli setup?

That sounds a really good idea !

README.md
gr2m: > Kind of feel bad that I asked @Apidcloud to add it...

I know, I understand, but remember Open Source is free as in puppy, once we take it, we will have to maintain it until the end of time :) Better have a longer, uncomfortable discussion upfront, that's how it goes some times, time gets spent on work only to realize it won't get merged. Still better than the alternative of accepting changes and than maintaining them for forever ;)

gr2m: also just to make that clear, I want to make LTS releases a first-class feature of semantic-release, so we can leave it safely out of scope for this feature

gr2m:
> > Maybe we could create a GitHub app for that which also gets setup when using semantic-release-cli setup?

> That sounds a really good idea !

I'll work on this just for fun and to explore how github apps could work against the npm registry API

PatrickHeneise:
Is there a possible workaround to get the gh-app running without the update endpoint?

gr2m:
@PatrickHeneise I can’t think of one. I think the only approach would be to add more features to the CLI as suggested above, e.g. `semantic-release-cli set-latest 1.2.3`, happy to get you going if you want to help with that

PatrickHeneise:
Not sure, we run semantic-release on travis, the CLI doesn't help much there. There's just a conflict between `master` and `develop`, I'd like to see all releases on `develop` on the `dev` pre-release tag, without changing the package.json manually.

gr2m:
> I'd like to see all releases on develop on the dev pre-release tag, without changing the package.json manually.

I would suggest to suggest to configure the release tag to to `dev` in general, so all releases will be using that tag. When you merge to master, no release will happen because there are no changes since the last release. Then once you are confident with the new version (e.g. `1.2.3`), you’d be able to run 

```
semantic-release-cli set-latest 1.2.3
```

PatrickHeneise:
If there's a way to not set the version number, that could work for us.

ie. `semantic-release deploy-latest`

Apidcloud:
Sorry for the delay guys. I'll try to do the changes on the weekend 😄

Apidcloud:
Finally made the changes 😅. Let me know if I can squash the commits and/or rebase caribou branch.

pvdlg:
I completely missed your previous message...sorry. The Github release has been moved to https://github.com/semantic-release/github.

The Github `prerelease` is something we want to support, but it's probably going to be done within a larger feature to handle multiple repo and dist-channel. I'm not 100% sure how everything is going to work though. I've been working on a proposal for some weeks now and did a lot of tests to see what's possible. I think I should be able to write a proposal for the implementation in a couple weeks.

Then we can discuss and see how and where to implement this feature. Stay tuned!

Apidcloud:
No problem 😄 Let me know how it goes!

pvdlg:
To be implemented in the `github` plugin as part of #563. Keeping open as some code might reused in the `github` plugin

pvdlg:
I'm closing as the code changed so much since this PR was opened.
The feature will be implemented via #563.

PR #463, opened 2017-10-06
Allow custom plugins when running globally

pvdlg:
Previously the plugins were imported relatively to `process.cwd()`.
If a custom plugin is defined in `package.json` via a globally installed npm package it would fails even when `semantic-release` is installed and ran globally.

This PR allow to load plugin:
- As an npm packaged depended on in `package.json` when run locally or globally
- A `.js` file, referenced relatively to the root of the repo, when run locally or globally
- A globally installed npm package when run globally

`semantic-release` can now be installed and ran globally, along with globally installed custom plugins without having to be included as a dependency in the `package.json`. It might be convenient for a future Github app :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "fs-extra": "^4.0.2",
     "git-head": "^1.2.1",
     "github": "^11.0.0",
+    "import-from": "^2.1.0",
     "lodash": "^4.0.0",
     "nerf-dart": "^1.0.0",
     "nopt": "^4.0.0",
@@ -32,7 +33,6 @@
     "npmlog": "^4.0.0",
     "p-series": "^1.0.0",
     "parse-github-repo-url": "^1.3.0",
-    "require-relative": "^0.8.7",
     "semver": "^5.4.1"
   },
   "devDependencies": {
--- a/src/lib/plugins.js
+++ b/src/lib/plugins.js
@@ -1,5 +1,5 @@
 const {promisify} = require('util');
-const relative = require('require-relative');
+const importFrom = require('import-from');
 const pSeries = require('p-series');
 
 module.exports = options => {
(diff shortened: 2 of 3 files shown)


gr2m:
Is this fixing an open issue that we have? I don’t quite understand the use case yet :) Can you explain a bit further?

pvdlg:
No we don't have an open issue regarding that, but it was mentioned in several conversation that it would be nice to be able to run `semantic-release` globally without having to depend on it in the `package.json`

What this change allow to do is to have in package.json:
```json
"release": {
  "analyzeCommits": "my-custom-analyzer"
}
```
and no dependencies to `semantic-release` or `my-custom-analyzer` in `devDependencies` and be able to run:
```bash
npm install -g semantic-release my-custom-analyzer
semantic-release pre && npm publish && semantic-release post
```

One possible benefit would be to not have the problem with `yarn` that refuses to install `semantic-release` as a `devDependency` on version of Node < 8 and would avoid to have to do the trick with `"engines": {"node": "4"}`.

Another possible advantage is the situation in which a release is triggered externally (for example when a Github app would detect via commit status that all CI pass successfully). You can trigger a job that install `semantic-release` globally and run. So for the user no need to include and manage `semantic-release` and custom plugins as a `devDependency`.

In conclusion, this is useful only when we run `semantic-release` globally and it's something that we don't officially recommend in the doc for now. But it might be useful in the future.

gr2m:
I fear that we introduce a functionality which we don’t yet have use cases for and that we will have to maintain for forever. Maybe let’s leave that open until we hit a blocker that this is the right solution for?

pvdlg:
Ok, works for me

pvdlg:
Irrelevant due to #524

Issue #406, opened 2017-07-30
npm ERR! "You cannot publish over the previously published version 1.1.0." 


dsslimshaddy:
Im running this. when i commit `"fix: some msg"` it always publishes the version specified in `package.json` and does not automatically changes version.

```
$ semantic-release pre && npm publish ./build && semantic-release post 

```

vanduynslagerp:
Could you provide a link to your `package.json`, `travis.yml` and the log of your CI that fail to publish ?

yoitsro:
I'm also getting this, but using Codeship. I also get the same even when running locally. Here's my setup:

package.json
```
{
  "name": "hops",
  "description": "Hops",
  "main": "index.js",
  "dependencies": {
    "google-protobuf": "^3.4.0"
  },
  "scripts": {
    "test": "exit 0",
    "semantic-release": "semantic-release pre && npm publish && semantic-release post"
  },
  "repository": {
    "type": "git",
    "url": "https://github.com/private-org/hops.git"
  },
  "author": "Ro Ramtohul",
  "license": "ISC",
  "bugs": {
    "url": "https://github.com/private-org/hops/issues"
  },
  "homepage": "https://github.com/private-org/hops#readme",
  "publishConfig": {
    "access": "restricted"
  },
  "engines": {
    "node": "> 8.4.0"
  },
  "version": "0.0.1",
  "devDependencies": {
    "@semantic-release/condition-codeship": "^1.1.0",
    "semantic-release": "^8.2.0"
  },
  "release": {
    "verifyConditions": [
      "@semantic-release/condition-codeship"
    ]
  }
}
```

I have the following Codeship environment variables set: `CI=true`, `GH_TOKEN` and `NPM_TOKEN`.

In the custom deployment script:
```
npm run semantic-release
```

pvdlg:
The problem happen when the next release determined by `semantic-release` already exists on the npm registry or has existed and has been unpublished.

In order to find a solution we need to figure out:
- The next release determined by `semantic-release`
- The release currently present on the npm registry
- Find out why the next release is already present on the npm registry

So ideally we need:
- The `semantic-release` logs for the build that failed
- The commit history of the project
- The metadata of the package on the npm registry (obtained with `npm view <package-name>`)

yoitsro:
So I'm doing some more digging and it looks like the `@semantic-release/condition-codeship` isn't working properly. Rather than bailing out like `@semantic-release/condition-travis` does, I get this warning:

```
[...]

PR #473, opened 2017-10-18
Handle errors when verifyConditions and verifyRelease are a pipeline

pvdlg:
The plugin module for `verifyConditions` and `verifyRelease` was returning:
- A `Promise` returning function if multiple plugins are defined
- The plugin itself (a function with callback) when there is only one plugin in the chain

The rest of the code expected `verifyConditions` and `verifyRelease` to be a function with callback and was promisifying it.
If there was multiple plugins in the pipeline that that was resulting in promisifying a promise and the errors returned were not caught and node was warning with ` UnhandledPromiseRejectionWarning: Unhandled promise rejection`.

With this fix the plugin module always return a `Promise` returning function for `verifyConditions` and `verifyRelease` and the rest of the code doesn't promisify them anymore.

See #406

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -75,7 +75,7 @@ module.exports = async () => {
     log.verbose('pre', 'Running pre-script.');
     log.verbose('pre', 'Veriying conditions.');
     try {
-      await promisify(plugins.verifyConditions)(config);
+      await plugins.verifyConditions(config);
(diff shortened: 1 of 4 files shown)


gr2m:
Thanks @pvdlg great work 👍

PR #475, opened 2017-10-19
Add error messages for unknown rejections

wejendorp:
I found the CLI extremely hard to use when writing custom plugins, and debugging various access errors when running it on our internal CI servers instead of Travis.

I will therefore propose to output error information when unexpected errors (non `@semantic-release/error`) occur, instead of exiting silently.

For example:
```json
{
    "release": { "verifyConditions": "./some-non-existing-file" }
}
```
Caused me a lot of grief, and I could not figure out why nothing happened.

### DIFF
--- a/bin/semantic-release.js
+++ b/bin/semantic-release.js
@@ -23,6 +23,8 @@ npx is bundled with npm >= 5.4, or available via npm. More info: npm.im/npx`
 }
 
 // node 8+ from this point on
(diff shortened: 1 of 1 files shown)


pvdlg:
Could change your commit to confirm with Angular convention ?
In this case: `fix: Log error messages on reject`

wejendorp:
Yes no problem.

PR #210, opened 2016-03-31
Allow running semantic release from node

meldiner:

Issue #465, opened 2017-10-08
use without npm


nxmohamad:
I'm a bit confused about how/if I can use this without publishing to NPM. I'm using this for a private project. The first release seemed to work fine 1.0.0, but then 4 commits later, there were still no releases made even after successful CI builds.

Is this because I have nothing published on NPM? Does it refer to the version number in NPM to generate the latest version number? If so, how can I get around this?

UPDATE: I've modified my setup to use a getLastRelease plugin that gets it from git tags instead. I still get the same thing, after the first publication of a release, it shows 'x commits have been made on master since this release' but no new releases are made. the versions are all just on 1.0.0 or 2.0.0.
after using the plugin the version number was bumped to 2.0.0.

pvdlg:
Currently `semantic-release` doesn't support a scenario without npm by default. This is something we want to support in the future.

If you provide the link of a CI job where you experience the problem I can look at it and try to help.

nxmohamad:
So the subsequent release is dependent on it being published on NPM? Is it right that semantic release would push a release on github on every commit? Or how does it calculate when to release?

felixfbecker:
I'd like to use semantic-release with PHP projects that use Composer. Composer doesn't publish tarballs of new versions, it just uses git tags in the repository. So I only need the functionality of semantic-release to tag new releases, i.e. instead of
```sh
semantic-release pre && npm publish && semantic-release post
```
just do
```sh
semantic-release pre && semantic-release post
```
without an NPM_TOKEN, just a GH_TOKEN.

felixfbecker:
Seems like this should work: https://www.npmjs.com/package/last-release-git

gr2m:
We are also discussing a GitHub app which you can install on any repository and it will do language agnostic releases to GitHub only, stay tuned :)

felixfbecker:
That sounds great, as long as it's still flexible enough to configure with custom scripts etc.

Got it to work btw: https://github.com/felixfbecker/php-semantic-release-test/releases/tag/v2.1.0

destroyerofbuilds:
> We are also discussing a GitHub app which you can install on any repository and it will do language agnostic releases to GitHub only, stay tuned :)

That's exciting to hear! Looking forward to giving that app a try.

destroyerofbuilds:
Actually, to move this discussion forward, could `npm` support be removed from `semantic-release`, and we just encourage developers to run `npm publish` as part of a build job that only runs on new tags`?

[...]

PR #480, opened 2017-10-25
Refactor CLI to run with one command, improve logs, modularize, add tests

### LINKED ISSUE #175: semantic-release pre: output preview for release notes
Would be great to see preview for release notes in console output when running locally.

### LINKED ISSUE #176: deprecated npmconf?

### LINKED ISSUE #239: Pass Commits to generateNotes function
I wrote a custom notes generator, expecting that like custom analyzer, that the commit messages would be passed in. It seems (to me) logical that if one were to generate release notes, they'd want the commits available. Alas, those commits are not available, and I've got to do some hackie things to make them available.

I'd like to request `commits` be passed in to the `generateNotes` function to allow custom release note generation to do so off of the commits provided.

### LINKED ISSUE #340: better logging level and exits status
some log messages don't need to be displayed as an `error` and certainly not with an failed exit status, [for example this](https://github.com/semantic-release/semantic-release/blob/caribou/bin/semantic-release.js#L130-L131) could easily be output as  successful info level 

```js
log.info('pre', message + ' Not publishing in debug mode.', release)
process.exit(0)
```

[this section](https://github.com/semantic-release/semantic-release/blob/caribou/bin/semantic-release.js#L130-L131) also indiscriminately throws out errors, for example: `ENOCHANGE There are no relevant changes, so no new version is released.` and few others I encountered.

~~before I dive in and attempt a PR:~~ _(PR POC linked below)_ is there a reason for why these are `errors` with exit status of `1`? as oppose to `info` or `warn`, or even still an `error` but with a successful exit status?

This would make it significantly easier to use `semantic-release` in more complex build steps and detect success/failure of process, so operation is not disrupted and further tasks can be processed after the release.

### LINKED ISSUE #356: Support dry run on publish, regardless of branch?
would be nice to have something like `npm run semantic-release --dry-run` which could be run in the CI to see what version would be published, and if everything's set up correctly

### LINKED ISSUE #398: An in-range update of npmlog is breaking the build 🚨
## Version **4.1.1** of [npmlog](https://github.com/npm/npmlog) just got published.

<table>
  <tr>
    <th align=left>
      Branch
    </th>
    <td>
      <a href="/semantic-release/semantic-release/compare/greenkeeper%2Fnpmlog-4.1.1">Build failing 🚨</a>
    </td>
  </tr>
  <tr>
    <th align=left>
      Dependency
    </td>
    <td>
      npmlog
    </td>
  </tr>
  <tr>
    <th align=left>
      Current Version
    </td>
    <td>
      4.1.0
    </td>
  </tr>
  <tr>
    <th align=left>
      Type
    </td>
    <td>
      dependency
    </td>
  </tr>
</table>

This version is **covered** by your **current version range** and after updating it in your project **the build failed**.

npmlog is a direct dependency of this project **this is very likely breaking your project right now**. If other packages depend on you it’s very likely also breaking them.
I recommend you give this issue a very high priority. I’m sure you can resolve this :muscle:

<details>
<summary>Status Details</summary>

- ❌ **coverage/coveralls** Coverage pending from Coveralls.io [Details](https://coveralls.io/builds/12140297)
- ❌ **continuous-integration/travis-ci/push** The Travis CI build failed [Details](https://travis-ci.org/semantic-release/semantic-release/builds/247264858?utm_source=github_status&utm_medium=notification)
</details>

---

<details>
<summary>Commits</summary>
<p>The new version differs by 3 commits.</p>
<ul>
<li><a href="https://urls.greenkeeper.io/npm/npmlog/commit/a1093e72aebee4febc807a0a6fcece7e253c06c4"><code>a1093e7</code></a> <code>Only publish the minimum</code></li>
<li><a href="https://urls.greenkeeper.io/npm/npmlog/commit/cf2131b733b7529514e9a4977cdf67b450a62fa3"><code>cf2131b</code></a> <code>4.1.1</code></li>
<li><a href="https://urls.greenkeeper.io/npm/npmlog/commit/c07471288ebe3ea81b0d75e43c7edd584d85b8cf"><code>c074712</code></a> <code>Use Object.defineProperty to stringify stack (fixes #50)</code></li>
</ul>
<p>See the <a href="https://urls.greenkeeper.io/npm/npmlog/compare/46f88c4bcb35a39b22aa5b65d0958221dadb2bbe...a1093e72aebee4febc807a0a6fcece7e253c06c4">full diff</a></p>
</details>

<details>
<summary>Not sure how things should work exactly?</summary>

There is a collection of [frequently asked questions](https://greenkeeper.io/faq.html) and of course you may always [ask my humans](https://github.com/greenkeeperio/greenkeeper/issues/new).
</details>

---

Your [Greenkeeper](https://greenkeeper.io) Bot :palm_tree:

### LINKED ISSUE #400: Add a way to preview the changelog
It looks like it's possible to preview the new version number that will be created by running `semantic-release` locally or with the `--debug` option. But I also want to make sure I've written my commit messages in the right format, and it looks like there's no easy way to do that. In particular, I want to make sure BREAKING CHANGE lines are handled properly, and I'd much rather just try it out than read a bunch of docs or dig through the dependencies to find some more specific package to run manually.

In particular, some things that weren't obvious from the README:
* Where can the BREAKING CHANGE line be in the commit message? (I learned the hard way that it needs to go at the bottom, or else the release notes will have the rest of the commit message as a breaking change.)
* If I want to indicate multiple breaking changes in the same commit, can/should include multiple BREAKING CHANGE lines?
* If I want multiple paragraphs of breaking changes, what's the right way to format it?

(I don't really care so much about these specific questions; I'm mostly just saying that it would be useful to be able to experiment with commit messages ahead of time rather than finding out that I've released a new version with bad release notes.)

### LINKED ISSUE #407: Way to just print the (new) version
I've playing around with options for a little bit now and I'd like to have a way to just print the version that will be generated (even if it didn't change). But I need so many options and especially the `NPM_TOKEN` variable just doesn't make any sense at this point.

Regardless of how many options I provide I'm not able to print the version.

My latest try:

```shell
CIRCLE_BRANCH=master CIRCLECI=true NPM_TOKEN=fake CI=TRUE GH_TOKEN=<redacted> GH_REPOSITORY_URL=repo yarn semantic-release pre

semantic-release ERR! pre ENOCHANGE There are no relevant changes, so no new version is released.
error Command failed with exit code 1.
```

### LINKED ISSUE #430: Invalid config warning

### LINKED ISSUE #438: Improve plugin logs
Sometimes it's quite difficult to understand why a release is not happening because the default plugins are not logging.

For example it would be helpful to get logs for:
- The list of commit analyzed
- The type of release determined for each commit and info regarding why a release type is triggered
- What make the verification fails
- Which build is determined to be the leader

We could just add a logger to each plugin or we could pass a logger from `semantic-release` to each plugin in the `config` object.
The second option might be better so we can have more consistency in the logging.

It should solve #432

Thoughts @gr2m ?

### LINKED ISSUE #460: Pass lastRelease, nextRelease and commits to release-note-generator
The commit used for the last release is now determined based on the metadata in the npm registry or the git tag.
By default `conventional-changelog-core` use the last git tag.

That can create issues if there is discrepancies between the npm registry and the git tags, as the main `semantic-release` and the `release-notes-generator` plugin use different algorithm to determine the last release commit. Also there is no point in doing the job twice for determining the last release and retrieving the commits or figuring out the new version.

In addition passing `nextRelease` and `commits` might be useful for other plugin implementations.

### LINKED ISSUE #477: Errors from plugins are not logged very well
I tried to make semantic-release work with https://github.com/finom/last-release-git and hit this error in CI:
```
semantic-release ERR! pre Command failed: git show-ref --tags Error: Command failed: git show-ref --tags
semantic-release ERR! pre at checkExecSyncError (child_process.js:592:13)
semantic-release ERR! pre at execSync (child_process.js:632:13)
semantic-release ERR! pre at module.exports (/home/travis/build/felixfbecker/php-semantic-release-test/node_modules/last-release-git/index.js:13:16)
semantic-release ERR! pre at bound (internal/util.js:227:26)
semantic-release ERR! pre at module.exports (/home/travis/build/felixfbecker/php-semantic-release-test/node_modules/semantic-release/src/pre.js:11:54)
semantic-release ERR! pre at module.exports (/home/travis/build/felixfbecker/php-semantic-release-test/node_modules/semantic-release/src/index.js:108:44)
semantic-release ERR! pre at <anonymous>
```
The error comes from a call to `execSync('git show-ref --tags')`, which throws an error because git exited with a non-zero exit code.
This error is useless without any STDERR output from git, the exit code etc.
Child process errors attach all these as properties to the Error object. Calling `console.error()` on the error will not only print message an stack, but also all the metadata like stderr, stdout, exit code, etc. semantic-release seems to only print the message and stack though.

pvdlg:
## New features / Fixes
- Run with one command and do not rely on error exit codes to stop the process when a release is not necessary. Fix #340
- Break `index.js` in smaller modules in order to improve testability and simplify the code
- Add several missing unit and integration tests to reach 100% coverage
- Integration tests now test end to end, including publishing to Github (with http://www.mock-server.com on Docker)
- Use `tj/commander.js` to print an help message, verify and parse CLI arguments
- Semantic-release can now be called via Javascript API: `require('semantic-release')(options)`. Fix #210
- Remove npmlog dependency and add more log messages. Fix #398. Fix #477. Fix #430. Fix #432
- Logger is now passed to plugins. Fix #438
- Add debug logs with `visionmedia/debug`. `debug` is enabled for both semantic-release and plugins with `--debug`
- Use `kevva/npm-conf` in place of the deprecated `npm/npmconf`. Fix #176
- Pass lastRelease, nextRelease and commits to generate-notes plugin. Fix #239. Fix #460
- In dry-run mode, print the release note instead of publishing it to Github as draft, and skip the CI verifications. Fix #175. Fix #356. Fix #400. Fix #407
- The dry-run mode does not require npm and Github TOKEN to be set anymore and can be run locally

## Breaking changes
- Semantic-Release must now be executed with `semantic-release` instead of `semantic-release pre && npm publish && semantic-release post`.
- The `semantic-release` command now returns with exit code 0 on expected exception (no release has to be done, running on a PR, gitHead not found, other CI job failed etc...). It only returns with 1 when there is an unexpected error (code error in a plugin, plugin not found, git command cannot be run etc..).
- Calling the `semantic-release` command with unexpected argument(s) now exit with 1 and print an help message.
- Semantic-Release does not rely on `npmlog` anymore and the log level cannot be configured. Debug logs can be activated with CLI option `--debug` or with environment variable `DEBUG=semantic-release:*`
- The CLI options `--debug` doesn't enable the dry-run mode anymore but activate the debug logs. The dry run mode is now set with the CLI command `--dry-run` or `-d`.

## Publish on npm
In order to publish on `npm` registry I run `npm publish` with `execa`. Another option would have been to do that programmatically with [npm-registry-client#clientpublish](https://github.com/npm/npm-registry-client#clientpublishuri-params-cb) but that would come with several issues:
- We would still have to run `npm pack` before doing the publish
- `npm publish` does a lot of things not covered by [npm-registry-client](https://github.com/npm/npm-registry-client) (see [publish.js](https://github.com/npm/npm/blob/latest/lib/publish.js) and its dependencies). Among others:
  - Add metadata to `package.json`
  - Read the `.npmrc` config and transform some of its parameters
  - Handle a cache
  - Handle `prepublish` and `prepublishOnly` scripts

A programmatic solution would result in doubling or tripling the `semantic-release` code base, add complexity and maintenance, and force us to replicate every change done in `npm`.
Plus it seems more important for `semantic-release` to guarantee that packages are published exactly the same way they would be without it.

## Next steps
In order to complete some of the features in this PR some plugin modification should follow:
- [ ] Use the logger passed to the plugin to add more logs in plugins
- [ ] Use `commits`, `lastRelease` and `nextRelease` in the `generate-notes` plugin
- [ ] Possibly move `publish-npm.js` to it's own plugin and the `npm` related code to `last-release-npm`. See #484
- [ ] Possibly move `github-release.js` to it's own plugin along with the Github related code. See #484


pvdlg:
We should also merge #457 before this one, to avoid asking #457 author to do a big rebase/merge conflict. It would be easier to handle the merge here.

gr2m:
is the PR ready for review? You still have the WIP in the title so just want to make sure :)

gr2m:
> Add JSDoc

do you have a preferred library for API docs based on JSDoc comments? I'm looking for one to update https://octokit.github.io/node-github/, too :)

pvdlg:
The code ready for review.
I have the WIP in the title as I still have to add some JSdoc and update the README to reflect the changes.

pvdlg:
> do you have a preferred library for API docs based on JSDoc comments? I'm looking for one to update https://octokit.github.io/node-github/, too :)

I used [jsdoc-to-markdown](https://github.com/jsdoc2md/jsdoc-to-markdown) on another project that generate markdown file, then generate the site with Jekyll: https://pvdlg.github.io/uprogress/getting-started

[slate](https://github.com/lord/slate) is a popular solution but I never used it.

[readme.io](https://readme.io) might also be a solution and it has and open source license: https://readme.readme.io/v2.0/docs/open-source. Not sure if it can generate doc based on JSDoc, but it might be able to use the markdown files generated by [jsdoc-to-markdown](https://github.com/jsdoc2md/jsdoc-to-markdown).

I'm not sure it makes a lot senses for `semantic-release` to publish a website with the doc generated from JSDoc as most of the functions are private. The only exposed function is `index.js` (the one in this PR, not the old one.)

gr2m:
sorry I didn't get to review the PR today, but can't wait to get this landed :)

pvdlg:
> sorry I didn't get to review the PR today, but can't wait to get this landed :)

No problem. I have to finish the doc anyway :) Would you mind looking at #457 before?

pvdlg:
The PR should be fully ready for review now.

I didn't add JSDoc everywhere yet. That's quite a big task, so it will come in a future PR, probably after the biggest changes planned are completed.

README.md
```diff
@@ -83,11 +83,17 @@ When `semantic-release` is set up it will do that after every successful continu
...
+When pushing new commits with `git push` a CI build is triggered. After running the tests the command `semantic-release` will execute the following tasks in series:
+
+| Step               | Description                                                                                 |
+| ------------------ | ------------------------------------------------------------------------------------------- |
+| Verify Conditions` | Run the [verifyConditions](#verifyConditions) plugin)                                       |
+| Get last release`  | Obtain last release with the [getLastRelease](#getLastRelease) plugin                       |
+| Analyze commits    | Determine the type of release to do with the [analyzeCommits](#analyzeCommits) plugin       |
+| Verify release     | Call the [verifyRelease](#verifyRelease) plugin                                             |
+| npm publish        | Update teh version in `package.json` and call `npm publish`                                 |
```
gr2m: typo: teh

src/index.js
```diff
@@ -1,153 +1,60 @@
...
-
-      const args = ['pre', (err.code ? err.code + ' ' : '') + err.message];
-      if (err.stack) args.push(err.stack);
-      log.error.apply(log, args);
-      process.exit(1);
-    }
-
-    const message = 'Determined version ' + release.version + ' as "' + npm.tag + '".';
+  if (!options.dryRun) {
+    logger.log('Call plugin %s ', 'verify-conditions');
```
gr2m: nitpick: space after `%s`

README.md
pvdlg: Done!

src/index.js
pvdlg: It wasn't intentional. It's fixed.

felixfbecker:
> Semantic-Release must now be executed with semantic-release instead of semantic-release pre && npm publish && semantic-release post.

What is the benefit from this change? Imo this makes it unnecessarily hard to publish to things other than npm. It is nice to be able to publish e.g. a VS Code extension without having to write any custom plugins for publishing that, or using semantic-release to publish a docker image on Docker hub. In a non-JS project you don't want to maintain related JS code, a CLI command is very easy to understand for everyone. Am I missing some better alternative here?

VS Code extension:

https://github.com/felixfbecker/vscode-php-intellisense/blob/d293ba3323f705404dcae84959f95dc1dce128f0/package.json#L46

Docker image:

https://github.com/felixfbecker/php-language-server/blob/1db6b7bbb3af458feffe356427fae6f85f8a025c/package.json#L7

bin/semantic-release.js
```diff
@@ -23,8 +23,6 @@ npx is bundled with npm >= 5.4, or available via npm. More info: npm.im/npx`
 }
 
 // node 8+ from this point on
-require('../src')().catch(err => {
-  console.error('An error occurred while running semantic-release');
-  console.error(err);
-  process.exit(1);
+require('../src/cli')().catch(() => {
```
felixfbecker: Why is this error not logged anymore?

src/cli.js
```diff
@@ -0,0 +1,59 @@
...
+    if (program.args.length > 0) {
+      program.outputHelp();
+      process.exitCode = 1;
+    } else {
+      await require('./index')(program);
+    }
+  } catch (err) {
+    // If error is a SemanticReleaseError then it's an expected exception case (no release to be done, running on a PR etc..) and the cli will return with 0
+    // Otherwise it's an unexpected error (configuration issue, code issue, plugin issue etc...) and the cli will return 1
+    if (err instanceof SemanticReleaseError) {
```
felixfbecker: Could a plugin have a different version of `@semantic-release/error` that makes this check fail? Would checking `err.name` work too?

pvdlg:
Having 3 different command was forcing to return a 1 exit code in order to prevent the next command to run. For example if `semantic-release pre` determines no release is necessary, the only to prevent `npm publish` to be executed was to return and error.

In addition of being counter intuitive and giving the impression that an error happened, it forced us to run `semantic-release` only on a CI step that wouldn't fail the build in case of an error.
This was creating a lot of confusion and difficult issues to troubleshoot.

In addition to that it's was not really possible to use `semantic-release` without npm anyway for several reason:
- We rely on the npm registry to determine the last version released
- We were verifying at the very beginning of `semantic-release pre` taht a NPM token is set

Here is a proposal to support any type of release: #484

Once it's implemented you would be able to configure `semantic-release` as follow to skip npm publish:
```json
"release": {
 "publish: ["github-release"]"
}
```
As mentioned in #484 we would also provide a plugin or another mechanist that would allow to set a command line to run, allow to do any type of release that can be done via command line.

src/lib/logger.js
```diff
@@ -0,0 +1,18 @@
...
+module.exports = {
+  log(...args) {
+    const [format, ...rest] = args;
+    console.log(`${chalk.grey('[Semantic release]:')} ${format}`, ...rest.map(arg => chalk.magenta(arg)));
+  },
+  error(...args) {
+    const [format, ...rest] = args;
+    console.error(
+      `${chalk.grey('[Semantic release]:')} ${chalk.red(format instanceof Error ? format.stack : format)}`,
+      ...rest.map(arg => chalk.red(arg instanceof Error ? arg.stack : arg))
```
felixfbecker: why only log `arg.stack`? Afaik `console.error()` should log the stack of errors, but if the error has more metadata it will log that too. For example, child_process.exec() errors have an `stderr` property that gives you the actual error output of the command that failed, without this the error is often times useless

bin/semantic-release.js
pvdlg: It is logged here: https://github.com/semantic-release/semantic-release/blob/caribou/src/cli.js#L56

In addition it's tested there: https://github.com/semantic-release/semantic-release/blob/caribou/test/integration.test.js#L618

src/cli.js
pvdlg: Could you clarify what do you mean by "a different version of @semantic-release/error" ?

The idea here is that throwing a `@semantic-release/error` will stop the process and exit gracefully without an error.
While throwing any other error is considered and actual problem that prevent `semantic-release` to work and will result in logging the error with a stack trace and exiting with 1.

felixfbecker: If a plugin has a dependency on `@semantic-releae/error` `^1.0.0` and this module has a dependency on `@semantic-release/error` `^2.0.0` the `instanceof` expression will always be false because it will be different classes, so this logic depends on the npm deduplication logic.

felixfbecker:
@pvdlg ah, that makes sense.

> In addition to that it's was not really possible to use semantic-release without npm anyway

The examples I posted work ;)

> We rely on the npm registry to determine the last version released

That is just the default, it can be overridden. https://www.npmjs.com/package/last-release-git makes it possible to determine the last release from git tags.

> We were verifying at the very beginning of semantic-release pre taht a NPM token is set

Yeah, that was annoying. It can be worked around by setting `NPM_TOKEN` to a dummy value.
See https://github.com/semantic-release/semantic-release/issues/465#issuecomment-339809654

> As mentioned in #484 we would also provide a plugin or another mechanist that would allow to set a command line to run, allow to do any type of release that can be done via command line.

That would be great, if the plugin just had an option to specify an arbitrary command to run in package.json 👍

src/lib/logger.js
pvdlg: I checked that `arg.stack` logs the error type and the stacktrace. Can you confirm it won't log any other parameters?

This log is useful for debugging purpose. Is there a scenario in which the stacktrace is not enough to find where the problem happen and reproduce?
For example is the case of `child_process.exec()` with the stacktrace you could find the line in the source that caused the error and see the command executed there.

On a side note a plugin can also catch its own errors and log them before re-throwing them to `semantic-release`.

pvdlg:
> That would be great, if the plugin just had an option to specify an arbitrary command to run in package.json 👍

Absolutely. Implementing a solution that let users bypass npm publish or run another command instead is the next priority.

src/lib/logger.js
felixfbecker: This is what a logged child process `err.stack` looks like:

```
Error: Command failed: git status
    at checkExecSyncError (child_process.js:591:13)
    at Object.execSync (child_process.js:631:13)
    at repl:1:32
    at ContextifyScript.Script.runInThisContext (vm.js:44:33)
    at REPLServer.defaultEval (repl.js:239:29)
    at bound (domain.js:301:14)
    at REPLServer.runBound [as eval] (domain.js:314:12)
    at REPLServer.onLine (repl.js:440:10)
    at emitOne (events.js:120:20)
    at REPLServer.emit (events.js:210:7)
```

This is what `console.error(err)` looks like:
```
{ Error: Command failed: git status
fatal: Not a git repository (or any of the parent directories): .git

    at checkExecSyncError (child_process.js:591:13)
    at Object.execSync (child_process.js:631:13)
    at repl:1:32
    at ContextifyScript.Script.runInThisContext (vm.js:44:33)
    at REPLServer.defaultEval (repl.js:239:29)
    at bound (domain.js:301:14)
    at REPLServer.runBound [as eval] (domain.js:314:12)
    at REPLServer.onLine (repl.js:440:10)
    at emitOne (events.js:120:20)
    at REPLServer.emit (events.js:210:7)
  error: null,
  cmd: 'git status',
  file: '/bin/sh',
  args: [Array],
  options: [Object],
  envPairs: [Array],
  stderr: 'fatal: Not a git repository (or any of the parent directories): .git\n',
  stdout: '',
  pid: 1492,
  output: [Array],
  signal: null,
  status: 128 }
```
The former log doesn't help, the latter does because the important metadata (`stderr`) is included.

This was an actual real life situation: https://github.com/semantic-release/semantic-release/issues/477
Took me forever to find out what went wrong

This is just an example, it's quite common to attach more metadata to properties, e.g. when wrapping (an)other error(s)
```js
throw Object.assign(new Error('Multiple other operations failed'), { errors })
```

src/cli.js
pvdlg: Yes indeed. I didn't think about this case.
I realize there is a problem in the definition of @semantic-release/error as the name is not set properly.
```js
const SemanticReleaseError = require('@semantic-release/error');

const err = new SemanticReleaseError('message', 'CODE');
console.log(err.name);
// Error
```
Ideally we need to fix that and use `err.name` in this test.

src/lib/logger.js
pvdlg: Ok make sense. I'll fix it shortly. Thanks for the feedback !

marcbachmann:
Nice update 👏 
One change you didn't mention anywhere is how to handle the shrinkwrap.
Previously we had to use
`semantic-release pre && npm shrinkwrap && semantic-release post`

According to the changes in this pr that changed to
`npm shrinkwrap && semantic-release`.

gr2m:
Thanks @marcbachmann, I've added this to the release notes

> * If you depend on `npm shrinkwrap`, you have to run it before `semantic-release` now. You can set `"presemantic-release": "npm shrinkwrap"` in your package.json file

Does that sound good to you?

pvdlg:
If I'm not mistaken, `npm shrinkwrap` just create the shrinkwrap file but doesn't publish anything.
So what would be the purpose of `semantic-release pre && npm shrinkwrap && semantic-release post` ?

I think `npm shrinkwrap` is meant to be used in your dev environment , and the shrinkwrap should be committed to your repo. During the release on the CI you probably don't want to re-generate your shrinkwrap file and use the one from your repo.

gr2m:
> Lock down dependency versions for publication

https://docs.npmjs.com/cli/shrinkwrap

`npm shrinkwrap` writes a `npm-shrinkwrap.json` file which is published to the npm registry, unlike `package-lock.json`

pvdlg:
> npm shrinkwrap writes a npm-shrinkwrap.json file which is published to the npm registry, unlike package-lock.json

`npm-shrinkwrap.json` should be published on the repo and the release process should use that file and not regenerate one, as we want to publish the `npm-shrinkwrap.json` from the repo not a generated one.

gr2m:
ah I get what you mean now. Yeah makes sense to me. But either way, if someone had a use case to use `npm shrinkwrap` before, they still can do it now

marcbachmann:
> I think npm shrinkwrap is meant to be used in your dev environment , and the shrinkwrap should be committed to your repo. During the release on the CI you probably don't want to re-generate your shrinkwrap file and use the one from your repo.

If you write it in development, the shrinkwrap doesn’t contain the actual version. There were also issues in earlier versions where we had to ececute `npm prune —production` to prepare the modules for the shrinkwrap command. 

Generating the file in development is more error-prone. It introduces more steps for developers and also doesn’t work in combination with bots like greenkeeper. And then there are the merge conflicts on github. Your suggested workflow might work for libraries but not on projects that have too many prs.

felixfbecker:
@marcbachmann if you want to lock versions in a project and not a library, you should use package-lock.json. npm 5 updates the file automatically. I would generally recommend to only lock versions on the top-level consuming project, not the library level.The point of lockfiles is to have the same versions in development and production. Greenkeeper has support for updating lockfiles.

hbetts:
> I would generally recommend to only lock versions on the top-level consuming project, not the library level.The point of lockfiles is to have the same versions in development and production.

For the same reason, I strongly recommend library developers commit their `package-lock.json` file to their repository.

This ensures that if I pull down your library to fix a bug, or add an enhancement, that I will get the same working dependencies used by the last working build.

Without a `package-lock.json` file it's possible that I won't be able to contribute to your library because of a broken upstream dependency.

hbetts:
> npm-shrinkwrap.json should be published on the repo and the release process should use that file and not regenerate one, as we want to publish the npm-shrinkwrap.json from the repo not a generated one.

I really agree with @pvdlg here. You want to be publishing the lockfile that the developer pushed to the repository that was then successfully built. You should never `npm publish` a lockfile that hasn't passed your CI tests and isn't part of your repository (so that future builds works the exact same).

felixfbecker:
> For the same reason, I strongly recommend library developers commit their package-lock.json file to their repository.
> 
> This ensures that if I pull down your library to fix a bug, or add an enhancement, that I will get the same working dependencies used by the last working build.
> 
> Without a package-lock.json file it's possible that I won't be able to contribute to your library because of a broken upstream dependency.

Imo that gives a false sense of security, because consumers of the library would still get the broken package. So I would rather have the build be broken than to pretend everything's fine and then pin the dependency if needed - for contributors _and_ consumers. After all, the package.json specifies the versions that are guaranteed to work. This is the primary reason I run nightly builds for OS libraries, so that a green build status actually indicates that the package is 100% functional for consumers and a red status indicates it's broken for consumers. But I know this is a very controversial topic.

Issue #423, opened 2017-08-25
Can I use semantic-release with yarn?


sudo-suhas:
Can I replace the `npm publish` step with `yarn publish`? Are there any other changes I would need to do? I tried searching on google and in the issues here but was not able to find anything.

gr2m:
Hey @sudo-suhas I’m not using `yarn` myself, but as far as I know that should work. Please let us know if it does :) Would be great if you could send a pull request to adjust our README to mention how how to use semantic-release with yarn

sudo-suhas:
@gr2m I tried this out by creating a [POC repo](https://github.com/sudo-suhas/yarn-semantic-release-poc) but it does not work 😞. `yarn publish` does not publish the current version but instead tries to prompt for the new version. Which [errors out on travis](https://travis-ci.org/sudo-suhas/yarn-semantic-release-poc/builds/276691619#L477):

```console
$ semantic-release pre && yarn publish && semantic-release post
yarn publish v0.27.5
[1/4] Bumping version...
info Current version: 1.0.0
error An unexpected error occurred: "Can't answer a question unless a user TTY".
info If you think this is a bug, please open a bug report with the information provided in "/home/travis/build/sudo-suhas/yarn-semantic-release-poc/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/publish for documentation about this command.
```

I also tried `yarn publish --non-interactive` but that [didn't work either](https://travis-ci.org/sudo-suhas/yarn-semantic-release-poc#L470).

As suggested in this issue, https://github.com/yarnpkg/yarn/issues/3011, one alternative is to [pass the new version](https://yarnpkg.com/lang/en/docs/cli/publish/#toc-yarn-publish-new-version) `yarn publish --new-version <version>` but I can't quite figure out a simple and clean way to do this.

gr2m:
doesn’t yarn end up publishing to the same npm registry? Maybe you can use npm to publish while keep using yarn for dependency management?

sudo-suhas:
True. Wanted to use yarn just for the sake of consistency. I slightly dislike having to use both yarn and npm in the travis file.

gr2m:
Yeah I agree, I’m just not familiar with yarn myself. I’d be happy to accept a pull request if it makes things simpler. It could also be an addition to our README to inform people how to use semantic-release with yarn once someone figures it out :)

sudo-suhas:
[...]

PR #490, opened 2017-10-29
docs: Add FAQ section for using with yarn

sudo-suhas:
This adds a section to the FAQ for "Can I use `semantic-release` with yarn?" which summarises the discussion in #423.

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,7 +273,7 @@ Of course you can, but this doesn’t necessarily mean you should. Running your 
 
 ### Can I manually trigger the release of a specific version?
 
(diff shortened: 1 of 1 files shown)


sudo-suhas:
I wasn't aware of that. Perhaps @gr2m can clarify why he felt we could add a section to the FAQ.

gr2m:
yes @pvdlg is right, the new version of semantic-release does not expose the `npm publish` step any longer, basically we don't give users any choice any more. I'm sorry for that, but I fear the FAQ addition become obsolete with v9.0.0, I'm sorry I should have told you but forgot :(

sudo-suhas:
No problem @gr2m, wanted to be sure before closing the PR.

PR #491, opened 2017-10-29
Log error properties and check SemanticReleaseError with name

pvdlg:
`semantic-release` now log all information included in unexpected errors.
In order to verify if an error is a `SemanticReleaseError` we now use the `name` property rather than the instance type. The test will work properly if a plugin use a different version of `@semantic-release/error` than `semantic-release`.

We should probably deprecate all version of `@semantic-release/error` < 2.0.1 as it contains a bug that prevent the name and stacktrace to be included. See https://github.com/semantic-release/error/pull/71.

See https://github.com/semantic-release/semantic-release/pull/480#pullrequestreview-72660167

cc /@felixfbecker

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -52,7 +52,7 @@ module.exports = async () => {
       logger.log(`%s ${err.message}`, err.code);
     } else {
       process.exitCode = 1;
-      logger.error(err);
+      logger.error('An error occurred while running semantic-release: %O', err);
     }
   }
 };
--- a/src/cli.js
+++ b/src/cli.js
@@ -1,5 +1,4 @@
 const program = require('commander');
-const SemanticReleaseError = require('@semantic-release/error');
 const logger = require('./lib/logger');
 
 function list(values) {
@@ -49,7 +48,7 @@ module.exports = async () => {
   } catch (err) {
     // If error is a SemanticReleaseError then it's an expected exception case (no release to be done, running on a PR etc..) and the cli will return with 0
     // Otherwise it's an unexpected error (configuration issue, code issue, plugin issue etc...) and the cli will return 1
-    if (err instanceof SemanticReleaseError) {
+    if (err.semanticRelease) {
       logger.log(`%s ${err.message}`, err.code);
     } else {
       process.exitCode = 1;
--- a/src/lib/logger.js
+++ b/src/lib/logger.js
@@ -6,13 +6,18 @@ const chalk = require('chalk');
 module.exports = {
   log(...args) {
     const [format, ...rest] = args;
-    console.log(`${chalk.grey('[Semantic release]:')} ${format}`, ...rest.map(arg => chalk.magenta(arg)));
+    console.log(
+      `${chalk.grey('[Semantic release]:')}${typeof format === 'string'
+        ? ` ${format.replace(/%[^%]/g, seq => chalk.magenta(seq))}`
+        : ''}`,
+      ...(typeof format === 'string' ? [] : [format]).concat(rest)
+    );
   },
   error(...args) {
     const [format, ...rest] = args;
     console.error(
-      `${chalk.grey('[Semantic release]:')} ${chalk.red(format instanceof Error ? format.stack : format)}`,
-      ...rest.map(arg => chalk.red(arg instanceof Error ? arg.stack : arg))
+      `${chalk.grey('[Semantic release]:')}${typeof format === 'string' ? ` ${chalk.red(format)}` : ''}`,
+      ...(typeof format === 'string' ? [] : [format]).concat(rest)
     );
   },
 };
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
   "dependencies": {
     "@semantic-release/commit-analyzer": "^3.0.1",
     "@semantic-release/condition-travis": "^6.0.0",
-    "@semantic-release/error": "^2.0.0",
+    "@semantic-release/error": "^2.1.0",
     "@semantic-release/last-release-npm": "^2.0.0",
     "@semantic-release/release-notes-generator": "^5.0.0",
     "chalk": "^2.3.0",
--- a/test/fixtures/plugin-error-a.js
+++ b/test/fixtures/plugin-error-a.js
@@ -1,3 +1,5 @@
 module.exports = function(config, options, cb) {
-  cb(new Error('a'));
+  const error = new Error('a');
+  error.errorProperty = 'errorProperty';
+  cb(error);
 };
--- a/test/fixtures/plugin-error-inherited.js
+++ b/test/fixtures/plugin-error-inherited.js
@@ -1,0 +1,14 @@
+const SemanticReleaseError = require('@semantic-release/error');
+
+class InheritedError extends SemanticReleaseError {
+  constructor(message, code, newProperty) {
+    super(message);
+    Error.captureStackTrace(this, this.constructor);
+    this.name = this.constructor.name;
+    this.code = code;
+  }
+}
+
+module.exports = function(config, options, cb) {
+  cb(new InheritedError('Inherited error', 'EINHERITED'));
+};
--- a/test/integration.test.js
+++ b/test/integration.test.js
@@ -36,7 +36,6 @@ test.beforeEach(async t => {
 });
 
 test.afterEach.always(async t => {
-  console.log();
   // Restore process.env
   process.env = Object.assign({}, t.context.env);
   // Restore the current working directory
--- a/test/integration.test.js
+++ b/test/integration.test.js
@@ -17,6 +17,7 @@ const env = {
 const cli = require.resolve('../bin/semantic-release');
 const noop = require.resolve('../src/lib/plugin-noop');
 const pluginError = require.resolve('./fixtures/plugin-error-a');
+const pluginInheritedError = require.resolve('./fixtures/plugin-error-inherited');
 
 test.before(async t => {
   await mockServer.start();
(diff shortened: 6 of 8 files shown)


src/cli.js
```diff
@@ -49,11 +48,11 @@ module.exports = async () => {
   } catch (err) {
     // If error is a SemanticReleaseError then it's an expected exception case (no release to be done, running on a PR etc..) and the cli will return with 0
     // Otherwise it's an unexpected error (configuration issue, code issue, plugin issue etc...) and the cli will return 1
-    if (err instanceof SemanticReleaseError) {
+    if (err.name === 'SemanticReleaseError') {
       logger.log(`%s ${err.message}`, err.code);
     } else {
       process.exitCode = 1;
-      logger.error(err);
+      logger.error('An error occurred while running semantic-release: %s', err);
```
felixfbecker: what is `%s` expanding to for an `Error` instance? `error.toString()`? As mentioned in another thread, it should include error properties

src/lib/logger.js
```diff
@@ -10,9 +10,9 @@ module.exports = {
   },
   error(...args) {
     const [format, ...rest] = args;
-    console.error(
-      `${chalk.grey('[Semantic release]:')} ${chalk.red(format instanceof Error ? format.stack : format)}`,
-      ...rest.map(arg => chalk.red(arg instanceof Error ? arg.stack : arg))
-    );
+    console.error(`${chalk.grey('[Semantic release]:')} ${chalk.red(format)}`, ...rest.map(arg => chalk.yellow(arg)));
+    args.filter(arg => arg instanceof Error).forEach(error => {
+      console.error(error);
```
felixfbecker: since you use `%s`, would this not print the error twice? Instead of this handling, why not simply log like this:
```js
logger.error('An error occurred while running semantic-release: ', err)
```
(no `%s`)

src/cli.js
pvdlg: The error itself will be printed with https://github.com/semantic-release/semantic-release/blob/include-eror-props/src/lib/logger.js#L15

The output will be:
```
[Semantic release]: An error occurred while running semantic-release: Error: error message
{ Error: error message
    at module.exports (/Users/vanduynslager/git/semantic-release/test/fixtures/plugin-error-a.js:2:17)
    at bound  (internal/util.js:227:26)
    at /Users/vanduynslager/git/semantic-release/src/lib/plugins.js:22:58
    at <anonymous> errorProperty: 'errorProperty' }
```

The `%s` is used to print `[Semantic release]: An error occurred while running semantic-release: Error: error message`

felixfbecker: Yeah, realised that later down. Seems redundant though to log the message twice?

src/lib/logger.js
pvdlg: The function `logger.error` can be used to print anything, not only `Error` objects. So it has to work with string formatting and handled multiple parameters.

The function format the the parameters and print them in red, and in addition if there is errors among the parameters it will print them completely.

So the error message will appear in the formatted string and the complete error (with message, stacktrace and properties) will be logged after.

felixfbecker: `console.log`/`console.error` support format strings and arbitrary structures which result in being `util.inspect()`ed. It behaves differently depending on whether the first argument is a format string or not, which changes the semantic of the rest arguments.
This logger seems to only support the format string mode, and I assume `chalk.yellow()` implicitly converts the argument to a string ala `toString()`, so for arbitrary structures it would just print `[object Object]`.
I would make it behave like the native `console` as in not add custom handling of errors, but of format strings if that is needed.

src/cli.js
pvdlg: I made the change, the message shouldn't be logged twice anymore

src/lib/logger.js
```diff
@@ -6,13 +6,18 @@ const chalk = require('chalk');
 module.exports = {
   log(...args) {
     const [format, ...rest] = args;
-    console.log(`${chalk.grey('[Semantic release]:')} ${format}`, ...rest.map(arg => chalk.magenta(arg)));
+    console.log(
+      `${chalk.grey('[Semantic release]:')}${typeof format === 'string' ? format : ''}`,
```
felixfbecker: `console.log()` will check if the first arg contains any format sequences, not just if it's a string. I.e. these two act different:
```
console.log('an error occured: ', err)
console.log('an error occured: %s', err)
```

src/lib/logger.js
pvdlg: I made some changes. The Object will now be logged properly.
Both `log` and `error` support the following now:
- `logger.log(obj)`
- `logger.log('Something %O something else', obj)`
- `logger.log('Something %s something else','some string')`
- `logger.log('Something %s something else %O','some string', obj)`

src/lib/logger.js
pvdlg: That's expected I think.
`console.log('an error occured: ', err)` will log `err` as an object (with stacktrace etc)
`console.log('an error occured: %s', err)` will log only the equivalent of `err.toString()`

If a user use `%s` in their format it means they want to be logged a string.

Can you give me an example that won't work as expected ?

felixfbecker: also here's one idea: why not simply wrap all format sequences with chalk? I.e.
```ts
if (typeof format === 'string') {
  format.replace(/%[^%]/g, seq => chalk.magenta(seq))
}
```

felixfbecker: > console.log('an error occured: ', err) will log err as an object (with stacktrace etc)

Are you sure? As I understand the code this will wrap `err` in `chalk.magenta()` because the first argument is a string and it doesn't check whether it actually contains a format sequence like `console` does.

pvdlg:
```js
logger.log('this is an object', {a: 'a'});
// [Semantic release]: this is an object { a: 'a' }
```
```js
const error = new Error('err message');
error.property = 'propValue';
logger.error('this is an error', error);

//[Semantic release]: this is an error { Error: err message
//    at module.exports (/Users/vanduynslager/git/semantic-release/src/cli.js:57:21)
//    at <anonymous> property: 'propValue' }
```

pvdlg: > also here's one idea: why not simply wrap all format sequences with chalk? I.e.

That seems a good idea. Let me investigate a bit more (tomorrow though).

src/cli.js
```diff
@@ -49,11 +48,11 @@ module.exports = async () => {
   } catch (err) {
     // If error is a SemanticReleaseError then it's an expected exception case (no release to be done, running on a PR etc..) and the cli will return with 0
     // Otherwise it's an unexpected error (configuration issue, code issue, plugin issue etc...) and the cli will return 1
-    if (err instanceof SemanticReleaseError) {
+    if (err.name === 'SemanticReleaseError') {
```
gr2m: If plugins want to extend the SemanticReleaseError the name will change, see @felixfbecker's question at https://github.com/semantic-release/error/pull/71#discussion_r147571577?

What was your thinking behind changing the condition to check for the name?

pvdlg: Yes that's a problem indeed. Maybe the safe side solution is `if (err.name === 'SemanticReleaseError' || err instanceof SemanticReleaseError) {`.

@felixfbecker what do you think?

pvdlg: Thinking about it best solution might be to set a property in the SemanticReleaseError that indicate it's a semantic-release error. This way it would work with inheritance and wouldn't have the issue of multiple version of the error being different instances (and failing the `err instanceof SemanticReleaseError` test).

felixfbecker: What we need here is a common interface to be implemented by all errors that should not cause the build to not fail / are "expected". This could be
- common `name` that is always `SemanticReleaseError`, even for subclasses
- a property like `expected: true`
- a property `exitCode: 0` (this is non ideal because ChildProcess errors have an `exitCode` property)

pvdlg: Yes. See https://github.com/semantic-release/error/pull/72

Once that is merge I can do `if (err.semanticRelease) {`.

I choose to name the property `semanticRelease` instead of `expected` or `exitCode` to avoid conflicts.

gr2m: love it, good solution 👍

pvdlg: Ok. Once we merge semantic-release/error#72 it would be a good idea to deprecate older versions, to make sure third-party plugins use the last version that include the property.

pvdlg: Modification done

src/lib/logger.js
pvdlg: I implemented the `format.replace(/%[^%]/g, seq => chalk.magenta(seq))`. Thanks @felixfbecker !

pvdlg:
I think every feedback has been addressed now.
@felixfbecker @gr2m could you review ?

gr2m:
I'd squash the commits to just `fix: log error properties and check SemanticReleaseError by .semanticRelease property`, the current commit message is no longer correct, right?

pvdlg:
> I'd squash the commits to just fix: log error properties and check SemanticReleaseError by .semanticRelease property, the current commit message is no longer correct, right?

I renamed the commit properly. any reason to squash the 3 commits in one? They fix different things. So as much as possible I rather to smaller, specific commit. That create a release note with more info and make the repo history clearer.

felixfbecker:
Agree, as long as every commit works on it's own (a build on this would succeed) I wouldn't squash them

pvdlg:
> Agree, as long as every commit works on it's own (a build on this would succeed) I wouldn't squash them

Yes I make sure of that when I create multiple commits in a PR. If for some reasons I fix two related thing but can't make 2 commits that work by themselves, then I create 1 commit and I mention both fix in the body.

Make me think it would be nice to include commit body in release note (maybe in segment with ▶ that would allow to unfold/expand the body). But that's another story.

gr2m:
> I renamed the commit properly

Even better, I just wanted to safe you the extra work and make sure we don't merge it with the old commit message to avoid misleading change logs :)

PR #496, opened 2017-10-30
fix: Update npx command line

### LINKED ISSUE #445: Proposal regarding the node 8 limitation
Recently `semantic-release` moved to support only Node 8 or above (See [why-does-semantic-release-require-node-version--8](https://github.com/semantic-release/semantic-release#why-does-semantic-release-require-node-version--8)).

The workaround proposed [here](https://github.com/semantic-release/semantic-release#why-does-semantic-release-require-node-version--8) and [there](https://github.com/semantic-release/semantic-release/blob/12b35396d88d37de8000d7838b41de9f2b276f80/bin/semantic-release.js#L15) is to use `npx` but that doesn't solve the issue as `npx` is just a tool to run locally installed packages as global ones (basically allow to run `npm my-local-package` rather than `npm node_modules/.bin/my-local-package`). That doesn't allow to run Node 8 only package on an older version of node.
So this should be removed.
A solution that would allow to run Node 8 package on older version is [babel-node](https://babeljs.io/docs/usage/cli/#babel-node) but it's not recommended for production, and it doesn't provide polyfill for Node 8 API, it only transpile and polyfill Javascript features.
And it would't work either as the Node version is checked [here](https://github.com/semantic-release/semantic-release/blob/caribou/bin/semantic-release.js#L8).

In addition supporting only Node 8 creates some additional frictions for users:
- In a build with multiple node versions, `semantic-release` will be executed on each job, but will exit right away if it's not Node 8 and display an error. The error is expected and should be ignored but it might be confusing
- Require to create an extra Node 8 job in (rare) cases the CI doesn't already have one configured
- Forces to set in the `package.json` to `{"engines" : ""node": ">=4"}` in order to be installed by `yarn` (see #442)

All that can certainly be dealt with by users, but it adds on the cognitive load to use `semantic-release` as it requires for users to understand some of the internal mechanisms related to the interaction between `semantic-release` and the CI : CI commands run for every version of Node in the build, semantic-release exit right away with an error on every jobs but one, elect a build leader to figure out which one should not exit etc....

An alternative solution would be to use [babel-register](https://babeljs.io/docs/usage/babel-register/), it would allow to preserve the compatibility with Node 4, without creating much complexity for maintainer and contributors.
Here is how it would work.

Add on top of [bin/semantic-release.js](https://github.com/semantic-release/semantic-release/blob/caribou/bin/semantic-release.js):
```js
require('babel-register)({only: ['.src', '@semantic-release/*']}); // Transpile only sources and plugins
require('babel-polyfill);
```
And in `package.json`:
```json
"babel": {
  "presets": [
    ["env", {
      "targets": {
        "node": "current" // Transpiled only what's required by the currently running node
      }
    }]
  ]
}
```
And that's it !
The benefits:
- When executing `semantic-release` it would be transparently transpiled to support the currently running Node version (so that would skip transpilation when `semantic-release` is running on node)
- Allow to use all the last JS features, including `async/await`
- No additional build steps

According to the quick tests I did, the performance impact is negligible, ~200ms on Node4 and nonexistent on Node 8. Anyway most of `semantic-release` run time is spent on I/O (github API, npm registry, git commands etc..).
The only shortcoming is that the Node API released after 4 wouldn't be available. I checked the code and the only Node > 4 API we use is [promisify](https://nodejs.org/api/util.html#util_util_promisify_original) that can be replaced by [pify](https://github.com/sindresorhus/pify).

/cc @gr2m

pvdlg:
Per https://github.com/semantic-release/semantic-release/pull/414#issuecomment-331995349

Fix #445

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,7 +287,7 @@ It is indeed a great idea because it _forces_ you to follow best practices. If y
 
 Being able to write code for just the most recent node versions greatly simplifies development. More language features are available, no transpilation is required, less test builds are to be run, awaited and debugged.
 
-For a special purpose tool like `semantic-release`, that's only meant to be used in controlled CI environments, we think it's okay to have such a high version requirement. As `semantic-release` handles package publishing we expect almost every project to have at least one build job running node 8 already – and that's all it takes. Even if that's not that case `semantic-release` can still be executed with the help of [npx](https://www.npmjs.com/package/npx) (`npx -p node@8 npm run semantic-release`).
+For a special purpose tool like `semantic-release`, that's only meant to be used in controlled CI environments, we think it's okay to have such a high version requirement. As `semantic-release` handles package publishing we expect almost every project to have at least one build job running node 8 already – and that's all it takes. Even if that's not that case `semantic-release` can still be executed with the help of [npx](https://www.npmjs.com/package/npx) (`npx -p node@8 -c "semantic-release"`).
 
 ## Badge
 
--- a/bin/semantic-release.js
+++ b/bin/semantic-release.js
@@ -15,7 +15,7 @@ the build leader and you can safely ignore this message.
 If you don't have node 8 in your build matrix you can use "npx" to restore
 compatibility with minimal overhead:
 
-$ npx -p node@8 npm run semantic-release
+$ npx -p node@8 -c "semantic-release"
 
 npx is bundled with npm >= 5.4, or available via npm. More info: npm.im/npx`
   );


gr2m:
I've tested the new command but got an error:
https://travis-ci.org/gr2m/semantic-release-cli-test#L564

It looks like it cannot access `NPM_TOKEN` but it is set at https://travis-ci.org/gr2m/semantic-release-cli-test#L422

I'll look more into it

gr2m:
I don't think it's the access to the environment variable, but the location of the `.npmrc` file

gr2m:
@pvdlg by the way the `%O` placeholder doesn't get replaced correctly: https://travis-ci.org/gr2m/semantic-release-cli-test#L564, I don't know why, works for me when testing locally with Node 8

```
> console.error('test 12 %O', new Error('funky'))
test 12 Error: funky
    at repl:1:29
    at ContextifyScript.Script.runInThisContext (vm.js:50:33)
    at REPLServer.defaultEval (repl.js:240:29)
    at bound (domain.js:301:14)
    at REPLServer.runBound [as eval] (domain.js:314:12)
    at REPLServer.onLine (repl.js:441:10)
    at emitOne (events.js:120:20)
    at REPLServer.emit (events.js:210:7)
    at REPLServer.Interface._onLine (readline.js:282:10)
    at REPLServer.Interface._line (readline.js:631:8)
```

But it doesn't when running in Node 6 or 4. Maybe the `execa` is messing with the npx command somehow?

pvdlg:
Yes I noticed that as well. I'm looking at it.

pvdlg:
On a local project if I run `npx -p node@8 -c "semantic-release"`
I get the error: `/bin/sh: semantic-release: command not found`

If I have `semantic-release` installed globally it works.

I'm really confused by what `npx -p node@8 -c "semantic-release"` actually does...
The problem with the NPM_TOKEN missing might be due to `semantic-release` not being executed from the root directory of the project...

pvdlg:
Running on Node 8:
```
node -e "console.log('TEST %O', {a: 1})"
=> TEST { a: 1 }
```
```
npx -p node@8 -c "node -e \"console.log('TEST %O', {a: 1})\""
=> TEST %O { a: 1 }
```
¯\_(ツ)_/¯

pvdlg:
Ok the `%O` problem is due to https://github.com/nodejs/node/pull/14558 that has been implemented in Node 8.4.0.

When running `npx -p node@8` the version used is `8.3.0` as it's the latests published on npm.

So we can either decide to support only node `>=8.4.0` or I'll have to find another way to format objects in logs. I'm sure their library that can do that.

gr2m:
I'm okay with supporting only Node >=8.9 as it became LTS today 🎉 

>    npx -p node@8 -c "node -e \"console.log('TEST %O', {a: 1})\""
      => TEST %O { a: 1 }

I was not able to reproduce this ¯\\_(ツ)\_\/¯

My guess it had an older Node 8 version installed and because it matched the wanted node version it used the local version instead of getting the latest? Try

```
npx -p node@8.9.0 -c "node -e \"console.log('TEST %O', {a: 1})\""
```

gr2m:
by the way `node` is just an npm package: https://www.npmjs.com/package/node

gr2m:
> I get the error: /bin/sh: semantic-release: command not found

can you try the `--shell-auto-fallback` flag? 

Docs say

> --shell-auto-fallback [<shell>] - Generates shell code to override your shell's "command not found" handler with one that calls npx. Tries to figure out your shell, or you can pass its name (either bash, fish, or zsh) as an option. See below for how to install.

pvdlg:
> can you try the --shell-auto-fallback flag?

Now running `npx -p node@8 -c "semantic-release"` works... I don't have `/bin/sh: semantic-release: command not found`

Maybe a weird bug somewhere between a specific version of `node` / `npx` / `npm`?

README.md
```diff
@@ -287,7 +287,7 @@ It is indeed a great idea because it _forces_ you to follow best practices. If y
...
 
-For a special purpose tool like `semantic-release`, that's only meant to be used in controlled CI environments, we think it's okay to have such a high version requirement. As `semantic-release` handles package publishing we expect almost every project to have at least one build job running node 8 already – and that's all it takes. Even if that's not that case `semantic-release` can still be executed with the help of [npx](https://www.npmjs.com/package/npx) (`npx -p node@8 npm run semantic-release`).
+For a special purpose tool like `semantic-release`, that's only meant to be used in controlled CI environments, we think it's okay to have such a high version requirement. As `semantic-release` handles package publishing we expect almost every project to have at least one build job running node 8 already – and that's all it takes. Even if that's not that case `semantic-release` can still be executed with the help of [npx](https://www.npmjs.com/package/npx) (`npx -p node@8 -c "semantic-release"`).
```
felixfbecker: I think it should also be mentioned that if you're using Travis build stages, you have the freedom to specify a separate node version for that stage anyway

pvdlg:
Fixed by #593

PR #519, opened 2017-11-22
Lint with XO

pvdlg:
Making a lot of modification in `semantic-release` I grew increasingly frustrated with `eslint-config-standard` as `standard` disable many checks for common mistakes.

Here is a few examples (among many other instances) encountered across the `semantic-release` repos:
- Enforce using radix in parseInt() => [example](https://github.com/semantic-release/npm/pull/1#discussion_r149567842)
- Avoid negated conditions => [example](https://github.com/semantic-release/npm/pull/1#discussion_r149568538) and [example](https://github.com/semantic-release/npm/pull/1#discussion_r149568678)
- Use `const` or `let` instead of `var`
- Use `const` instead of `let` when possible => [example](https://github.com/semantic-release/semantic-release/compare/xo?expand=1#diff-3e42fed5223348e612fea401149a0b8eR77)
- Ensure there is no `.only` or `.skip` in test files (convenient in dev but must be removed before pushing the code) => I add to rebase several times on I realized the coverage was anomaly low
- Ensure there is no duplicate tests title
- Don't set function as `async` if they are not => [example](https://github.com/semantic-release/semantic-release/compare/xo?expand=1#diff-e6f6dfcb1ea3a2cce9db1c42bb9eee3eL87)
- Enforce correct Error subclassing => [example](https://github.com/semantic-release/error/pull/71)
- Consistent case (camelcase) for variables and properties: [example](https://github.com/semantic-release/semantic-release/pull/457#discussion_r147536084)
- Make sure that async tests are awaited => I spent lot of time debugging this kind of issue, trying to figure out why my test was failing

As a result I have to spend a considerable amount of time (and cognitive load) checking these kind of things. And even by spending this time and having meticulous PR reviews, a lot of mistakes are passing through the cracks.

So, I propose to switch from `eslint` + `standard` + `prettier` to [xo](https://github.com/sindresorhus/xo) + `prettier` as it provides several advantages:
- All the "missing" rules mentioned above are covered
- All the rules currently checked are covered
- [Less dependencies](https://github.com/semantic-release/semantic-release/compare/xo?expand=1#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L45) in `package.json`
- [Convenient rules for `ava`](https://github.com/avajs/eslint-plugin-ava)
- Automatically exclude file in `.gitignore`. That avoid to have to whitelist the files to lint, and therefore avoid to forget some (That happened to me several times...)
- Violation output is much [nicer and readable](https://raw.githubusercontent.com/sindresorhus/eslint-formatter-pretty/master/screenshot.png)
- Great [editor plugins](https://github.com/sindresorhus/xo#editor-plugins) even if installing one is not required

If we go that road, I'll be happy to update the other repo :)

### DIFF
--- a/bin/semantic-release.js
+++ b/bin/semantic-release.js
@@ -3,6 +3,7 @@
 // Bad news: We have to write plain ES5 in this file
 // Good news: It's the only file of the entire project
 
+// eslint-disable-next-line no-var
 var semver = require('semver');
 
 if (semver.lt(process.version, '8.0.0')) {
@@ -22,7 +23,7 @@ npx is bundled with npm >= 5.4, or available via npm. More info: npm.im/npx`
   process.exit(1);
 }
 
-// node 8+ from this point on
+// Node 8+ from this point on
 require('../cli')().catch(() => {
   process.exitCode = 1;
 });
--- a/lib/get-next-version.js
+++ b/lib/get-next-version.js
@@ -3,15 +3,15 @@ const SemanticReleaseError = require('@semantic-release/error');
 
 module.exports = (type, lastRelease, logger) => {
   let version;
-  if (!lastRelease.version) {
-    version = '1.0.0';
-    logger.log('There is no previous release, the next release version is %s', version);
-  } else {
+  if (lastRelease.version) {
     version = semver.inc(lastRelease.version, type);
     if (!version) {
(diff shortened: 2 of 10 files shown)


pvdlg:
> sorry for the trouble you run to. Okay with me

No problems :)

PR #524, opened 2017-11-24
Make semantic-release language agnostic

pvdlg:
The last language specific thing `semantic-release` relies on is the `package.json`.
This PR remove this reliance by allowing to configure `semantic-release` with rc files in addition of the current `release` property in the `package.json`.

This way, with the proper plugins, `semantic-release` can release any type of project.
We should now recommend to install `semantic-release` globally:
```
npm install semantic-release -g
semantic-release
```
or with `npx`
```
npx semantic-release
```
or in Unix based CI:
```
if [[ `node -v | egrep -o '[0-9|\.]+'` > 8* ]]; then npm install semantic-release -g && semantic-release; fi
```

Here is the list of changes:
- Do not rely on `package.json` anymore
- Use `cosmiconfig` to load the configation. `semantic-release` can be configured:
  - via CLI options (including plugin names but not plugin options)
  - in the `release` property of `package.json` (as before)
  - in a `.releaserc.yml` or `.releaserc.js` or `.releaserc.js` or `release.config.js` file
  - in a `.releaserc` file containing `json`, `yaml` or `javascript` module
- Add the `repositoryUrl` options (used across `semantic-release` and plugins). The value is determined from CLi option, or option configuration, or package.json or the git remote url
- Verifies that `semantic-release` runs from a git repository
- `pkg` and `env` are not passed to plugin anymore
- `semantic-release` can be run both locally and globally. If ran globally with non default plugins, the plugins can be installed both globally or locally.

In addition with PR `semantic-release` now expect plugin to provide an async or `Promise` returning function rather than a function with a callback.

BREAKING CHANGE: `pkg` and `env` are not passed to plugin anymore.
Plugins relying on a `package.json` must verify the presence of a valid `package.json` and load it.
Plugins can use `process.env` instead of `env`.

BREAKING CHANGE: Each plugin is expected to return an async function or a Promise returning function. The callback parameter is not passed to plugins anymore.

This PR should be merged only after all the default plugin are updated. It currently uses the plugin directly from Github. So we'll have to change the `package.json` to point to the released version of the plugins.

Fix #526, Fix #527, Fix #528, Fix #529

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,8 +210,6 @@ module.exports = function (pluginConfig, config, callback) {}
   - `options`: `semantic-release` options like `debug`, or `branch`
   - `pkg`: Parsed `package.json`
   - For certain plugins the `config` object contains even more information. See below.
-- `callback`: If an error occurs pass it as first argument. Otherwise pass your result as second argument.
-
 
 ### `analyzeCommits`
 
--- a/README.md
+++ b/README.md
@@ -167,6 +167,7 @@ semantic-release
 
 These options are currently available:
 - `branch`: The branch on which releases should happen. Default: `'master'`
+- `repositoryUrl`: The git repository URL. Default: `repository` property in `package.json` or git origin url. Any valid git url format is supported (See [Git protocols](https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols)). If the [Github plugin](https://github.com/semantic-release/github) is used the URL must be a valid Github URL that include the `owner`, the `repository` name and the `host`. The Github shorthand URL is not supported.
 - `dry-run`: Dry-run mode, skipping verifyConditions, publishing and release, printing next version and release notes
 - `debug`: Output debugging information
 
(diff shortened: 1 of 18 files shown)


README.md
```diff
@@ -167,6 +167,7 @@ semantic-release
 
 These options are currently available:
 - `branch`: The branch on which releases should happen. Default: `'master'`
+- `repositoryUrl`: The fit repository URL. Default: git origin url or `repository` property in `package.json`
```
gr2m:
```diff
- The fit repository URL
+ the git repository URL
```

gr2m: What format can the `git` repository URL have? Maybe we could add a few examples or link to the package we use to parse the URL?

pvdlg: I fixed the typo and added a note regarding the supported format.
We do not support the shorthand format (for example `npm/npm`) because [release-notes-generator](https://github.com/semantic-release/release-notes-generator) needs to determine the hostname to create links in the release notes. With the shorthand format we don't have a way to figure out if the hostname is `github.com`, `gitlab.com` or any custom GHE domain.
I will check if we can generate release notes with relative URLs. If it's the case I will create a PR on `release-notes-generator` to handle that and remove the restriction in the `semantic-release` doc.

PR #531, opened 2017-11-25
feat: extract `getCommits` into a plugin

pmowrer:
This PR, at least initially, is a bit of a POC/suggestion. I think `semantic-release` could work really well for an `npm` package monorepo if...

- [ ] ... the commits could be filtered based on what files they touch (e.g. only include commits that touch files within the same folder as the `package.json` for a given package).

- [x] ... the git tag could be customized (e.g. instead of `v1.0.0`, `my-package-v1.0.0`). The new plugin system allows for this (albeit in kind of a roundabout way)!

Please see the following plugin repo with crude implementation of aforementioned requirements:
https://github.com/Updater/semantic-release-monorepo

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,7 +3,6 @@ const TerminalRenderer = require('marked-terminal');
 const SemanticReleaseError = require('@semantic-release/error');
 const getConfig = require('./lib/get-config');
 const getNextVersion = require('./lib/get-next-version');
-const getCommits = require('./lib/get-commits');
 const logger = require('./lib/logger');
 const {gitHead: getGitHead, isGitRepo} = require('./lib/git');
 
@@ -27,19 +26,21 @@ module.exports = async opts => {
   }
 
   logger.log('Call plugin %s', 'get-last-release');
-  const {commits, lastRelease} = await getCommits(
-    await plugins.getLastRelease({options, logger}),
-    options.branch,
-    logger
-  );
+  const lastRelease = await plugins.getLastRelease({options, logger});
+
+  logger.log('Call plugin %s', 'get-commits');
+  const {commits, lastRelease: {version, gitHead, gitTag}} = await plugins.getCommits({lastRelease, options, logger});
+  lastRelease.version = version;
+  lastRelease.gitHead = gitHead;
+  lastRelease.gitTag = gitTag;
 
   logger.log('Call plugin %s', 'analyze-commits');
   const type = await plugins.analyzeCommits({options, logger, lastRelease, commits});
   if (!type) {
     throw new SemanticReleaseError('There are no relevant changes, so no new version is released.', 'ENOCHANGE');
   }
-  const version = getNextVersion(type, lastRelease, logger);
-  const nextRelease = {type, version, gitHead: await getGitHead(), gitTag: `v${version}`};
+  const newVersion = getNextVersion(type, lastRelease, logger);
+  const nextRelease = {type, version: newVersion, gitHead: await getGitHead(), gitTag: `v${newVersion}`};
 
   logger.log('Call plugin %s', 'verify-release');
   await plugins.verifyRelease({options, logger, lastRelease, commits, nextRelease});
--- a/lib/get-config.js
+++ b/lib/get-config.js
@@ -14,6 +14,7 @@ module.exports = async (opts, logger) => {
   debug('repositoryUrl: %O', options.repositoryUrl);
   debug('analyzeCommits: %O', options.analyzeCommits);
   debug('generateNotes: %O', options.generateNotes);
+  debug('getCommits: %O', options.getCommits);
   debug('verifyConditions: %O', options.verifyConditions);
   debug('verifyRelease: %O', options.verifyRelease);
   debug('publish: %O', options.publish);
--- a/lib/plugins/definitions.js
+++ b/lib/plugins/definitions.js
@@ -29,6 +29,14 @@ module.exports = {
         'The "getLastRelease" plugin output if defined, must be an object with an optionnal valid semver version in the "version" property.',
     },
   },
+  getCommits: {
+    default: './get-commits.js',
+    config: {
+      validator: conf => Boolean(conf) && validatePluginConfig(conf),
+      message:
+        'The "getCommits" plugin is mandatory, and must be a single plugin definition. A plugin definition is either a string or an object with a path property.',
+    },
+  },
   analyzeCommits: {
     default: '@semantic-release/commit-analyzer',
     config: {
--- a/lib/plugins/get-commits.js
+++ b/lib/plugins/get-commits.js
@@ -1,0 +1,7 @@
+const getCommits = require('../get-commits');
+
+module.exports = {
+  getCommits: async (pluginConfig, {lastRelease, options: {branch}, logger}) => {
+    return getCommits(lastRelease, branch, logger);
+  },
+};
--- a/test/plugins/definitions.test.js
+++ b/test/plugins/definitions.test.js
@@ -23,6 +23,17 @@ test('The "getLastRelease" plugin is mandatory, and must be a single plugin defi
   t.true(definitions.getLastRelease.config.validator(() => {}));
 });
 
+test('The "getCommits" plugin is mandatory, and must be a single plugin definition', t => {
+  t.false(definitions.getCommits.config.validator({}));
+  t.false(definitions.getCommits.config.validator({path: null}));
+  t.false(definitions.getCommits.config.validator([]));
+  t.false(definitions.getCommits.config.validator());
+
+  t.true(definitions.getCommits.config.validator({path: 'plugin-path.js'}));
+  t.true(definitions.getCommits.config.validator('plugin-path.js'));
+  t.true(definitions.getCommits.config.validator(() => {}));
+});
+
 test('The "analyzeCommits" plugin is mandatory, and must be a single plugin definition', t => {
   t.false(definitions.analyzeCommits.config.validator({}));
   t.false(definitions.analyzeCommits.config.validator({path: null}));


pvdlg:
Thanks for your interest in `semantic-release` and your contribution !

As a general rule when you want to propose an important change it's better to open an issue first to discuss the change. So if the change is refused you'll save time :)

After discussing with the maintainer team, we would be open to support monorepo, but only on certain conditions:
- It would be supported through third-party plugins
- Only minor changes would have to be made to `semantic-release` core
- The monorepo implementation would not go against the core principle of `semantic-release` (i.e. [release-on-every-push](https://github.com/semantic-release/semantic-release#is-it-really-a-good-idea-to-release-on-every-push))

Here is some of the reasons for this:
- There is currently multiple monorepo solutions: [lerna](https://github.com/lerna/lerna), [builder](https://github.com/FormidableLabs/builder) I think, [yarn workspaces](https://yarnpkg.com/lang/en/docs/workspaces/), [zelda](https://github.com/feross/zelda) and many custom ones
- All this solution are quite different and it's not really clear which one is going to become the most common one
- It doesn't seems there is a strong agreement yet on how to manage naming convention (commits, tags, versions etc...) so it makes it difficult to determine which commit belong to which package etc...
- Some of the features provided by monorepo are covered by `semantic-release` (arguably in a better way)
  - monorepo make the release process easier as only one command has to be ran to release multiple packages => `semantic-release` goes further by automating the release process completely
  - Monorepo make changes across package easier => `semantic-release` enforces `semver` and allow to make a release on each push. This is arguably a better way to handle modularization as it guarantees the dependencies are unidirectional and that changes will be displayed outside the organization properly (by releasing a major version on breaking changes)

My humble opinion, from what I read regarding monorepo, is that the most promising solution is [zelda](https://github.com/feross/zelda) as it's non intrusive, provide most of the benefits of other solution, without the additional complexity/constraints of dealing with commit/tag naming convention, mixing the issue tracker, releases etc...

Regarding your proposed change I don't think it's the right approach. `semantic-release` must continue to identify all the commits since the last release, monorepo or not, as plugins downstream will need that.
Moreover your solution seems to be based on determining the files modified by a commit. It will probably be complex to handle situation in which a commit modify files across multiple sub-package. It's probably better to make the commit/package association based on naming convention, possibly reporting errors when a commit modify multiple packages (I don't know if such case is expected or not...).

The right way to handle monorepo might be via plugins:
- `analyzeCommit`: to parse the monorepo commit format
- `publish`: to do multiple releases of each sub-modules
- `generateNotes`: to generate a custom release note that group together the commits related to the same package

But all that depends mostly on monorepo specifics:
- Should we make one release note with different block for each package released ?
- Should we make one tag ? Or a tag per package released ? what if a commit span across packages?
- What happen if a new dependency is set in the root `package.json`? Does that mean all the sub-package have to release so they include the new dependency? Or is each sub-module has its own independent `package.json`?

pvdlg:
Related to #193

pmowrer:
Hi @pvdlg! Thanks for your review and very thorough comments!

I don't think we're at odds here. Addressing the three criteria you listed:

  1. The proposed [semantic-release-monorepo](https://github.com/Updater/semantic-release-monorepo) plugins are entirely **third party** and are not intended to be coerced onto `semantic-release` (I'd be happy to change the name if you feel it implies a relation to `semantic-release`). 

  2. The proposed change to `semantic-release` (this PR) **doesn't change (or propose to change) its functionality at all**. It simply allows customizing `semantic-release` further, `getCommits` being one of the few functions that **isn't already extracted into a plugin**. 
  
  3. The proposed workflow of `semantic-release-monorepo` is, as far as I can gather, entirely in-line with `semantic-release`'s core principles. The proposed workflow is simply scaling the `semantic-release` workflow/core principles across multiple packages.

To address your other points:

> There is currently multiple monorepo solutions: lerna, builder I think, yarn workspaces, zelda and many custom ones

I'm familiar with `lerna`, `builder`, `yarn workspaces` and `zelda`. I believe they're essential tools to making a monorepo work, but they handle dependency/task management across many packages, not automatic releases. They are complementary to `semantic-release`, allowing it to scale across a monorepo. In fact, I use `lerna` and `yarn workspaces` in the `semantic-release-monorepo` workflow.

> Moreover your solution seems to be based on determining the files modified by a commit. It will probably be complex to handle situation in which a commit modify files across multiple sub-package. 

In the proposed workflow, a commit can belong to multiple packages all at once (if the commit touched files across multiple packages). Personally, I don't see a problem with this. That said, this is intended as a third party plugin and not a workflow to be imposed on anyone. Simple use it if it fits your workflow.

> The right way to handle monorepo might be via plugins

This workflow could be achieved using the existing plugins, but it would require extra complexity and duplication. I could filter the commits passed to `analyzeCommits`, but I'd have to do the exact same filtering again to `generateNotes` since both plugins are passed the same commits coming from `getCommits`. If `getCommits` would be allowed to be customized the complexity would be reduced.

> semantic-release must continue to identify all the commits since the last release, monorepo or not, as plugins downstream will need that.

This PR does not interfere with this. The default behavior is still exactly what it was. Only through a third party plugin could this behavior be overridden.

> But all that depends mostly on monorepo specifics

I'd suggest leaving this to plugin implementors. `semantic-release-monorepo` is just a plugin to represent one approach.

pvdlg:
The objective of `semantic-release` plugin architecture is to support various workflow, third-parties  integration and to customize some rules. In order to keep a sane eco-system:
- the core has to do the decision making and the orchestration
- the plugins have to be limited to simple, well-defined tasks
- the plugins cannot be interdependents

Otherwise it will create a crazy amount of issue like "When I have plugin X, Y, Z this weird thing happen".

The goal is to allow extension and customization not to transform the core into an empty shell that run plugins sequentially.

The fundamentals of the `semantic-release` plugin architecture are:
- The core is in control, determine the logic and orchestrate
- The plugins (as in the exported function of a plugin module: `verifyConditions`, `getLastRelease` etc..) do one tasks
- The plugins have well-defined, simple input and outputs => Plugins know only the minimum they have to know to perform their simple task
- The core validate the output of each plugin
- The plugins cannot be interdependent: a plugin cannot call another plugin, and should perform its task with the input passed by the core, no matter what other plugin is used

Moving the commits retrieval to a commit would go against that, by creating plugin inter dependencies. In your implementation for example the `publish` would be dependent on the `getCommit`.

Moreover it doesn't make sense to exclude some commits. In git the commits (on the release branch) are sequential, and a release is the state at a moment of the commit history. In other terms a release is defined by a commit range, including everything from A (last release commit) to B (new release commit). Some commits within that range might not be written to the changelog, or not be considered to determine the version increase, but they are part of the release.

Finally, the commit retrieval part will evolve quite a lot in the near future, in order to handle workflow with multiple branches, dist-tags and pre-release. It will be in charge, among other things to detect merge/rebase commits, to determine if a new release has to be done or if an existing release has to be made available on a lower dist-tag.

For these reasons the commit retrieval part should stay in the core.

> They are complementary to semantic-release, allowing it to scale across a monorepo. In fact, I use lerna and yarn workspaces in the semantic-release-monorepo workflow.

To a certain extend yes. `semantic-release` and monorepo do very different things. But they both offer solutions to problems that overlap. As mentioned they both provide a way to make the release process easier for the developer, monorepo by doing multiple releases with one command, `semantic-release` by fully automating the process.

>  I could filter the commits passed to analyzeCommits, but I'd have to do the exact same filtering again to generateNotes since both plugins are passed the same commits coming from getCommits

We have implemented the concept of "multi-plugin" so a given plugin package can expose multiple plugin functions and share some code. For example in the [npm](https://github.com/semantic-release/npm) plugin the `verifyConditions`, `getLastRelease` and `publish` are exposed. They all use some shared code: `verifyNpm` and `getPkg`.
Keep in mind that they still have to be non interdependent. the `npm` `publish` can be used with a different implementation of the `getLastRelease`.
So if you need commit filtering function shared by `analyzeCommits` and `generateNotes` you can do that pretty easily.
I think what you want to achieve is possible with the current architecture and the plugin hooks available.

> semantic-release must continue to identify all the commits since the last release, monorepo or not, as plugins downstream will need that.
>> This PR does not interfere with this. The default behavior is still exactly what it was. Only through a third party plugin could this behavior be overridden.

What I mean is that it does change it because it allows to replace the commit retrieval by a different implementation

Sorry for the long message...It's a complex subject. Hopefully my explanations clarify a bit and provide reasonable points :)

pmowrer:
Thanks again for taking the time to write such in-depth replies. Very helpful! 😄 

> The fundamentals of the semantic-release plugin architecture are:
>
>* The core is in control, determine the logic and orchestrate
>* The plugins (as in the exported function of a plugin module: verifyConditions, getLastRelease etc..) do one tasks
>* The plugins have well-defined, simple input and outputs => Plugins know only the minimum they have to know to perform their simple task
>* The core validate the output of each plugin
>* The plugins cannot be interdependent: a plugin cannot call another plugin, and should perform its task with the input passed by the core, no matter what other plugin is used
>
>Moving the commits retrieval to a commit would go against that, by creating plugin inter dependencies. In your implementation for example the publish would be dependent on the getCommit.

Good point! It felt icky extracting `getCommits`. 

As an alternative, `filterCommits` could be a plugin after `getLastRelease`, a function that takes a predicate (commits -> `Boolean`) and returns a list of commits. It would enforce the options structure passed to the following plugins and avoid the interdependency. However, it'd be a much more specific plugin and I understand if you wouldn't want to pollute the plugin API.

> Moreover it doesn't make sense to exclude some commits. In git the commits (on the release branch) are sequential, and a release is the state at a moment of the commit history. In other terms a release is defined by a commit range, including everything from A (last release commit) to B (new release commit). Some commits within that range might not be written to the changelog, or not be considered to determine the version increase, but they are part of the release.

In the context of a monorepo, I think excluding commits from a release is fine. If the commit didn't touch files within a package, the commit isn't relevant to that package's release. The commits remain sequential, only some are "hidden" because they are irrelevant. I'm not sure I follow what your objection is. 

> So if you need commit filtering function shared by analyzeCommits and generateNotes you can do that pretty easily. I think what you want to achieve is possible with the current architecture and the plugin hooks available.

I've [refactored the implementation](https://github.com/Updater/semantic-release-monorepo/pull/1) according to your suggestion and it's fine. There are some drawbacks to it:

* The added complexity, as noted above.
* The need for optimization. The filtering must now be done twice, though that could be optimized away using some (somewhat icky) global state, as in the [npm plugin example you referenced](https://github.com/semantic-release/npm/blob/master/index.js#L6).

Given that my current implementation solves the problem just fine, I agree that it's not worth messing with the core architecture. Perhaps commit filtering is something to consider as the project evolves. For now, I am only experimenting with this workflow anyway. Thanks again for walking me through some of the project!

pvdlg:
Thanks for the update!

As mentioned we would be open to minor changes in the core to support monorepo. But the `commitFilter` you are proposing seems to be specific to your implementation of monorepo. For example (as far as I understand) `lerna` works differently and analyze all the commits and make multiple releases at once.

Unfortunately none of the maintainer knows monorepo solutions enough to be able to come up a comprehensive design.

In order to move forward we would like to have a proposal, in the the form of an issue with:
- A detailed explanation of the major monorepo solutions:
  - File structure and naming convention
  - Commit formats
  - Scripts used to make a release
  - Version number / Git tag format
- A detailed explanation of the release process
  - How to determine if a 1 or more release(s) have to be done
  - How to retrieve the last release(s), on npm and on git, especially in the case of multiple releases
  - How to parse the commits and details about the mapping commits <-> releases
  - How to make the release(s): change directory? run a script? what to do if one of the multiple release fail?
  - How to generate the changelog: One changelog per packaged released? Or concatenate all of them in one changelog? How many releases on Github? 

Then when we have all this information we can involve the community and discuss the best approach.

This idea is that if we support monorepo we do it properly:
- The solution has to generic enough so the core (and possibly the default plugins) would support any monorepo solution
- The specificities of each monorepo has to stay in plugin
- The changes in the core and the potential plugin input/output has to be defined beforehand so the community can work on plugin without constant breaking changes
- As monorepo are very complex solutions with a lot of constraints/worarround, we would need detailed documentation regarding how to implement a workflow for each major monorepo solution

It doesn't mean we have to support all solutions right away. But the design as to be generic and robust enough so anyone can create plugins for other/new monorepo solutions without changing the core.

I know it's a lot, but would you want to work that?

pmowrer:
Thanks again for the thorough write-up! 

For now, I'm going to continue experimenting with applying `semantic-release` to a monorepo using the plugin I wrote: https://github.com/Updater/semantic-release-monorepo. As it evolves, I'll think about what, if anything, can be added to the `semantic-release` core to better support the workflow.

PR #532, opened 2017-11-25
docs: Update readme for 0.x.x

unional:
Add to Q&A section on how to configure `semantic-release` for 0.x.x development.

I couldn't find this information until I'm almost ready to publish a `commit-analyzer-zero` for this purpose. 🌷

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,6 +287,26 @@ Being able to write code for just the most recent node versions greatly simplifi
 
 For a special purpose tool like `semantic-release`, that's only meant to be used in controlled CI environments, we think it's okay to have such a high version requirement. As `semantic-release` handles package publishing we expect almost every project to have at least one build job running node 8 already – and that's all it takes. Even if that's not that case `semantic-release` can still be executed with the help of [npx](https://www.npmjs.com/package/npx) (`npx -p node@8 npm run semantic-release`).
 
+### Does `semantic-release` work with Major Version Zero (0.x.x) development?
+
+Many package releases with Major Version Zero scheme (e.g. [ava](https://github.com/avajs/ava)). Can you use `semantic-release` for those package?
+
+Yes you can. All you need is configure the `release` block of `package.json` as follow:
+
+```json
+{
+  "release" {
+    "analyzeCommits": {
+      "releaseRules": [
+        { "breaking": true, "release": "minor" }
+      ]
+    }
+  }
+}
+```
+
+This mechanism is supported by the default [commit-analyzer](https://github.com/semantic-release/commit-analyzer/).
+
 ## Badge
 
 Use this in one of your projects? Include one of these badges in your README.md to let people know that your package is published using `semantic-release`.
--- a/README.md
+++ b/README.md
@@ -295,7 +295,7 @@ Yes you can. All you need is configure the `release` block of `package.json` as 
 
 ```json
 {
-  "release" {
+  "release": {
     "analyzeCommits": {
       "releaseRules": [
         { "breaking": true, "release": "minor" }


gr2m:
Hey @unional thanks for the pull request and the 💐😁 

We usually don’t recommend to use 0.x.x at all, just like npm does. All versions before `1.0.0` can be potentially breaking so it does not make much sense for semantic-release to be used. 

But on the other hand, this questions comes up every now and then. So I’m okay with adding a question to the FAQ and I love your workaround, I’m surprised how simple it is :)

I’d just change the wording a bit. I’d like to make the point that we don’t recommend to publish 0.x.x versions and why as a disclaimer. But if folks still want to do it, they can use your workaround. 

I’ll need a bit to think about the right wording and to lookup some resources, okay if we leave your PR open until then?

unional:
Sure, no problem. 🌷

unional:
I think using 0.x.x as its place and not necessarily a bad thing, especially for open source projects.

Open Source projects can easily introduce breaking changes, partly because the consumer is more remote compared to an internal project that is used within an organization. You get the heat from your colleagues breathing through your neck when you introduce a breaking change.

The other half is the author tends to experiment things for his own use cases and just want to share the work to the world so other people MAY benefit from it. But as more use cases are identified, it is very likely the design and architecture of the project will change multiple times before the "final" design emerge.

That's why we see (IMO) many open source projects uses the 0.x.x scheme until it is stable.
IMO, it is a nice way to state that the projects are still in "initial development" phase and breaking change can be more frequent than more matured projects.

Another reason I think open source projects use more of 0.x.x scheme because many of them are created in the spare-time. There are probably many features the author wants to add in but just doesn't have the time to do so.

If we move too soon to version 1.0.0, we will either see the major version craze (e.g. `jest` and `react`), or the innovation slows down because we want to keep the project more stable.

My two cents. 🌷

gr2m:
Ideally people loose the fair of increasing the major (breaking) version number. These numbers are for machines, not humans. There is really no difference if a project is on version 1.2.3 or 132.2.3, not for our code anyway. At least above 1.0.0, the semantics of each number is clearly specified.

unional:
Definitely. 👍 
I agree that for engineering (or, as engineers), we should un-tie our emotion from the version number.

But coming from business and company perspective, the stability of versions and the perception of "stable major" does have some value, even thou the value can be artificial or even "harmful" (as it slows down innovation). 😆 

When a business sees a version `45.0.0`, and the version follows SemVer, one response would be: "this library introduce breaking changes 45 times in its lifespan. That means it breaks its consumer at least 45 times in the past. Should I let my team investing time in using this library? How likely would it introduce another breaking change that would cost both time and money for my business?"

> At least above 1.0.0, the semantics of each number is clearly specified.

If there is a SemVer 2.1 that improves the clarity of 0.x.y ... 😄 

Again, I'm in sync with you on promoting proper versioning, using SemVer, and let the machine deal with versioning. Just adding my two cents on the value of having better clarify and support of 0.x.y. 🌷 

Some mumbling:

Uncle Bob mentioned that the maturity of a library can be measured by how many libraries it depends on, vs how many libraries it is depended upon. But in open source projects, this is less true because the author can introduce breaking changes and bump the major version anytime.

At the end, the chance of introducing breaking changes is correlated to the complexity of the library. Mitigate strategies include: separation of concern, cohesion, coupling, dependency management.... i.e. better architecture and design at the component level.

unional:
> author can introduce breaking changes and bump the major version anytime.

I'm saying this not because of ranting. It is because I do that many times myself on my open source projects to figure out better naming and better API.

gr2m:
no worries, doesn’t feel like ranting at all :) I’ll get back to you with feedback on the wording. Thanks again for the pull request and insights into your perspective, much appreciated

pvdlg:
Just to clarify a bit regarding the problem of having too frequent breaking change releases. Using `0.x.x` version will not change anything vs having a `1.x.x` version. 
If you configure your dependency in the `package.json` with `^0.1.0` for example than each release that change the second number will be consider a breaking by npm and the user will to update manually the version.
In other terms for `npm`, `0.1.0 => 0.2.0` and `1.0.0 => 2.0.0` are both considered a breaking change release.
 
If you are concerned about "spamming" your users with too frequent major release, you could use `dist-tag`. From the README.md:

> If you fear the loss of control over timing and marketing implications of software releases you should know that `semantic-release` supports [release channels](https://github.com/npm/npm/issues/2718) using `npm`’s [dist-tags](https://docs.npmjs.com/cli/dist-tag). This way you can keep control over what your users end up using by default, you can decide when to promote an automatically released version to the stable channel, and you can choose which versions to write blogposts and tweets about. You can use the same mechanism to [support older versions of your software](https://gist.github.com/boennemann/54042374e49c7ade8910), for example with important security fixes.

You could configure `semantic-release` like this:
```json
"plublishConfig": {
  "tag": "next"
}
```
This way all releases will be done on the `next` dist-tag. That means only the user that explicitly choose to use the `next` dist-tag will get the breaking change.
When you want to make a release available for everyone then you can add it to the default `latest` dist-tag.

That would solve the problem of having too frequent breaking changes for your users.

See this article for more details: https://blog.greenkeeper.io/one-simple-trick-for-javascript-package-maintainers-to-avoid-breaking-their-user-s-software-and-to-6edf06dc5617

If the problem is the "emotional" part attached to high version numbers then you could create a `next` branch and release from`master` only when you decide.
You would have to set `next` as your default branch on Github and push your code there.
When you are ready to make a release, you would merge `next` into `master`. That would trigger a release process and `semantic-release` would release based on the commits since the last release.

In the early stage of your project, when you break the API very often, you would do a release less frequently. Once your project is stable you can start making releases more often, maybe even committing directly on `master` and make commit on each push.

pvdlg:
Here more details about the reason we favor starting at `1.0.0`: [Introduction to SemVer](https://blog.greenkeeper.io/introduction-to-semver-d272990c44f2)

Closing in favor of #566

gr2m:
Thanks for sharing your perspective on this. I’m sorry that we can’t make everyone happy, we have to make choices somewhere, and we cannot change it in regards of 0.x releases.

Same holds true for "version numbers are for machines, not humans". We still try to keep breaking versions to a minimum and there are things we plan to do to reduce the maintenance overhead for supporting things like LTS version and scheduled releases.

jwalton:
That's ok.  We can still be friends.  ;)  I mean, the reason I'd like to see something like this in the FAQ is only because I'm in the middle of writing a big library that's not at all ready for public consumption, but I'm publishing it to npm because I'm using it in another project in parallel with developing it (which is finding lots of friction points in the API), and while I've only just started using semantic-release, it's completely ruined me and it makes me angry that I now have to waste time typing `npm version patch && git push --follow-tags && npm publish` on this project over and over again.  You've completely ruined me.  :)

travi:
i've resorted to using pre-releases for new packages that i consider to still be in "beta". i try to keep that period short, but having the very rough initial iterations come before releasing `v1.0.0` has still been valuable for me.

like you, @jwalton, i've become very spoiled by `semantic-release`, so i'm very much looking forward to when [`semantic-release` might be able to help me with my pre-releases](https://github.com/semantic-release/semantic-release/issues/563)

gr2m:
I hear you, I’ve been there. The only cure is too loose your hesitation about 1.0.0 and major releases in general :) It was hard but things got so much easier afterwards. You can add a clear notice in the README that the module is still experimental.

We also have an FAQ entry here: https://semantic-release.gitbooks.io/semantic-release/docs/support/FAQ.html#can-i-set-the-initial-release-version-of-my-package-to-001

unional:
Having used `semantic-release` for more than 6 months and on multiple packages, IMO the currently recommended workflow is not optimal.

I'm planning to write up an article on it, but don't know when I'll have the time to.

The main issue with the current workflow is that it assumes "deliverables are releasable".
But the fact is this is just not true.

Other issues include the workflow on `dist-tag` is not established, and the modification described in this PR is needed for it.

I understand the stand of the team on "version numbers are for machines, not humans", but I disagree.
The word "semantic" itself means "meaning" and it directly implies those meaning carried in the "semantic versioning" should be meaningful to both human and machines.

travi:
> The main issue with the current workflow is that it assumes "deliverables are releasable".
But the fact is this is just not true.

I'm interested in what is meant by this statement. 

the continuous delivery philosophy is essentially that `master` is always releasable, but not necessarily deployed/published. `semantic-release` enables going the step beyond continuous delivery to continuous deployment, where you might as well deploy/publish the value that has been delivered to `master` automatically.

are you referring to something else?

unional:
One way to explain this is in Scrum, at the end of a sprint you have a deliverable product. But that does not mean you will release it.

It's similar to your wording that `master` is releasable but not released/deployed/published.
But the term `releasable` is confusing as it suggests the code is at a stage that can be released, but not necessarily true.

For example, you are writing a web app. Completing a feature "login" makes it deliverable (as all tests pass), but not releasable because it does not support an use case that is valuable to the user.

The same is true for libraries. You may complete a story of an epic, but until all stories in that epic are completed, it cannot be released.

travi:
> is being published to the wild

correct, which is why it is important for _all_ commits in master to be safe for being published. keep in mind that not all commits trigger `semantic-release` to publish. in my case i use a type of `wip` (as opposed to `feat` or `fix`) in commits that are forward progress, but incomplete, that i want to get merged into `master`. you do need to make sure it is "safe" to publish because another commit could trigger a publish, but not everything has to trigger the publish immediately after a merge to `master`

> So with no human involved, this difference ("while not necessarily deployed/published, it could be at any time") is muted.

right, my point above is that the primary purpose of using `semantic-release` is to enable continuous deployment, not just continuous delivery. if that is not your goal, `semantic-release` might not be a good fit

> Your comment basically suggesting all PRs on github should not merge to master, because having one PR merged does not mean the code is "publishable" in your term.

this depends on what is considered mergeable. i am suggesting that it a contribution would result in something that would be problem if published, it shouldn't be merged to master. 

> Or, you will have to keep each PR around to collect many more changes until it is in a "feature-completed" state

this is one technique for keeping partial implementations that shouldn't be published out of `master`. however, just because it is not "complete" doesn't mean that it can't be published without causing a problem. incomplete and "dark" may not be a valuable publish, but it could be published without being noticed by a consumer. again, the practice of continuous integration is very much about the careful integration of as much of an (even incomplete) implementation into `master`, while carefully ensuring that the bit that would make the feature noticeable to a consumer is not included.

it's completely up to you to decide what workflow is the best fit for your project and team, but `semantic-release` is pretty clearly targeted at continuous-deployment and is _very_ good at it. if continuous deployment is not the desired workflow, i'm not sure what value would be expected from using `semantic-release`

esciara:
Whilst I did not all comments (will catch up in details later), I just wanted to point that a workaround exists: create a `v0.0.0` tag will make `semantic-release` bump the version from there. (in my case, it became `v0.0.1` at the first release...). But thinking about it, using suffixes like `pre-alpha`, `alpha` et al might also be a good approach... (hope I did not restart a bitter discussion on the subject... 😅 )

PR #553, opened 2017-12-10
Shareable config

### LINKED ISSUE #549: Support shareable config
this has probably been discussed before, but i'm not finding any info about it if it has. semantic-release provides a considerable amount of customizability, but i've always been hesitant to get too deep because i didn't want to take on deviating from the default conventions consistently across all repos where my organization(s) use semantic-release. 

i think i'd be far more interested in tweaking semantic-release further if doing so could apply across all of the projects that i manage fairly automatically.

at minimum, this would enable a config to be published as an npm package, such as `semantic-release-config-travi` and used in each semantic-relase managed project using something along the lines of `extends: 'travi'`. ideally, a config could also extend another config, such as `semantic-release-config-greenkeeper-keeper` extending `semantic-release-config-travi`, allowing projects under the `greenkeeper-keeper` org to extend `semantic-release-config-greenkeeper-keeper` but have that mostly configured from my personal config.

has the concept of a shareable config, similar to what [eslint enables](https://eslint.org/docs/developer-guide/shareable-configs) been considered for semantic-release? would something like this be a huge undertaking to add?

pvdlg:
Fix #549

#### Global plugin options
Currently users are forced to redefined all plugin in the pipeline to define a plugin options. That creates a lengthy option object and its error prone.
In addition several plugin uses the same options (i.e. `githubUrl`).

It's now possible to defined options directly under the `release` property, that will be passed to each plugins.
It's still possible to define options within the plugin definition to have it passed only to a specific plugin instance.

#### Shareable config
Adds the options `extends`, which can be defined via configuration file or CLI arguments to a single path or an array of paths of shareable configuration.
A shareable configuration is a file or a module that can be loaded with `require`.
Options is defined by merging in the following order of priority:
- CLI/API
- Configuration file
- Shareable configuration (from right to left)

Options set in a shareable configuration can be unset by setting it to `null` or `undefined` in the main configuration file. If a default value applies to this property it will be used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,9 @@ _[This is what happens under the hood.](https://github.com/semantic-release/cli#
 
 ## Options
 
-You can pass options either via command line (in [kebab-case](https://lodash.com/docs#kebabCase)) or in the `release` field of your `package.json` (in [camelCase](https://lodash.com/docs#camelCase)). The following two examples are the same, but CLI arguments take precedence.
+You can pass options either via command line (in [kebab-case](https://lodash.com/docs#kebabCase)) or in the `release` field of your `package.json` (in [camelCase](https://lodash.com/docs#camelCase)). Alternatively the configuration can also be defined in `.releaserc.yml`, `.releaserc.js`, `.releaserc.js` or `release.config.js`.
+
+The following two examples are the same, but CLI arguments take precedence.
 
 ##### CLI
 ```bash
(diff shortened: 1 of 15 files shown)


pvdlg:
Here is an example to what could be done with shareable configs: https://github.com/shroudedcode/semantic-release-apm/issues/4#issuecomment-353409057

Ideally when that is merge we could create an `apm-config` that would serve as an example. It would include:
- the config itself
- unit test of the config (using the `verifyConditions`) of each plugin
- integration tests with a release end to end

We might also consider making public the `semantic-release` API to load the config and run the `verifyConditions` pipeline. That would make shareable config unit test a lot easier.

pvdlg:
I had to do some more changes since your reviewed, in order to support loading plugins defined in a shareable config from that context.
That will allow to have shareable config, defining the plugin they depends on in `dependencies` so they don't have to be added to the `package.json` of the user using the shareable config.

PR #573, opened 2017-12-22
Add recipes

felixfbecker:
I love how powerful semantic-release 11 is. It enables so many workflows, especially with third-party plugins. But figuring out the configuration needed to get all of them working was quite a bit of work, and I feel like we should share this so future users don't have to figure it out themselves.
Not all config possibilities will become part of the CLI and it is way easier to share some examples than extending a CLI.

That's why I think it's a good idea to add a folder of recipes to the repo, that people can refer to and extend with PRs. It should contain both information of how to configure semantic-release and CI to enable specific workflows, but also link to external plugins to give them some visibility.

This is much inspired by prior art:
https://github.com/gulpjs/gulp/tree/master/docs/recipes

### DIFF
--- a/docs/recipes/README.md
+++ b/docs/recipes/README.md
@@ -1,9 +1,14 @@
 # Recipes
 
 ## CI configurations
-- [CircleCI 2.0 workflows](circleci-workflows.md)
-- [Travis CI](travis.md)
-- [Travis CI with build stages](travis-build-stages.md)
-- [GitLab CI](gitlab-ci.md)
+
+* [CircleCI 2.0 workflows](circleci-workflows.md)
+* [Travis CI](travis.md)
+* [Travis CI with build stages](travis-build-stages.md)
+* [GitLab CI](gitlab-ci.md)
 
 ## Package managers and languages
+
+* [Docker](https://github.com/felixfbecker/semantic-release-docker)
+* [PHP / Composer](php-composer.md)
+* [Visual Studio Code extensions](https://github.com/raix/semantic-release-vsce)
--- a/docs/recipes/circleci-workflows.md
+++ b/docs/recipes/circleci-workflows.md
@@ -1,1 +1,75 @@
-# CircleCI 2.0 workflows
+# Using semantic-release with [CircleCI 2.0 workflows](https://circleci.com/docs/2.0/workflows/)
+
+![CircleCI workflows screenshot](https://github.com/circleci/circleci-docs/raw/928cbb042453ae182996ce05f4e4042b02a24634/jekyll/assets/img/docs/workflow_detail.png)
+
+Similar to Travis build stages, CircleCI 2.0 workflows are a great way to run semantic-release in an independent build step.
+We can let CircleCI take care of only running a release on master, after tests passed and running it on an image with Node 8 - your test jobs could use a completely different environment, your project may not even be written in JavaScript.
+
+Make sure to set `NPM_TOKEN` and `GH_TOKEN` in Circle's project settings UI.
+
+### Annotated example `.circleci/config.yml`
+
+```yml
+version: 2
+
+jobs:
+  # Test on Node 6
+  test_node_6:
+    docker:
+      - image: circleci/node:6
+    steps: *test_steps
+  # Test on Node 8
+  test_node_8:
+    docker:
+      - image: circleci/node:8
+    steps: *test_steps
+  # Release if both test jobs passed
+  release:
+    docker:
+      # Use Node 8 for releasing
+      - image: circleci/node:8
+    steps:
+      - checkout
+      - *restore_cache
+      - run: npm install
+      - *save_cache
+      - run: npm run semantic-release
+
+# Define when/under what conditions the jobs should be executed
+workflows:
+  version: 2
+  chromeless:
+    jobs:
+      # Two jobs to test on Node 6 and Node 8
+      - test_node_6
+      - test_node_8
+      # Release job that runs only on master if the two test jobs succeeded
+      - release:
+          requires:
+            - test_node_6
+            - test_node_8
+          filters:
+            branches:
+              only: master
+
+# The steps shared between the two test jobs
+test_steps: &test_steps
+  - checkout
+  - *restore_cache
+  - run: npm install
+  - *save_cache
+  - run: npm test
+
+# Restore npm cache (important to make the jobs start up fast)
+restore_cache: &restore_cache
+  restore_cache:
+    keys:
+      - npm-cache-{{ checksum "package-lock.json" }}
+
+# Save npm cache (important to make the jobs start up fast)
+save_cache: &save_cache
+  save_cache:
+    key: npm-cache-{{ checksum "package-lock.json" }}
+    paths:
+      - ~/.npm
+```
--- a/docs/recipes/travis-build-stages.md
+++ b/docs/recipes/travis-build-stages.md
@@ -1,1 +1,63 @@
-# Using semantic-release with [Travis CI build stages](https://docs.travis-ci.com/user/build-stages)
+# Using semantic-release with [Travis build stages](https://docs.travis-ci.com/user/build-stages/)
+
+![Travis Build Stages screenshot](https://cloud.githubusercontent.com/assets/3729517/25229553/0868909c-25d1-11e7-9263-b076fdef9288.gif)
+
+Build stages are a great way to run semantic release compared to running it in `after_success`:
+
+* The release stage can use Node 8, while the other stages can use other versions, even use a completely different language.
+* It allows you to easily release a project that is not even written in JavaScript, without impacting the job configuration of your test jobs in any way.
+* Instead of having to figure out a "build leader" to do the release and poll the Travis API to wait until all jobs passed, Travis can handle only running the release step after the other jobs passed.
+* If the release failed, you can see at a glance that only the release stage failed while your test stage succeded.
+* Travis can take care of not running the whole release stage at all on non-master branches and PRs.
+
+### Annotated example `.travis.yml`
+
+```yml
+# This could even be a different language
+language: node_js
+
+# Define any version matrix you want for testing
+node_js:
+  - '8'
+  - '6'
+
+# This is inherited by all jobs
+cache:
+  directories:
+    - ~/.npm
+
+# This is the script that will only run for your test stage
+script:
+  - npm run build
+  - npm run lint
+  - npm test
+
+jobs:
+  include:
+    - stage: release
+      # Set the language to node if you used a different language at the top
+      # Otherwise this will be inherited
+      language: node_js
+      # Explicitely use only Node 8 for release (no matrix)
+      node_js: '8'
+      # Override script to build and release
+      # We use script, not after_success, so we get notified if the release failed for unexpected reasons.
+      script:
+        # Don't forget to build before releasing if your project has a build step.
+        # The release stage doesn't share state with the test stage.
+        # Alternatively you could add a prepublishOnly script to package.json
+        - npm run build
+        - npm run semantic-release
+
+# Define when/under what conditions the jobs should be executed
+stages:
+  - test
+    # Make sure that semantic-release only runs on master and not on PR builds
+  - name: release
+    if: branch = master AND type = push AND fork = false
+
+# Don't build tags
+branches:
+  except:
+   - /^v\d+\.\d+\.\d+$/
+```
--- a/docs/recipes/php-composer.md
+++ b/docs/recipes/php-composer.md
@@ -1,0 +1,72 @@
+# 🐘 Using semantic-release to release a PHP/Composer package
+
+You can use semantic-release to release a [Composer](https://getcomposer.org/) package to [packagist.org](https://packagist.org/).
+Composer works with GitHub repositories, so all that needs to happen is tag a new release on GitHub.
+The last release can be figured out by looking at the git tags of the repository.
+
+### Example annotated `package.json`
+
+```js
+{
+  "scripts": {
+    "semantic-release": "semantic-release"
+  },
+  "release": {
+    // Configure verifyConditions to not check conditions for npm publishing
+    "verifyConditions": [
+      "@semantic-release/github"
+    ],
+    // Get the last release from local git tags (instead of from the npm registry)
+    "getLastRelease": "@semantic-release/git",
+    // Only create a release and tag on GitHub (don't publish to npm)
+    "publish": "@semantic-release/github"
+  },
+  "devDependencies": {
+    "@semantic-release/git": "^2.0.0",
+    "semantic-release": "^12.0.0"
+  }
+}
+```
+
+### Example annotated `.travis.yml`
+
+```yml
+language: php
+
+# Makes jobs install dependencies faster
+cache:
+  directories:
+    - $HOME/.composer/cache
+    - $HOME/.npm
+
+jobs:
+  include:
+    # The test stage expands to two jobs on PHP 7.0 and 7.2
+    - stage: test
+      php:
+        - '7.0'
+        - '7.2'
+      install:
+        - composer install --prefer-dist
+      script:
+        - vendor/bin/phpunit
+
+    # The release stage is running Node 8
+    - stage: release
+      language: node_js
+      node_js: '8'
+      script:
+        - npm run semantic-release
+
+stages:
+  - test
+  - name: release
+    if: branch = master AND type = push AND fork = false
+
+# Don't build tags
+branches:
+  except:
+   - /^v\d+\.\d+\.\d+$/
+```
+
+This configuration can be adapted for other CI providers too, see the other recipies for examples.


docs/recipes/README.md
```diff
@@ -0,0 +1,9 @@
+# 📄 Recipes for workflows with semantic-release
+
+* CI configuration
```
pvdlg: Maybe it would be better to use heading in the readme?
Something like this:
```
## CI configuration
### [Travis CI build stages](./travis_build_stages.md)
```

docs/recipes/circleci_workflows.md
```diff
@@ -0,0 +1,95 @@
+# Using semantic-release with [CircleCI 2.0 workflows](https://circleci.com/docs/2.0/workflows/)
+
+![CircleCI workflows screenshot](https://github.com/circleci/circleci-docs/raw/928cbb042453ae182996ce05f4e4042b02a24634/jekyll/assets/img/docs/workflow_detail.png)
+
+Similar to Travis build stages, CircleCI 2.0 workflows are a great way to run semantic-release in an independent build step.
+We can let CircleCI take care of only runnin a release on master, after tests passed and running it on an image with Node 8 - your test jobs could use a completely different environment, your project may not even be written in JavaScript.
```
pvdlg: `runnin` => `running`

docs/recipes/php_composer.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Example annotated `package.json`
+
+```js
+{
+  "scripts": {
+    "semantic-release": "semantic-release"
+  },
+  "release": {
+    // Configure verifyConditions to not include condition-travis (we don't need it because we use Travis build stages)
```
pvdlg: In order to make this recipe simpler and more focused on `Composer` maybe it be better to not mix Travis Build Stage and keep the `conditions-travis` plugin.
Not sure if it's possible though. I think node and npm are installed by default on each travis images, so you might be able to do something like that:
```yaml
php:
  - '7.0'
  - '7.2'

install:
  - composer install --prefer-dist
script:
  - vendor/bin/phpunit

after_success:
  - npx semantic-release
```

docs/recipes/travis_build_stages.md
```diff
@@ -0,0 +1,90 @@
...
+  - npm test
+
+jobs:
+  include:
+    - stage: release
+      # Set the language to node if you used a different language at the top
+      # Otherwise this will be inherited
+      language: node_js
+      # Explicitely use only Node 8 for release (no matrix)
+      node_js: '8'
```
pvdlg: Here you can configure with `node_js: node` so the last Node stable will be used and conditions-travis will always pick this job as the build leader, without having to add `BUILD_LEADER`.

docs/recipes/php_composer.md
```diff
@@ -0,0 +1,73 @@
...
+      script:
+        - npm run semantic-release
+
+stages:
+  - test
+  - name: release
+    if: branch = master AND type = push AND fork = false
+
+# Don't build tags
+branches:
```
pvdlg: We would prefer to recommend this config for `branches`
```yaml
branches:
  only:
    - master
    - /^greenkeeper.*$/
```
By default Travis do a push and PR build on a PR. This way that avoid doing 2 builds and run only the PR build, except if this is a Greenkeeper branch.

docs/recipes/php_composer.md
```diff
@@ -0,0 +1,73 @@
+# 🐘 Using semantic-release to release a PHP/Composer package
+
+You can use semantic-release to release a [Composer](https://getcomposer.org/) package to [packagist.org](https://packagist.org/).
+Composer works with GitHub repositories, so all that needs to happen is tag a new release on GitHub.
+The last release can be figured out by looking at the git tags of the repository.
+
+### Example annotated `package.json`
```
pvdlg: `semantic-release` now support various config files. As php project typically do not use `package.json` maybe it would be better to make the `semantic-release` config in `releaserc.yml` or `releaserc.json` ?

docs/recipes/travis_build_stages.md
```diff
@@ -0,0 +1,90 @@
...
+  directories:
+    - ~/.npm
+
+# This will be inherited in the release stage
+install:
+  - npm install
+
+# This is the script that will only run for your test stage
+script:
+  - npm run build
```
pvdlg: Can you instead recommend to build in a `pretest` and in a `prepublishOnly` script?
This way that avoid forgetting the build step, both on the CI and for contributor who fork the repo.

docs/recipes/travis_build_stages.md
```diff
@@ -0,0 +1,90 @@
...
+  - npm run lint
```
pvdlg: Can you instead recommend to lint in a `pretest` script?
This way contributors are less likely to push commits without linting?

docs/recipes/travis_build_stages.md
```diff
@@ -0,0 +1,90 @@
...
+      # Otherwise this will be inherited
+      language: node_js
+      # Explicitely use only Node 8 for release (no matrix)
+      node_js: '8'
+      # Override script to build and release
+      # We use script, not after_success, so we get notified if the release failed for unexpected reasons.
+      script:
+        # Don't forget to build before releasing if your project has a build step.
+        # The release stage doesn't share state with the test stage.
+        - npm run build
```
pvdlg: Same as above. The build step would be run automatically if set in a `prepublishOnly` script.

docs/recipes/travis_build_stages.md
```diff
@@ -0,0 +1,90 @@
...
+        - npm run semantic-release
+
+# Define when/under what conditions the jobs should be executed
+stages:
+  - test
+    # Make sure that semantic-release only runs on master and not on PR builds
+  - name: release
+    if: branch = master AND type = push AND fork = false
```
pvdlg: No need for this filter as it's handled by the `conditions-travis` plugin.

docs/recipes/travis_build_stages.md
```diff
@@ -0,0 +1,90 @@
...
+
+# Don't build tags
+branches:
```
pvdlg: Same comment as above:
```yaml
branches:
  only:
    - master
    - /^greenkeeper.*$/
```

docs/recipes/travis_build_stages.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Annotated example `package.json`
+
+```js
+{
+  "scripts": {
+    "semantic-release": "semantic-release"
+  },
+  "release": {
+    // Configure verifyConditions to not include condition-travis
```
pvdlg: No need to disable `condition-travis` as it supports Travis Build stages

docs/recipes/travis_build_stages.md
```diff
@@ -0,0 +1,90 @@
...
+node_js:
+  - '8'
+  - '6'
+
+# This is inherited by all jobs
+cache:
+  directories:
+    - ~/.npm
+
+# This will be inherited in the release stage
```
pvdlg: Isn't it the default? I don't think it's necessary. The release Build Stage will use the same as its the default.

docs/recipes/travis_build_stages.md
```diff
@@ -0,0 +1,90 @@
+# Using semantic-release with [Travis build stages](https://docs.travis-ci.com/user/build-stages/)
+
+![Travis Build Stages screenshot](https://cloud.githubusercontent.com/assets/3729517/25229553/0868909c-25d1-11e7-9263-b076fdef9288.gif)
+
+Build stages are a great way to run semantic release compared to running it in `after_success`:
+
+* The release stage can use Node 8, while the other stages can use other versions, even use a completely different language.
+* It allows you to easily release a project that is not even written in JavaScript, without impacting the job configuration of your test jobs in any way.
+* Instead of having to figure out a "build leader" to do the release and poll the Travis API to wait until all jobs passed, Travis can handle only running the release step after the other jobs passed.
```
pvdlg: Here instead you can mentioned that using `node_js: node` will guarantee that the release will happen on that job.

docs/recipes/travis_build_stages.md
```diff
@@ -0,0 +1,90 @@
+# Using semantic-release with [Travis build stages](https://docs.travis-ci.com/user/build-stages/)
+
+![Travis Build Stages screenshot](https://cloud.githubusercontent.com/assets/3729517/25229553/0868909c-25d1-11e7-9263-b076fdef9288.gif)
+
+Build stages are a great way to run semantic release compared to running it in `after_success`:
+
+* The release stage can use Node 8, while the other stages can use other versions, even use a completely different language.
+* It allows you to easily release a project that is not even written in JavaScript, without impacting the job configuration of your test jobs in any way.
```
pvdlg: Can Travis runs `npx semantic-release` on a non Node image? If yes this comment is irrelevant.

pvdlg:
Thanks very much. That's a great idea!

docs/recipes/php_composer.md
felixfbecker: Not sure what the advantage of that would be. You still need to install semantic-release, so you need to declare the dependency. You could just install it imperatively in CI of course but then you can't use Greenkeeper/dependencies.io/Renovate to update your deps. And given PHP is primarily used for web applications I think PHP people are familiar with npm/package.json for frontend frameworks.

pvdlg: That's a good occasion to mention that other type of config files can be used.
`semantic-release` could be used with `npx semantic-release`, which avoid having to go through the `npm install` step.
Greenkeeper can also be used with `npx` I think. I don't know about renovate.

That also avoid to configure all the mandatory fields in `package.json`.

docs/recipes/travis_build_stages.md
felixfbecker: Yeah you're right

docs/recipes/travis_build_stages.md
felixfbecker: Imo that is not really feasible for any medium-sized project and not the convention at least from what I've seen in build setups (TS, webpack, Babel, ...). When developing, you want to use the watch mode of your build pipeline. Running _another_ build before every test run is just wasted time (it could be anywhere between a few seconds to over a minute). Travis also collapses each command separately so it's easier to see what command actually failed. I think it would also make the example a bit more complex because I would have to highlight that in package.json.

In general, this is just a recipe and as such you are of course always free to not follow it 100%. All of these are mostly directly copied from real projects.

docs/recipes/travis_build_stages.md
felixfbecker: I think tools like lint-staged and husky are great for catching lint errors before pushing but CI always needs to repeat the steps and it's kinda out of scope for this example. This is just copied from a real project where linting takes a significant amount of time that you don't want to spend on every test run.

docs/recipes/travis_build_stages.md
felixfbecker: That would make the build irreproducible and prone to breakage. What if semantic-release doesn't work with the next major Node version?

docs/recipes/travis_build_stages.md
felixfbecker: Then after all test jobs succeeded on a PR, Travis would spin up a container for the release stage, install the correct Node version, install dependencies, just so condition-travis notices it is on master and doesn't need to run - wasting build time. Seems pointless?

docs/recipes/travis_build_stages.md
felixfbecker: Yeah wasn't sure about this one. I prefer it too but some people may be confused if their branches don't build. The CLI generates the `except` config so I chose that

docs/recipes/php_composer.md
felixfbecker: Not Node 8 though, you'd have to run `nvm install 8; nvm use 8` first, and then `npm install`. I think using build stages lowers the barrier especially for any non-JS project because you can just "drop in" that totally isolated release stage that is completely independent of the rest of your build config.

docs/recipes/php_composer.md
felixfbecker: Yeah it's a good idea to mention it.
How would Greenkeeper+npx work? Do you have an example?

docs/recipes/travis_build_stages.md
felixfbecker: Not afaik because npx was introduced in npm 5. Also are you sure that every image has Node preinstalled?

docs/recipes/README.md
felixfbecker: I don't think that would look better, because the headings would have no indentation when rendered. This is intended to be a ToC

docs/recipes/travis_build_stages.md
felixfbecker: Not really till https://github.com/semantic-release/condition-travis/issues/96 is resolved, right?

docs/recipes/php_composer.md
gr2m: Greenkeeper no longer needs to install any dependency, it works as a GitHub app that needs to be installed on the repo. It does require a `package.json` file though. But GK is only relevant for node apps

docs/recipes/php_composer.md
pvdlg: ok, works for me

docs/recipes/php_composer.md
felixfbecker: Yeah, that's what I meant. Just using `npx` in .travis.yml without package.json won't work with updates.

I disagree that it is only relevant for Node apps. Any user who uses semantic-release may want to auto-update semantic-release. There are also tools like dependencies.io or dependabot that work for both Node/npm and PHP/Composer, so there would be no reason not to use it for both in a project.

So imo using a package.json is a better practice to recommend.

docs/recipes/travis_build_stages.md
```diff
@@ -0,0 +1,90 @@
...
+
+# This will be inherited in the release stage
+install:
+  - npm install
+
+# This is the script that will only run for your test stage
+script:
+  - npm run build
+  - npm run lint
+  - npm test
```
gr2m: I suggest to leave out the `script` part altogether, I think the receipt should focus on the semantic-release, weather you have a build or lint script that you want to run next to `npm test` depends on the respective project

docs/recipes/travis_build_stages.md
gr2m: Agree. Maybe change the comment though to something like "Use the latest Node LTS version"

docs/recipes/travis_build_stages.md
felixfbecker: I wanted to highlight that you must not forget to run build again in the release step if you have a build process. I've forgot to do that a couple of times and ended up releasing uncompiled packages

docs/recipes/travis_build_stages.md
gr2m: agreed

docs/recipes/travis_build_stages.md
pvdlg: As a side there is something that is quire related to `semantic-release` and that has been ask several time on the issue tracker: Having the version set in package.json before running the build script.
This is because some build script need the version to be set to do something with it.

In such situation the best way to solve it is to have the build script being triggered in `prepublishOnly`.

That might a good candidate for another recipe though

felixfbecker: Yeah, that would be a valid use case that could fit into a recipe.

docs/recipes/travis_build_stages.md
gr2m: I agree with Pierre. I see your point Felix, but I think you try to solve multiple things at once with these instructions. The main objective is to help folks setup `semantic-release`. The `: branch = master AND type = push AND fork = false` is an optimization ahead of where the user currently is. The less other code we show, the simpler it will be to follow instructions.

docs/recipes/travis_build_stages.md
gr2m: I’d prefer the `except`, too. Keep it simple. Maybe folks haven’t heard of greenkeeper and it adds to the mental overhead while learning to setup semantic-release

docs/recipes/travis_build_stages.md
gr2m: It’s addressed via semantic-release/travis-deploy-once#30, let’s leave it in

docs/recipes/php_composer.md
gr2m: okay, I’m lacking the experience with non-Node projects, I feel like we are guessing here :) let’s go ahead with it and see how folks adapt it and what trouble they run in 👍

docs/recipes/README.md
pvdlg: Comment to be ignored. Let's keep it that way.

pvdlg: Ok

docs/recipes/travis_build_stages.md
pvdlg: Ok, let's forget about `npx`

docs/recipes/travis_build_stages.md
pvdlg: Ok for putting that in another recipe

docs/recipes/travis_build_stages.md
pvdlg: Ok. The build defined in the release Stage will be determined as the build leader automatically due semantic-release/travis-deploy-once#30. No need to specify `node_js: node`

docs/recipes/travis_build_stages.md
pvdlg: Ok, let's keep what the CLI generates

docs/recipes/circleci_workflows.md
pvdlg: Change still needed

docs/recipes/travis_build_stages.md
pvdlg: Maybe we should remove this sentence? I'm not sure the internal mechanics of the "non build stage" is relevant for the reader?

docs/recipes/travis_build_stages.md
pvdlg: Let's remove for simplification, it as it's the default then.

docs/recipes/travis_build_stages.md
pvdlg: Per @gr2m comment

docs/recipes/travis_build_stages.md
pvdlg: Similar to https://github.com/semantic-release/semantic-release/pull/573/files#r158536677

docs/recipes/travis_build_stages.md
pvdlg: Per @gr2m comment

docs/recipes/travis_build_stages.md
pvdlg: Per @gr2m comment

pvdlg:
There is a lot of comment on this PR, many (from me) that are not relevant anymore. Unfortunately Github doesn't allow to remove a single change request in a review.... 
So I dismissed my review and created a new one with only the actual changes requested. There should be 7. All other comments can be ignored. Hopefully that helps...

pvdlg:
@felixfbecker would you like to update this PR with the changes requested and adapt it to fit in the new doc?  I can also cherry pick your commit and make the changes (that you can review).
Let me know.

docs/recipes/travis_build_stages.md
felixfbecker: I think it does matter because the Travis UI will show better what is going on, vs having the first job pending until all jobs completed. And it is more reliable than polling for sure.

docs/recipes/travis_build_stages.md
felixfbecker: Keeping this now that condition-travis is not run by default anymore

felixfbecker:
Updated the PR, should be good like this

docs/recipes/travis-build-stages.md
```diff
@@ -1 +1,63 @@
...
+Build stages are a great way to run semantic release compared to running it in `after_success`:
+
+* The release stage can use Node 8, while the other stages can use other versions, even use a completely different language.
+* It allows you to easily release a project that is not even written in JavaScript, without impacting the job configuration of your test jobs in any way.
+* Instead of having to figure out a "build leader" to do the release and poll the Travis API to wait until all jobs passed, Travis can handle only running the release step after the other jobs passed.
+* If the release failed, you can see at a glance that only the release stage failed while your test stage succeded.
+* Travis can take care of not running the whole release stage at all on non-master branches and PRs.
+
+### Annotated example `.travis.yml`
+
```
pvdlg: Can you add a section about `package.json` configuration mentioning it's required only for local installs similarly to [travis.md](https://github.com/semantic-release/semantic-release/blob/caribou/docs/recipes/travis.md#packagejson-configuration-for-single-node-job):

```md
### `package.json` configuration

A `package.json` is required only for [local](../usage/installation.md#local-installation) **semantic-release** installation.

{
  "devDependencies": {
    "semantic-release": "^11.0.0"
  },
  "scripts": {
    "semantic-release": "semantic-release"
  }
}
```

docs/recipes/README.md
```diff
@@ -1,9 +1,14 @@
...
-- [Travis CI with build stages](travis-build-stages.md)
-- [GitLab CI](gitlab-ci.md)
+
+* [CircleCI 2.0 workflows](circleci-workflows.md)
+* [Travis CI](travis.md)
+* [Travis CI with build stages](travis-build-stages.md)
+* [GitLab CI](gitlab-ci.md)
 
 ## Package managers and languages
+
```
pvdlg: Can you add these plugins in [plugins-list.md#community-plugins](https://github.com/semantic-release/semantic-release/blob/caribou/docs/extending/plugins-list.md#community-plugins) instead?

docs/recipes/circleci-workflows.md
```diff
@@ -1 +1,75 @@
-# CircleCI 2.0 workflows
+# Using semantic-release with [CircleCI 2.0 workflows](https://circleci.com/docs/2.0/workflows/)
+
+![CircleCI workflows screenshot](https://github.com/circleci/circleci-docs/raw/928cbb042453ae182996ce05f4e4042b02a24634/jekyll/assets/img/docs/workflow_detail.png)
+
+Similar to Travis build stages, CircleCI 2.0 workflows are a great way to run semantic-release in an independent build step.
+We can let CircleCI take care of only running a release on master, after tests passed and running it on an image with Node 8 - your test jobs could use a completely different environment, your project may not even be written in JavaScript.
+
+Make sure to set `NPM_TOKEN` and `GH_TOKEN` in Circle's project settings UI.
```
pvdlg: Can you reference the CircleCi documentation to do that? Similarly to [travis.md#environment-variables](https://github.com/semantic-release/semantic-release/blob/caribou/docs/recipes/travis.md#environment-variables).

docs/recipes/circleci-workflows.md
```diff
@@ -1 +1,75 @@
...
+
+### Annotated example `.circleci/config.yml`
+
+```yml
```
pvdlg: Can this configuration be simplified to keep only what's required by semantic-release, similarly to [travis.md](https://github.com/semantic-release/semantic-release/blob/caribou/docs/recipes/travis.md)?

You can mention it's a minimal configuration for the semantic-release part only and provide a link to the Circle CI doc for users can refer to it for the non semantic-release related parts.

docs/recipes/circleci-workflows.md
```diff
@@ -1 +1,75 @@
...
+      - test_node_6
+      - test_node_8
+      # Release job that runs only on master if the two test jobs succeeded
+      - release:
+          requires:
+            - test_node_6
+            - test_node_8
+          filters:
+            branches:
+              only: master
```
pvdlg: There is no need to filter on branch as it's done by semantic-release. It's also something that is configurable so it shouldn't be hardcoded.

pvdlg:
Thanks for updating the PR!

I added some comments, mostly related to consistency with other docs and to the scope of what semantic-release recipes are (focused on the semantic-release config, not on the CI best practices, advices or opinions).

Regarding the `PHP / Composer` the Travis and semantic-release configuration doesn't seems different than the one in the Build Stage recipe. It seems to be really more about PHP / Composer than specific to semantic-release. I would prefer to write a more generic section in the Build Stage recipe for Non Javascript project similar to [travis.md#non-javascript-projects-configuration](https://github.com/semantic-release/semantic-release/blob/caribou/docs/recipes/travis.md#non-javascript-projects-configuration). This section can re-use the PHP / Composer example the same way [travis.md#non-javascript-projects-configuration](https://github.com/semantic-release/semantic-release/blob/caribou/docs/recipes/travis.md#non-javascript-projects-configuration) uses Go examples.

docs/recipes/README.md
felixfbecker: So this section is reserved for "blessed" plugins? Why hide community plugins away on a different page?

pvdlg: This is a section to list all plugins available, from `@semantic-release` org and community.

There no such thing as "blessed" plugins. There is the plugins maintained by the semantic-release team (for which bug report have to be opened in the semantic-release orga for example) and the community plugins that are not maintained by the semantic-release team.
Anyone can make a PR to add his plugin to the list.

I don't know what you mean by "hide". This page is clearly visible, linked in the README and in the sidemenu of the doc website. This is meant to be a registry of plugins.

docs/recipes/circleci-workflows.md
felixfbecker: That would trigger a `release` build on every branch or PR build, spin up the container, checkout code, npm install, block the build queue, report a GitHub status check for that specific build etc, just to have semantic-release figure out that it shouldn't do anything. Why not include these 3 lines in the example that users will copy+paste?

docs/recipes/circleci-workflows.md
felixfbecker: I am not sure what you would want it to look like - only have a release step? Imo in opposite to documentation, recipes should be examples that can be copy+pasted and then modified if needed. They should examples show how the general documentation can be adapted for specific projects, more as inspiration than a guarantee to work for every project.

docs/recipes/README.md
felixfbecker: So to be clear, you would want the PHP/Composer link to stay here but the VSCE and Docker links to be moved? Imo that is not very user-oriented, because as a user I just want a single place where I can find out how to get semantic-release working with my package manager

docs/recipes/travis-build-stages.md
```diff
@@ -1 +1,63 @@
...
+      # We use script, not after_success, so we get notified if the release failed for unexpected reasons.
+      script:
+        # Don't forget to build before releasing if your project has a build step.
+        # The release stage doesn't share state with the test stage.
+        # Alternatively you could add a prepublishOnly script to package.json
+        - npm run build
+        - npm run semantic-release
+
+# Define when/under what conditions the jobs should be executed
+stages:
```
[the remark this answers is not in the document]
felixfbecker: We seem to disagree on this but imo if you don't use the benefits of build stages (separate stages that can be filtered by Travis) then you might as well not use build stages at all. So if the recipe doesn't show a good example of how build stage features can be utilised, I don't really see a point in having this recipe at all. This is exactly the things that took me hours of tinkering, digging docs and failed attempts to figure out that I would like to share with others so they don't have to go through that (for example, who knew all PR builds have branch set to master?).

docs/recipes/travis-build-stages.md
```diff
@@ -1 +1,63 @@
...
+      language: node_js
+      # Explicitely use only Node 8 for release (no matrix)
+      node_js: '8'
+      # Override script to build and release
+      # We use script, not after_success, so we get notified if the release failed for unexpected reasons.
+      script:
+        # Don't forget to build before releasing if your project has a build step.
+        # The release stage doesn't share state with the test stage.
+        # Alternatively you could add a prepublishOnly script to package.json
+        - npm run build
```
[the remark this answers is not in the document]
felixfbecker: > Can you remove npm run build in order to be consistent with other docs and to keep it focused on what's required to run semantic-release?

Partly repeating what I mentioned in other comments, I see the value of recipes in showing off examples of end-to-end solutions that can often be copy+pasted and would rather have a bit to much than not enough (removing is always easy). Forgetting to run the build script before releasing specifically is one of the biggest pitfalls when migrating to build stages, that has tripped me multiple times (resulting in empty broken releases to users), that I would like others to be aware of.
(Note that I added the `prepublishOnly` script approach too)

> Also can you mention both global and local install?

Will do

docs/recipes/README.md
gr2m: we need a clear distinction between plugins that are supported by the semantic-release team and plugins that are supported by the community, to set the right expectations in terms of compatibility and support. I agree that as a user I’d prefer all plugins to be working with the latest semantic-release and be well maintained and supported, but that will not happen and we have to make which plugins are core and which are not.

Once there are more plugins, I’m sure we can build better ways for discoverability, but for starting out that is the best we can do :)

docs/recipes/circleci-workflows.md
gr2m: I would leave out anything that is not necessary to make semantic-release work. Please leave out all the steps related to caching. I’m not familiar with CircleCI so I don’t know how we could reduce it further, but the simpler it is, the less error prone, the better.

docs/recipes/circleci-workflows.md
gr2m: I agree with @pvdlg, please leave it out, it will work without. I do similar optimizations in my travis.yml files (https://github.com/octokit/rest.js/blob/34109d614958f8e423beb94c6959920e79c14862/.travis.yml#L16) but I would not make it a default trough the CLI or document it that way.

For example, if I set my default branch to something else than `master` and then copy/paste the config which only runs the release on the `master` branch, it will never work. 

It’s great to give people a starting point to make semantic-release working, but it shouldn’t be the goal to put in best practices. People who look into this are likely new to both, semantic-release & circeCI and they will run into problems anyway for which they will then create issues which increases the work load for the maintainers. And maintainers time is the scarces of resources, so when in doubt, I’d always optimize for that

docs/recipes/travis-build-stages.md
```diff
@@ -1 +1,63 @@
...
+
+Build stages are a great way to run semantic release compared to running it in `after_success`:
+
+* The release stage can use Node 8, while the other stages can use other versions, even use a completely different language.
+* It allows you to easily release a project that is not even written in JavaScript, without impacting the job configuration of your test jobs in any way.
+* Instead of having to figure out a "build leader" to do the release and poll the Travis API to wait until all jobs passed, Travis can handle only running the release step after the other jobs passed.
+* If the release failed, you can see at a glance that only the release stage failed while your test stage succeded.
+* Travis can take care of not running the whole release stage at all on non-master branches and PRs.
+
+### Annotated example `.travis.yml`
```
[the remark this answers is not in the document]
gr2m: 👍

docs/recipes/README.md
felixfbecker: How about adding "(community plugin)" after the community plugins in this list?
FWIW gulp highlights almost exclusively community plugins in their recipes.

docs/recipes/travis-build-stages.md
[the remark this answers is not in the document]
gr2m: forgetting a `build` seems like a challenge / pitfall that should be addressed somewhere else, it is not unique to any CI. 

The current setup does not what you think it does

```
script:
  - npm run build
  - npm run lint
  - npm test
```

see https://github.com/semantic-release/semantic-release/issues/609

And the builds for the different stages are separated, you have to build in the same job as semantic-release. 

The right thing to do is to put `npm run build` into `prepublishOnly`

docs/recipes/travis-build-stages.md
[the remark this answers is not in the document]
gr2m: How about we remove it for our docs, but then link to a list of more complex configurations in other projects? That would have the benefit that these would be configurations that are actually in use and hence being updated. We could add links to a list of different configurations for projects of different nature, only ask that the respective projects throughly comment their settings?

gr2m:
Hey @felixfbecker, okay if we start a new PR to document configuration for build stages and Circle CI and then we take it from there?

I think that will make it easier to move things forward after the discussion stalled here.

I suggest we create the documentation for the parts we agree on and then do follow up issues for the parts that can be improved.

pvdlg:
Per [this comment](https://github.com/semantic-release/semantic-release/pull/573#issuecomment-363875766) I opened #659.

@felixfbecker you have a lot of knowledge regarding CI workflows and that can be really beneficial for the overall developer community. We think the best way to share it is outside the semantic-release documentation though. 

We try to keep the semantic-release official documentation focused on semantic-release only and stay away from general recommendations, best practices and opinionated configs.

CI workflows and semantic-release are intrinsically related but we don't want to be responsible for hosting and maintaining recommendations or best practices and becoming a forum about pros/cons of specifics CI config/practices.

If you'd like to provide advices, best practices, opinionated config and more specific scenario,  feel free to create blog posts or articles and we'll happily add them to [resources](https://github.com/semantic-release/semantic-release/blob/caribou/docs/resources.md) and promote them on Twitter or other channels.

PR #576, opened 2017-12-27
Move npm workaround for missing gitHead to the npm plugin

pvdlg:
Follow up to semantic-release/npm#26 and semantic-release/npm#29.

Move the workaround for npm/read-package-json#77 implemented in the core when npm publishing was still part of it, to the npm plugin.


lib/get-commits.js
```diff
@@ -44,22 +44,28 @@ const getVersionHead = require('./get-version-head');
...
-  let gitTag;
-  if (gitHead || version) {
-    try {
-      ({gitHead, gitTag} = await getVersionHead(gitHead, version, branch));
-    } catch (err) {
-      if (err.code === 'ENOTINHISTORY') {
-        logger.error(notInHistoryMessage(err.gitHead, branch, version));
-      } else {
-        logger.error(noGitHeadMessage(branch, version));
+  if (gitHead) {
```
gr2m: Maybe add a comment referencing https://github.com/npm/read-package-json/pull/77 for explanation?

pvdlg: It's not related to npm/read-package-json#77. The workaround for npm/read-package-json#77  has been moved to the npm plugin.

This is related to the `ENOTINHISTORY` error. That allow users to create a tag for a given commit in case they have rewritten their master history and the last release commit is missing.

felixfbecker:
I think this was a breaking change. The VS Code marketplace (and I assume lots of other registries) does not provide a `gitHead` field. This was working fine before but now it breaks with this error:
```
[Semantic release]: An error occurred while running semantic-release: Error: The "getLastRelease" plugin output if defined, must be an object with a valid semver version in the "version" property and the corresponding git reference in "gitHead" property. Received: { version: '1.12.1' }
    at Object.getLastRelease (/home/travis/build/felixfbecker/vscode-php-debug/node_modules/semantic-release/lib/plugins/normalize.js:44:13)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
```

pvdlg:
The npm plugin was never intended to be used to retrieve the last release on a package manager other than the npm registry.
You can use the git plugin to retrieve the last release based on git tag.

felixfbecker:
Not talking about the npm plugin, but the vsce plugin: https://github.com/raix/semantic-release-vsce

pvdlg:
The workaround changed by this PR is related to the npm plugin using the `gitHead` from the package metadata that is sometimes not present. If you want to implement a `getLastRelease` plugin you must return a version and a `gitHead`. If the package manager you are using doesn't provide a `gitHead` then you should use the git plugin.

Relying on the workaround removed in this PR is effectively the same thing as using the git plugin: it uses the git tags to retrieve the last version.

felixfbecker:
Well as I understand it with this workaround the last release version would be determined from the marketplace, and then only the matching tag would be determined from git. Whereas the git plugin would do it all through git (not saying that is a big problem for me).

But it's still a breaking change, right?

pvdlg:
No, in the sense that is what not intended not documented to work that way. Plugins are supposed to handle the specificities of the system they integrate with (like having a `gitHead` or not or other ways to retrieve the commit associated with the last release).

felixfbecker:
@pvdlg what is the proper thing to do for plugins that work with registries that can get you the latest version, but not the gitHead? Should they take a dependency on `@semantic-release/git` and call that under the hood to resolve the commit ID that matches the version?
I am assuming it is beneficial to resolve the version from the registry, otherwise we could just always use `@semantic-release/git` for all registries including npm. Which means we would have to split the process into two steps and in the best case not have to copy the code for resolving the commit ID.
I would guess the vast majority of registries do not expose a `gitHead` field so this is an important issue to solve and communicate.

felixfbecker:
https://github.com/felixfbecker/semantic-release-docker is also affected by this (I assume almost every plugin except npm)

pvdlg:
If your registry doesn't have a gitHead you have to use git tag.
The proper way to do that is to use the git plugin for the `getLastRelease` step and to make sure you have at least one publish plugin that creates the tag (with the git or github plugin for example).

No need for your plugin to use the git plugin as a dependency. Just use it for the `getLastRelease` step.

felixfbecker:
Let me ask the question then what the benefit is at all to get the last release from the registry? Why not always use the git plugin?

pvdlg:
As most registry do not provide a way to get the commit sha associated with the last release, using a git tag is necessary. Git tags are a quite convenient to get the last release sha especially when you have multiple branches and distribution channels as the tag follow the commit (contrary to the `gitHead` in the registry that is not affected by branch merging on the git side).

So I don see much benefits to use anything else than the git tags to get the last release. The only constraint is the tag has to be created during the release process, but I don't see any scenario in which you don't want to create a tag.

This is why I propose [here](https://github.com/semantic-release/evolution/blob/release-workflows/_drafts/release-workflows.md#deprecate-getlastrelease-and-move-tag-management-to-the-core) to always use create the git tag and to use it to retrieve the last release.

felixfbecker:
The only benefit I could see is have multiple `getLastRelease` map to the `publish` plugins, so that if for example the npm publish succeeded, but the github publish failed, the github publish will be retried because its `getLastRelease` is lagging behind. That would make the system a bit more error tolerant. But that is not what is currently implemented anyway and would also add complexity.

Wouldn't you still want the tag creation to live in the git plugin though? What if I want to make semantic-release work with Mercurial for example?

pvdlg:
> The only benefit I could see is have multiple getLastRelease map to the publish plugins, so that if for example the npm publish succeeded, but the github publish failed, the github publish will be retried because its getLastRelease is lagging behind. That would make the system a bit more error tolerant. But that is not what is currently implemented anyway and would also add complexity.

I don't think having multiple `getLastRelease` would solve that. The `getLastRelease` is called once before any `publish`. The code already allow any `publish` task to make a commit, as any subsequent `publish` will be called with the new `gitHead` (corresponding to the last commit made on the branch).

Any retry mechanism can be implemented at the plugin level. There is actually retry mechanism already implemented an most API client. In this case it's not really something that we can identify a situation that require a retry. Basically if it's a network issue, the API client will take care of retrying, but getting a 404 indicate a functional error (like the commit doesn't exist on the remote).

> Wouldn't you still want the tag creation to live in the git plugin though?

As proposed [here](https://github.com/semantic-release/evolution/blob/release-workflows/_drafts/release-workflows.md#deprecate-getlastrelease-and-move-tag-management-to-the-core) it would be moved to the core.

> What if I want to make semantic-release work with Mercurial for example?

That would be quite complex because as of now, semantic-release is really tied to git. There is a lot code that runs the `git` command. The solution probably to implement all that in the core with a mechanism to detect if we are on a git or mercurial repo, and have all the function in [git.js](https://github.com/semantic-release/semantic-release/blob/caribou/lib/git.js) ported to Mercurial.

> Or use tag prefixes in a multi-package repo?

The multi-package is still in discussion. The solution will probably to add a new option to the core that allow to define the git tag format. So you could define like `<package>@v<version>` for monorepo or `v<version>` for regular repo, or pretty much any format that you like.
Moving the  `getLAstRelease` step and the tag creation to the core would make that a lot easier and all the tag stuff will be handled in the same place. It would also simplify plugins and let them focus on the specific platform they integrate with (no need to deal with git stuff in plugins).

Issue #563, opened 2017-12-15
Support multi-branch, multi-channel, LTS and pre-release workflows


pvdlg:
This a proposal to support more complex release workflows in a fully automated way, including:
- LTS branches and distribution channels
- Future branches and distribution channels
- Pre-releases

Please see https://github.com/semantic-release/evolution/pull/1 for the proposed list of workflows to be supported by `semantic-release`.

**`semantic-release` modifications**
- [ ] Add the new `branches` options
- [ ] Normalize each branch configuration based on [naming convention](https://gist.github.com/pvdlg/9fa33dbc6792edc2cf3d0765c5d8b8b3#branches-naming-convention)
- [ ] [Validate](https://gist.github.com/pvdlg/9fa33dbc6792edc2cf3d0765c5d8b8b3#configuration-validation) the `branches` configuration
- [ ] Re-purpose the `branch` options as the current branch on which `semantic-release` is running
- [ ] Verify that only allowed release are made
  - [ ] On LTS branches verify it's in `range`
  - [ ] In default and future branches make sure that releasing will comply with the `type` of the following non pre-release future branch
- [ ] When no version has to be released, check if we are in a merge situation (the commit associated with the last release is present in following branch  and it is not a pre-release branch)
  - [ ] If yes, make a release in order for the publish plugins to update the channel
- [ ] Determine if the branch is a pre-release one, and if it is increment the version accordingly
- [ ] Pass the `channel` on which to release `nextRelease` to the publish plugins
- [ ] Move the git tag creation to the core

**`publish` plugins modifications**
- [ ] If release already exists, make it available on `nextRelease.channel` (in case of github, remove the `prerelease` flag if the target channel is the default one)
- [ ] For the `npm` plugin, always publish on the dist-tag passed in `nextRelease.channel` or `latest` if `nextRelease.channel` is not set
- [ ] For the `github` plugin, set `prerelease` flag if the current branch is a future or prerelease
- [ ] Remove the tag creation from the `publish` plugins

**Deprecate `getLastRelease` plugins**
- [ ] Move the `getLastRelease` logic of the git plugin into the core

This is fairly broad and complex subject, so I'd love to have feedback from the community, in particular:
- Is there a type of workflow that wouldn't be supported with this new feature?
- Is there some potential inconsistencies that are not covered? (i.e. would it be possible to create a release that shouldn't happen, or create an incoherent situation).
[...]

PR #578, opened 2017-12-29
Make semantic-release CI agnostic

### LINKED ISSUE #569: Fail builds on error
If no NPM token is set and semantic-release is run as `script` in Travis, the build still passes and semantic-release exits with 0.

```
1.81s$ npm run semantic-release


> iterare@1.0.0 semantic-release /home/travis/build/felixfbecker/iterare

> semantic-release


[Semantic release]: Load plugin verifyConditions from @semantic-release/npm

[Semantic release]: Load plugin verifyConditions from @semantic-release/github

[Semantic release]: Load plugin verifyConditions from @semantic-release/condition-travis

[Semantic release]: Load plugin getLastRelease from @semantic-release/npm

[Semantic release]: Load plugin analyzeCommits from @semantic-release/commit-analyzer

[Semantic release]: Load plugin generateNotes from @semantic-release/release-notes-generator

[Semantic release]: Load plugin publish from @semantic-release/npm

[Semantic release]: Load plugin publish from @semantic-release/github

[Semantic release]: Run automated release from branch master

[Semantic release]: Call plugin verify-conditions

[Semantic release]: Verify authentication for registry https://registry.npmjs.org/

[Semantic release]: ENONPMTOKEN No npm token specified.



The command "npm run semantic-release" exited with 0.
```

pvdlg:
This PR remove the need for a CI `verifyConditions` plugin by:
- Moving the verification of the branch and PR in the core. Works with any major (and not so major) CI services. See [env-ci](https://github.com/pvdlg/env-ci) for the list.
- Leaving the responsibility to run `semantic-release` only on one job to the CI and/or the user.

**Benefits**
- Allow to support any CI out of the box
- Avoid forcing users who do not use Travis to override a big chunk of configuration (the `verifyConditions` plugins)
- Allow users to do a dry-run locally to verify the main and plugins configuration
- Allow to use shareable config more easily (Currently any shareable config user who do not use Travis would have to override the `verifyConditions` configuration defeating the purpose of shareable config)
- Allow to handle errors properly, failing the build in case of error but not if running on a job that is not the build leader. See [comment](https://github.com/semantic-release/semantic-release/issues/569#issuecomment-353502532)
- Detecting the current branch from the core is required in order to implements #563

Checking that `semantic-release` runs only on one job is extremely dependent on the CI as well as the user workflow and personal preferences. Supporting any CI and workflow out there would be a herculean task, and mostly redundant as most CI provider are now starting to provide solutions to do that. In addition, imposing Travis in the default settings creates unnecessary frictions, and make it harder for new users to adopt `semantic-release`.

The only disadvantage is that current Travis users who use the default config will have to slightly change their `.travis.yml`, replacing `npm run semantic-release` by `npm install -g travis-deploy-once && travis-deploy-once "npm run semantic-release"`.

See semantic-release/travis-deploy-once#38.

We can update `semantic-release-cli` do do this setup automatically.

**List of changes**
- Remove `@semantic-release/condition-travis` from the default plugins
- Verify the current branch in the core
- Verify the build is not triggered by a PR in the core
- Run in dry-run mode if not triggered on CI
- Dry-run mode runs the `verifyConditions` plugins, allowing to detect configuration error locally
- Return without error when no version has to be released due to no changes
- Return without error if the build is triggered from a PR
- Return without error if the current branch is not the configured branch
- CLI return with exit code 1 if there is a `semanticReleaseError`, allowing to fail builds in case of config error, missing token etc...

**Breaking change**
`semantic-release` doesn't make sure it runs only on one Travis job anymore.
The CI configuration has to be done such that `semantic-release`
- runs only once per build
- runs only after all tests are successful on every jobs of the build
- runs on Node >=8

This can easily be done with [travis-deploy-once](https://github.com/semantic-release/travis-deploy-once).

**Migration Guide**

Modify your `.travis.yml` to use `travis-deploy-once`.
Replace:
```yaml
after_success:
  - npm run semantic-release
```
by:
Replace
```yaml
after_success:
  - npm install -g travis-deploy-once
  - travis-deploy-once "npm run semantic-release"
```

Fix semantic-release/condition-travis#96, Fix #569

TODO:
- [x] Release semantic-release/travis-deploy-once#38
- [x] Update `.travis.yml` to install `travis-deploy-once` directly rather than from `semantic-release/travis-deploy-once#cli` branch
- [ ] Release semantic-release/cli#165

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,6 @@
 const marked = require('marked');
 const TerminalRenderer = require('marked-terminal');
-const SemanticReleaseError = require('@semantic-release/error');
+const envCi = require('env-ci');
 const getConfig = require('./lib/get-config');
 const getNextVersion = require('./lib/get-next-version');
 const getCommits = require('./lib/get-commits');
@@ -8,19 +8,39 @@ const logger = require('./lib/logger');
 const {gitHead: getGitHead, isGitRepo} = require('./lib/git');
 
 module.exports = async opts => {
+  const {isCi, branch, isPr} = envCi();
+
+  if (!isCi && !opts.dryRun) {
+    logger.log('This run was not triggered in a known CI environment, running in dry-run mode.');
+    opts.dryRun = true;
+  }
+
+  if (isCi && isPr) {
+    logger.log('This run was triggered by a pull request and therefore a new version won’t be published.');
+    return;
+  }
+
   if (!await isGitRepo()) {
-    throw new SemanticReleaseError('Semantic-release must run from a git repository', 'ENOGITREPO');
+    logger.error('Semantic-release must run from a git repository.');
+    return;
   }
 
   const config = await getConfig(opts, logger);
   const {plugins, options} = config;
 
+  if (branch !== options.branch) {
+    logger.log(
+      `This test run was triggered on the branch ${branch}, while semantic-release is configured to only publish from ${
+        options.branch
+      }, therefore a new version won’t be published.`
+    );
+    return;
+  }
+
   logger.log('Run automated release from branch %s', options.branch);
 
-  if (!options.dryRun) {
-    logger.log('Call plugin %s', 'verify-conditions');
-    await plugins.verifyConditions({options, logger});
-  }
+  logger.log('Call plugin %s', 'verify-conditions');
+  await plugins.verifyConditions({options, logger});
 
   logger.log('Call plugin %s', 'get-last-release');
   const {commits, lastRelease} = await getCommits(
@@ -32,7 +52,8 @@ module.exports = async opts => {
   logger.log('Call plugin %s', 'analyze-commits');
   const type = await plugins.analyzeCommits({options, logger, lastRelease, commits});
   if (!type) {
-    throw new SemanticReleaseError('There are no relevant changes, so no new version is released.', 'ENOCHANGE');
+    logger.log('There are no relevant changes, so no new version is released.');
+    return;
   }
   const version = getNextVersion(type, lastRelease, logger);
   const nextRelease = {type, version, gitHead: await getGitHead(), gitTag: `v${version}`};
@@ -67,4 +88,5 @@ module.exports = async opts => {
     });
     logger.log('Published release: %s', nextRelease.version);
   }
+  return true;
 };
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,4 +25,5 @@ script:
 
 after_success:
   - npm run codecov
-  - npm run semantic-release
+  - npm install -g semantic-release/travis-deploy-once@4
+  - travis-deploy-once "npm run semantic-release"
--- a/README.md
+++ b/README.md
@@ -170,7 +170,7 @@ semantic-release
 These options are currently available:
 - `branch`: The branch on which releases should happen. Default: `'master'`
 - `repositoryUrl`: The git repository URL. Default: `repository` property in `package.json` or git origin url. Any valid git url format is supported (See [Git protocols](https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols)). If the [Github plugin](https://github.com/semantic-release/github) is used the URL must be a valid Github URL that include the `owner`, the `repository` name and the `host`. The Github shorthand URL is not supported.
-- `dry-run`: Dry-run mode, skipping verifyConditions, publishing and release, printing next version and release notes
+- `dry-run`: Dry-run mode, skip publishing, print next version and release notes
 - `extends`: Array of module or files path containing a shareable configuration. Options defined via CLI or in the `release` property will take precedence over the one defined in a shareable configuration.
 - `debug`: Output debugging information
 
--- a/cli.js
+++ b/cli.js
@@ -48,12 +48,10 @@ module.exports = async () => {
       await require('.')(pickBy(program.opts(), value => !isUndefined(value)));
     }
   } catch (err) {
-    // If error is a SemanticReleaseError then it's an expected exception case (no release to be done, running on a PR etc..) and the cli will return with 0
-    // Otherwise it's an unexpected error (configuration issue, code issue, plugin issue etc...) and the cli will return 1
+    process.exitCode = 1;
     if (err.semanticRelease) {
       logger.log(`%s ${err.message}`, err.code);
     } else {
-      process.exitCode = 1;
       logger.error('An error occurred while running semantic-release: %O', err);
     }
   }
--- a/lib/plugins/definitions.js
+++ b/lib/plugins/definitions.js
@@ -6,7 +6,7 @@ const validatePluginConfig = conf => isString(conf) || isString(conf.path) || is
 
 module.exports = {
   verifyConditions: {
-    default: ['@semantic-release/npm', '@semantic-release/github', '@semantic-release/condition-travis'],
+    default: ['@semantic-release/npm', '@semantic-release/github'],
     config: {
       validator: conf => !conf || (isArray(conf) ? conf : [conf]).every(conf => validatePluginConfig(conf)),
       message:
--- a/package.json
+++ b/package.json
@@ -16,7 +16,6 @@
   },
   "dependencies": {
     "@semantic-release/commit-analyzer": "^5.0.0",
-    "@semantic-release/condition-travis": "^7.0.0",
     "@semantic-release/error": "^2.1.0",
     "@semantic-release/github": "^3.0.0",
     "@semantic-release/npm": "^2.0.0",
@@ -25,6 +24,7 @@
     "commander": "^2.11.0",
     "cosmiconfig": "^3.1.0",
     "debug": "^3.1.0",
+    "env-ci": "^1.0.0",
     "execa": "^0.8.0",
     "get-stream": "^3.0.0",
     "git-log-parser": "^1.2.0",
--- a/test/index.test.js
+++ b/test/index.test.js
@@ -19,7 +19,6 @@ test.beforeEach(t => {
   t.context.log = stub();
   t.context.error = stub();
   t.context.logger = {log: t.context.log, error: t.context.error};
-  t.context.semanticRelease = proxyquire('../index', {'./lib/logger': t.context.logger});
 });
 
 test.afterEach.always(() => {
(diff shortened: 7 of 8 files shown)


gr2m:
```
after_success:
  - npm install -g travis-deploy-once
  - travis-deploy-once "npm run semantic-release"
```

I like the separation of `semantic-release` and `travis-deploy-once` 👍 

Installing `travis-deploy-once` in travis instead of as a devDependency might be problematic though in case we ever release a breaking change. Maybe change it it to

```
npm install -g travis-deploy-once@3
```

And when we release a v4, we can log a deprecation warning for the latest v3

pvdlg:
Good idea. Maybe I can add https://github.com/yeoman/update-notifier to `travis-deploy-once` instead?

gr2m:
> Moving the verification of the branch and PR in the core

I wonder if maybe instead of moving all the logic into core, we could replace `@semantic-release/condition-travis` with a generic `@semantic-release/ci`?

pvdlg:
> I wonder if maybe instead of moving all the logic into core, we could replace @semantic-release/condition-travis with a generic @semantic-release/ci?

Detecting the current `branch` will be required for #563 as we need to know on which branch we are to determine what to do (do a release, error out or move a release to a different dist-tag).

Regarding the PR detection it doesn't *have to be* in the core for now, but putting it there has the advantage of allowing in the future to do specific things for PRs like log/inform the user of the version that would be released.
In addition plugins are better suited for optional things users can enable/disable and releasing from a PR (by disabling `@semantic-release/ci`) is not really something that would make sense.

index.js
```diff
@@ -1,26 +1,46 @@
...
 const getNextVersion = require('./lib/get-next-version');
 const getCommits = require('./lib/get-commits');
 const logger = require('./lib/logger');
 const {gitHead: getGitHead, isGitRepo} = require('./lib/git');
 
 module.exports = async opts => {
+  const {isCi, branch, isPr} = envCi();
+
+  if (!isCi) {
+    logger.log('This run was not triggered in a CI environment, running in dry-run mode.');
```
gr2m: To be more precise, we could say "This run was not triggered in a _known_ CI environment (https://git.io/vbA4g). Running in dry-run mode."

gr2m: and can `opts.dryRun` be already set to true at this point? In that case the condition should maybe be

```js
if (!isCi && !opts.dryRun) {
```

so folks don’t see that message when running with the `dryRun` option explicitly

index.js
```diff
@@ -1,26 +1,46 @@
...
+    return;
+  }
+
   logger.log('Run automated release from branch %s', options.branch);
 
-  if (!options.dryRun) {
-    logger.log('Call plugin %s', 'verify-conditions');
-    await plugins.verifyConditions({options, logger});
-  }
+  logger.log('Call plugin %s', 'verify-conditions');
```
gr2m: why the `%s` placeholder here?

gr2m:
> plugins are better suited for optional things users can enable/disable and releasing from a PR (by disabling `@semantic-release/ci`) is not really something that would make sense. 

as long as we can assume that everyone wants to release from a CI, that works. But for example @bkeepers tried to do a release from his local machine the other day, as the automated release process is not a good setup for @probot. Would that be possible with the built-in CI detection? I agree that this is out of scope of semantic-release, automation is at its core, I just want to make sure we don’t block legit use cases where people want to run semantic-release on custom environments which would not be detected by [env-ci](https://github.com/pvdlg/env-ci)

index.js
pvdlg: I made both changes

index.js
pvdlg: It's to highlight `verify-conditions` with color. See [logger.js#L11](https://github.com/semantic-release/semantic-release/blob/caribou/lib/logger.js#L11).

In retrospective it's maybe unnecessary fancyness. But it's not related to this PR.

pvdlg:
Currently releasing from a local machine is not really supported. There might be a way to achieve currently but that would be very hacky.

I think we should support local release in a better way (even if we mention it's not the recommended way).

We could do something like that when running outside a CI:
```bash
$ semantic-release
=> Run a dry-run

$ semantic-release --no-dry-run or $ semantic-release --local
=> Do a release, ignoring the CI/PR checks
```
Or something even better:
```bash
$ semantic-release
? This run was not triggered in a known CI environment. Do you want to run a dry-run or make a release?
- dry-run (default)
- release
? We will release version x.y.z
- Ok
- Abort
? The release note will be .....
- Ok
- Abort
=> Do the release
```

pvdlg:
> let me know what you think about my other comments

Which comment are you referring to?

gr2m:
> Which comment are you referring to?

nevermind got lost in PR reviews :)

gr2m:
re: local release. I think it’s too much effort for a use case we don’t really want to support.

In Brandon’s case, I see ways to allow for semi-automated releases, but it will require a GitHub app. Can’t wait to finally get going on this :)

felixfbecker:
I really like the idea of this PR!

Only point I have is to not recommend imperatively installing dependencies in CI, because that prevents you from using tools like Greenkeeper and possibly many other tools/IDEs/... that read package.json. There really is no reason to not do it declaratively with a package.json.

pvdlg:
Good point.

What if we recommend the following instead?

```json
{
  "devDependencies": {
    "semantic-release": "^12.0.0",
    "travis-deploy-once": "^4.0.0"
  },
  "scripts": {
    "semantic-release": "semantic-release",
    "travis-deploy-once": "travis-deploy-once"
  }
}
```

```yaml
after_success:
  - npm run travis-deploy-once "npm run semantic-release"
```

felixfbecker:
I think in Travis you don't even need to define the script because Travis adds ./node_modules/.bin to the PATH.

pvdlg:
Do you mean that the following should work?

```json
{
  "devDependencies": {
    "semantic-release": "^12.0.0",
    "travis-deploy-once": "^4.0.0"
  },
  "scripts": {
    "semantic-release": "semantic-release"
  }
}
```

```yaml
after_success:
  - travis-deploy-once "npm run semantic-release"
```

Even if the case, some users might use this doc and adapt it to another CI that doesn't add `./node_modules/.bin` to the path. I think it's not the standard behavior, [npx] was created among other thing to do that: [npx#running-a-project-local-bin](https://github.com/zkat/npx#running-a-project-local-bin).

felixfbecker:
Yeah I think so.
This guide is specifically targetting Travis, so I don't think adaption for another CI is a concern here.
Anyway I just think that no deps should be installed imperatively in CI. I think it would be just as fine to just tell the user to add it to their devDeps with `npm install --save-dev travis-deploy-once` locally.

pvdlg:
> I think it would be just as fine to just tell the user to add it to their devDeps with npm install --save-dev travis-deploy-once locally.

Yes but we still have to mention how to run it: either `npm run travis-deploy-once "npm run semantic-release"` or `travis-deploy-once "npm run semantic-release"`.

I think I'll go with the first option (`npm run travis-deploy-once "npm run semantic-release"` + `"travis-deploy-once": "travis-deploy-once"`) as `./node_modules/.bin` being in the `PATH` is not documented in the Travis doc, therefore it might not be guaranteed to work in the future.

gr2m:
> - travis-deploy-once "npm run semantic-release"

This will not work by itself. You’ll need to do `./node_modules/.bin/travis-deploy-once "npm run semantic-release"`, but that’s not OS agnostic. I’d stick with `npm run travis-deploy-once` or do `npm run -s travis-deploy-once`

> npm run travis-deploy-once "npm run semantic-release"

I think you need to separate arguments you want to pass to the `travis-deploy-once` binary using `--`. So the full script would be 

> npm run travis-deploy-once -- "npm run semantic-release"

I wonder if we should just go ahead and create stages from the beginning instead of installing the additional `travis-deploy-once` dependency?

pvdlg:
> I think you need to separate arguments you want to pass to the travis-deploy-once binary using --. So the full script would be `npm run travis-deploy-once -- "npm run semantic-release"`

I though so as well but after testing it works without the `--` (at least on my local machine and on Travis).

> I wonder if we should just go ahead and create stages from the beginning instead of installing the additional travis-deploy-once dependency?

Agreed, but maybe in a later step?
- That would make the migration more complex
- Build Stages are still in beta
- I don't think Build Stages are supported on Travis pro and Enterprise

gr2m:
> Agreed, but maybe in a later step?

👍

felixfbecker:
> You’ll need to do ./node_modules/.bin/travis-deploy-once "npm run semantic-release", but that’s not OS agnostic

I don't see how that is not OS agnostic, especially in the context of Travis?

> I wonder if we should just go ahead and create stages from the beginning instead of installing the additional travis-deploy-once dependency?

My opinion on this is that build stages are the future, you can see every CI provider moving in that direction. And it's a great way to plug-in a release step that has its own environmental requirements (Node 8 etc) without touching the rest of the config at all.

Polling the Travis API and keeping one arbitrary job alive indefinitely always felt hacky to me.

gr2m:
good point, as long as it’s not run in Windows, `./node_modules/.bin/travis-deploy-once` should work just fine.

The polling is a hack but it’s fairly robust by now. Having to different solutions for travis-ci.org/com and Enterprise is reason enough to not go with Stages yet. We can switch once stages land in Travis Enterprise

pvdlg:
I agree that Build stages are the future and offer a lot of advantages and taht `travis-deploy-once` is somewhat "hacky". But at that point I think we've really made a lot of effort to remove dependencies to `travis-deploy-once`, give users more flexibility in their CI config and  promote Build Stages.

With version `12.0.0` you can use Build Stages freely, without any constraints, without `travis-deploy-once`, without any environment variable and build stages can be promoted via docs like in #573.

But I don't see any reasons to go on a crusade against users who do not use Build Stages yet as:
- Build stages are still in beta and can change anytime
- Build stages are not supported on Travis Pro and Enterprise (AFAIK)
- A large part of the semantic-release user do not use Build Stage and is not familiar with it yet
- Offering the possibility to use `travis-deploy-once` doesn't prevent to recommend Build Stages and to use them

Fo all those reasons, I would favor a smooth, step by step migration toward Build Stages, that doesn't break any current workflow in the short to mid-term:
- Use `travis-deploy-once` locally installed with `npm run` in the `12.0.0` migration guide in the CLI  for now
- Update docs with both `travis-deploy-once` and Build Stages, recommending Build stages, See #583
- Release `seamntic-release@12.0.0` on the `@latest` dist-tag in a few weeks and see how the migration goes for users
- In a few weeks, update the CLI to use Build Stages if the users select `Travis CI` option and `travis-deploy-once` if they select `Travis Pro` or `Travis Enterprise` option
- Remove `travis-deploy-once` from docs and CLI once it is out of beta and available on Travis Pro and Enterprise

PR #586, opened 2018-01-01
feat: allow to release from local machine

### LINKED ISSUE #584: Extract agnostic CI verification as a default `condition-ci` plugin
The following additions in `v12.0.0` seems awfully restrictive, especially since it used to be optional (by not including `condition-travis`). 

```js
  if (!isCi && !opts.dryRun) {
    logger.log('This run was not triggered in a known CI environment, running in dry-run mode.');
    opts.dryRun = true;
  }
```
Forcing `dry-run` unless the CI env vars are set makes it more cumbersome to develop plugins.

```
  if (isCi && isPr) {
    logger.log('This run was triggered by a pull request and therefore a new version won’t be published.');
    return;
  }
```
Forcing the PR env variable makes it impossible to upgrade [`semantic-release-github-pr`](https://github.com/Updater/semantic-release-github-pr) to `v12.0.0` since it needs to run on PRs. 

What do you think of extracting it as a CI-agnostic `condition-ci` plugin?

pvdlg:
Doing releases outside a CI environment is not the recommended approach, but it might be useful in specific circumstances (for example #584 and [comment](https://github.com/semantic-release/semantic-release/pull/578#issuecomment-354579201)).

Since #578, `condition-travis` is not a default plugin anymore and verifying that `semantic-release` runs after all other CI jobs are successful is done externally, therefore allowing local releases is just a couple line change. So why not.

In addition the `Can I run this on my own machine rather than on a CI server?` FAQ mention it is possible even though it's not since several releases, so it was probably possible at some point. No real reason to remove (or more exactly continue to remove) a feature that was present before.

The FAQ clearly mention that even if it's possible, it's not the recommended approach.

Fix #584 and [comment](https://github.com/semantic-release/semantic-release/pull/578#issuecomment-354579201).

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,13 +10,13 @@ const {gitHead: getGitHead, isGitRepo} = require('./lib/git');
 module.exports = async opts => {
   const {isCi, branch, isPr} = envCi();
 
-  if (!isCi && !opts.dryRun) {
+  if (!isCi && !opts.dryRun && !opts.noCi) {
     logger.log('This run was not triggered in a known CI environment, running in dry-run mode.');
     opts.dryRun = true;
   }
 
-  if (isCi && isPr) {
-    logger.log('This run was triggered by a pull request and therefore a new version won’t be published.');
+  if (isCi && isPr && !opts.noCi) {
+    logger.log("This run was triggered by a pull request and therefore a new version won't be published.");
     return;
   }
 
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -10,13 +10,13 @@ const {gitHead: getGitHead, isGitRepo} = require('./lib/git');
 module.exports = async opts => {
   const {isCi, branch, isPr} = envCi();
 
-  if (!isCi && !opts.dryRun) {
+  if (!isCi && !opts.dryRun && opts.noCi !== true) {
```
gr2m: why ` !== true` instead of `!opts.noCi`?

pvdlg: No reason, it was a mistake. I changed it.

PR #591, opened 2018-01-02
feat: allow to exclude commits from analysis

### LINKED ISSUE #41: add `[release skip]` commit message command

pvdlg:
Exclude commits from the analysis if they contains `[skip release]` or `[release skip]` in the subject or the body.

Fix #41


pvdlg:
I added a FAQ with explanation about `[skip release]`/`[release skip]`

pvdlg:
Oops I referenced the wrong PR...It should not be closed

PR #592, opened 2018-01-04
fix(security): Updated marked dependency

simlu:
The marked dependency has a known moderate severity security vulnerability in version range < 0.3.9 and should be updated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "get-stream": "^3.0.0",
     "git-log-parser": "^1.2.0",
     "lodash": "^4.0.0",
-    "marked": "^0.3.6",
+    "marked": "^0.3.9",
     "marked-terminal": "^2.0.0",
     "p-reduce": "^1.0.0",
     "read-pkg-up": "^3.0.0",


pvdlg:
For information, due to the `^` in the version definition, the latest version in the `3.x.x` range will always be used.

Is it about a XSS vulnerability? Doesn't it affect only web browsers? We use marked to log colorized marked down in the console.

simlu:
@pvdlg This is no longer true with package-log.json in later node version - whatever is written in the package-log.json is being installed and the file only changes when the dependencies change.

In that case this vulnerability might not concern you. However it still came up in my project - so better to fix it and prevent other people running into it.

travi:
this has impacted a number of my projects as well due to lockfiles pinning an older version. since there is a vulnerability, i think that it does make sense to move the minimum of the range defined by semantic-release to a version beyond the fix of the vulnerability. even though it probably isnt realistically impactful based on the usage, this would satisfy warnings for a lot of users with minimal impact to semantic-release

pvdlg:
> @pvdlg This is no longer true with package-log.json in later node version - whatever is written in the package-log.json is being installed and the file only changes when the dependencies change.

Per the [npm documentation](https://docs.npmjs.com/files/package-lock.json):

> One key detail about package-lock.json is that it cannot be published, and it will be ignored if found in any place other than the toplevel package.

The `package-lock.json` file will be used only in development when you `npm install` at the root of your project. When users of your project install it the `package-lock.json` will not be used as it's not published in your package.

Even if we merge this PR and release and updated version of `semantic-release` it will not be used by your project in dev, due to the `package-lock.json`. However the last version of `marked` will currently be used by the users of your package when they install it as a dependency, without having to change anything in `semantic-release`.

> In that case this vulnerability might not concern you. However it still came up in my project - so better to fix it and prevent other people running into it.

Updating the dependency version in `semantic-release` will not change that. You'll have to update your `package-lock.json` no matter what, either to update the version of `semantic-release` and `marked` if this PR is merged, or to update only `marked` otherwise.
In your project, just regenerate your `package-lock.json` so you will get all the last fixes of all your dependency. Also, you might have security vulnerabilities other than `marked` for which there is a fix, but you don't get them in development because the `package-lock.json` makes you use outdated versions.

I know it's a simple modification with no impact and I could just merge it and move on. But this change  doesn't solve the actual problem, and I don't want to give folks the false idea that it does. The actual problem is that using a `package-lock.json` make you have outdated dependencies in development, and the only solution is to not use a `package-lock.json` or to update it frequently.

We had several PRs of this type recently and they come from a misunderstanding about how SemVer and `package-lock.json` works. Mostly because it's frankly really confusing and also because `pacakge-lock.json` is the default behavior, so folks just use it without having to read the doc in details. Basically as a `npm` or `yarn` user, you will never get a warning telling you "Using a lockfile, which is the default behavior, will prevent you to get any fixes while in development, although those fixes will automatically be used in production".

As SemVer is at the core of `semantic-release` it's important for us to explain how it works and to avoid misconceptions such as the one around `package-lock.json`

travi:
It is correct that the lockfile of the semantic-release project would have no impact on consumers, so I don't think that is what is being proposed here. 

As you mentioned, the lockfile of the consuming projects is the issue, to an extent. npm only updates a local dependency from a lower in-range version to a higher in-range version in a clean install, even without a lockfile because it already has a version that satisfies the range.

With the semver range defined by this package, the version with the vulnerability is an acceptable version that satisfies the defined range. All we are proposing is to update the acceptable range to a minimum that does not include the version that has the vulnerability. As you mentioned, the use by this project probably doesn't really run into the case that the vulnerability impacts, so it could still be reasonable to say that the version with the vulnerability is still acceptable for this particular project. Fair?

The other benefit of updating the minimum of the range if you were to decide that the problematic version were not acceptable is that npm would then take care of updating to a newer version, both in cases with lockfiles (w/o hand editing the lockfile) and when installs happen into a dirty workspace.

Its not a misunderstanding of semver that raises the request, so I wanted to clarify at least my intent. It won't upset me if you still decline since the vulnerability shouldnt really impact semantic-release.

pvdlg:
Thanks for the feedback! That's a very interesting discussion to have (at least for me ^^).

> With the semver range defined by this package, the version with the vulnerability is an acceptable version that satisfies the defined range. All we are proposing is to update the acceptable range to a minimum that does not include the version that has the vulnerability.

I understand that, but the whole point of using npm and SemVer is to not have to do that. If we update this, then every time a project release a fix, we'll have to update all the projects that depends on it directly or indirectly. This is really going back to the days without npm (that was `maven` for me :)) where version range where not available (no `^` or `~`) and you had to contently update everything and live with bugs because you dependencies were not updating their own dependencies.

Basically I we update the `marked` version in `semantic-release` why not updating all the patch and minor versions of all dependencies every time one is released? If the npm ecosystem can evolve so fast this is because we don't have to do that.

> With the semver range defined by this package, the version with the vulnerability is an acceptable version that satisfies the defined range. All we are proposing is to update the acceptable range to a minimum that does not include the version that has the vulnerability.

Understood, but my point is that it won't change anything for anyone:
- Project that include `semantic-release` as a new dependency (with `npm install semantic-release`) will get the last version of `marked` whether or not we update the `marked` range in `semantic-release`.
- Project that currently use `semantic-release` with a `package-lock.json` will not get the updated version of `marked` until they update their `package-lock.json` whether or not we update the `marked` range in `semantic-release`.
- Project that currently use `semantic-release` without a `package-lock.json` will get the updated version of `marked` when they do `npm install` whether or not we update the `marked` range in `semantic-release`.

> The other benefit of updating the minimum of the range if you were to decide that the problematic version were not acceptable is that npm would then take care of updating to a newer version, both in cases with lockfiles (w/o hand editing the lockfile) and when installs happen into a dirty workspace.

That's true only if you update `semantic-release` manually. Otherwise the current version (with `marked` `^0.3.6`) will still be the one installed due to your `package-lock.json`. No matter what with a `package-lock.json` you will need a manual action.

> Fair?
I'm ok merging this PR if it makes people feels better. I just want to point out that it's a false feeling of security, as this would not solve the problem. If folks use a `package-lock.json` they implicitly (unknowingly?) make the choice of using a consistent but outdated dependency tree. Not matter what we do in `semantic-release` they will have to update their `package-lock.json` if they want to prevent security vulnerabilities.

simlu:
Very interesting discussion in deed!

Here is my point of view on this: We pin every dependency and don't use version ranges. When a new dependency is released we use greekeeper + travis + octoscript + semantic-release to test and immediately publish a new version of the module (everything fully automated). If the dependency update causes a test failure, no new version is release and we are notified about it and can fix it manually.

tldr; I don't trust people. They releases minor version updates with breaking changes all the time. Great way to get bitten in the behind.

pvdlg:
> Here is my point of view on this: We pin every dependency and don't use version ranges.

Pinning dependencies is contrary to the npm logic, and would kill it's main benefit: version ranges that allows to automatically uses non breaking changes (namely patches and new features).

> When a new dependency is released we use greekeeper + travis + octoscript + semantic-release to test and immediately publish a new version of the module (everything fully automated). If the dependency update causes a test failure, no new version is release and we are notified about it and can fix it manually.

With the current setting (range for dependencies) when a new dependency (in range or not) is released, Greenkeeper run the tests automatically to make sure nothing breaks. It currently doesn't work if you use a `package-lock.json` due to greenkeeperio/greenkeeper#506. But we are not using a `package-lock.json` for semantic-release as we prefer to have up to date dependencies in dev rather consistents but outdated ones.

If a patch or minor release of a dependency break something we are already notified immediately by Greenkeeper.

Basically it's already working like what you propose, but with version range, without having to merge a Greenkeeper PR every time a dependency releases a patch or feature, and without having to release a `semantic-release` version every time it happens.

simlu:
@pvdlg I understand your reasoning. This all circles back nicely to this PR. Let me explain our side.

Assume a dependency X of module A get's a minor version update with a critical fix for your module A. No new version of module A get's released. Hence there is a no way for a module B that uses module A to know that something critical has changed (since module A still has the same version). We simply prefer to make changes transparent to all the modules that use our modules.

Exactly the reasoning why we want this PR to be merged and a new version of semantic-release to be released. This prevents some users from having the current semantic release version with an old install and other users using the same semantic release version with a new install. The version are the same, but the functionality is different.

The main argument against this is "release overhead", but if that is all automated it mainly goes away. But this is opinion based. I don't think that part of npm is "great" - this was simply a fix for problems people had "release wars", but it's not necessarily the best solution.

You have a great point there wrt to package-lock.json and we might just abandon it until greenkeeper also updates and fully supports it.

travi:
I'm certainly not proposing needing to increase the minimum of the range every time any package updates. Security vulnerabilities are a special case in my book (but it is also influenced by the fact that github sent a notification for all of my projects that have this in their dep tree). Since this vulnerability doesnt actually impact sematic-release's use, I'd still say its probably not worth extra effort. This is why I only spoke up on an existing issue rather than opening my own :) The rest of this is mostly to talk through the theoretical aspects as long as we are still interested :)

However, I do think there are special cases like this when I think it is appropriate for a project to limit the semver range further than they normally might. The other case is a breakage that didnt bump semver appropriately. If the project knows that there is a version within their specified range that breaks the functionality of their package, I think it is helpful for them to exclude that version from their range.

> Project that currently use semantic-release with a package-lock.json will not get the updated version of marked until they update their package-lock.json whether or not we update the marked range in semantic-release.

This is true except for the manual editing of the lockfile. I may have to "manually" install the new semantic-release version, but by doing so, `npm` takes care of updating the lockfile for me. Regardless of opinions on whether or not a lockfile should be used, we should be able to agree that no human should manually edit that file. that is `npm`'s responsibility. like @simlu, I pin versions in my `package.json` (mostly due to https://github.com/greenkeeperio/greenkeeper/issues/434), but compensate for that by letting the combination of `greenkeeper` and [greenkeeper-keeper](https://github.com/greenkeeper-keeper) keep them up to date for me automatically.

> Project that currently use semantic-release without a package-lock.json will get the updated version of marked when they do npm install whether or not we update the marked range in semantic-release.

This is only true for clean workspaces or when `node_modules/` has been cleared or maybe for a `npm rebuild`, etc.

travi:
@simlu as far as greenkeeper updating your lockfile, i've had good luck so far with `greenkeeper-lockfile` outside of the fact that it has to make an additional commit for the lockfile updates. have you had trouble with that?

pvdlg:
To @simlu 

> Assume a dependency X of module A get's a minor version update with a critical fix for your module A. No new version of module A get's released. Hence there is a no way for a module B that uses module A to know that something critical has changed (since module A still has the same version). We simply prefer to make changes transparent to all the modules that use our modules.

That's not correct. If you define your module A with a version range and the fix is within this range, then Module B that uses Module will get the update automatically, as long as they don't explicitly refuse to get it by using a `package-lock.json`. Also they will get the update automatically in prod as the `package-lock.json` is not used there.
The notion of transparency is handled with Semver: If it's a fix update only the last number of the version.

> Exactly the reasoning why we want this PR to be merged and a new version of semantic-release to be released. This prevents some users from having the current semantic release version with an old install and other users using the same semantic release version with a new install. The version are the same, but the functionality is different.

As explained in this thread this is not correct. Updating semantic-release will not solve that issue.
Users who do not update their dependencies won't get the update. Users who do update their dependencies will get the update, no matter is semantic-release is updated or not, because the `marked` version with the fix is in the range defined by `semantic-release`.

> The main argument against this is "release overhead", but if that is all automated it mainly goes away. But this is opinion based. I don't think that part of npm is "great" - this was simply a fix for problems people had "release wars", but it's not necessarily the best solution.

This choice has been made by npm and it allowed to create a striving ecosystem evolving faster than any other before. The benefit of SemVer might be an opinion, but it's the choice made by npm. The entire objective of semantic-release is to enforce it, so you can imagine we love this SemVer :)

To @travi 

> However, I do think there are special cases like this when I think it is appropriate for a project to limit the semver range further than they normally might. The other case is a breakage that didnt bump semver appropriately. If the project knows that there is a version within their specified range that breaks the functionality of their package, I think it is helpful for them to exclude that version from their range.

Yep, if someone release a breaking in a path or minor release that blows up everything. This is why everyone should use semantic-release and Greenkeeper :) semantic-release will avoid folks to do such thing and Greenkeeper prevent us to be affected by it.

>  Security vulnerabilities are a special case in my book (but it is also influenced by the fact that github sent a notification for all of my projects that have this in their dep tree).

Fair point. But GitHub didn't reported the problem to the semantic-release repo. I imagine it's because the fix for it is in the range we declared?

> This is true except for the manual editing of the lockfile. I may have to "manually" install the new semantic-release version, but by doing so, npm takes care of updating the lockfile for me.

Fair point in that particular case. But keep in mind that updating semantic-release with this PR will solve your specific case for `remark`. But I'm pretty sure that you have in dependency tree dozen or hundreds of sub-dependencies outdated due to your `package-lock.json`. Many of them might have security for which a fix exist and won't get that fix until all the project you depends on update to a range that doesn't allow the bug. But by the time they do that you'll have many bug that would have appeared.

No matter in which side we put the problem, using a `package-lock.json` will have for consequence an outdated dependency tree. This is the whole concept of the `package-lock.json`: guaranteeing a consistent dependency tree by creating a snapshot of it at a given moment. By definition it makes your dependency tree outdated. **If you use a `package-lock.json` you give up the ability to get an up to date dependency tree.**

> Regardless of opinions on whether or not a lockfile should be used, we should be able to agree that no human should manually edit that file.

Ideally npm should provide a way to update a sub-dependency via CLI. But it's not the case....

pvdlg:
I'll go ahead an merge the PR as it seems to be a friction point and it's the specific case of a security issue.

But all the points made in this thread remain and it's illusory to think it is required to prevent to an outdate `marked`.

If other PRs of this type happen we'll update the documentation to explain why they are unnecessary and won't be merged.

simlu:
@pvdlg I think you misunderstood my point. Leaving out the lock file entirely, the problem I have with version ranging is that you no longer differentiate between dependency versions, but between install times. Basically you would have to constantly re-install all your dependencies, because a sub-dependency could have updated - even if non of your direct dependencies updated. And yes, that's how it was designed for npm - it doesn't mean we have to like it.

Anyways, thanks for merging :)

pvdlg:
> Leaving out the lock file entirely, the problem I have with version ranging is that you no longer differentiate between dependency versions, but between install times. Basically you would have to constantly re-install all your dependencies, because a sub-dependency could have updated - even if non of your direct dependencies updated. And yes, that's how it was designed for npm - it doesn't mean we have to like it.

Independently of npm, you have to choose between an automatically up to date but changing dependency tree or a consistent but outdated one. You can't get both.
npm chose the former to be the default and allow to use the later with `npm-shrinkwrap.json` for prod and `package-lock.json` for dev.

If you want a `npm install` to produce a consistent dependency tree, then use a `package-lock.json` but you loose the ability to be always up to date. If you also want to be up to date you have to frequently delete your `pacakge-lock.json` and regenerate it.

You can decide to not use npm with the default `^` and pin your dependencies or to have a `pacakge-lock.json` that you don't fully regenerate frequently, but you can't expect all the projects in the eco-system to all update all their dependencies every time and in-range dependency is released (like I did by merging this PR).

Unfortunately this is a situation in which a choice has to be made:
1. Either you use version range, no `package-lock.json` and Greenkeeper to test in range dependencies releases
2. Or you use a `package-lock.json` to freeze your dependencies (in dev only) and you regenerate it frequently to stay up to date
3. Or you ask all the developers in the world who publish on npm, to always and immediately update all the dependencies of all their packages every time a release is done

Obviously the 3rd choice is not reasonable. That was my point in this thread.

**In conclusion, if you want to use a `package-lock.json` regenerate it frequently to stay up to date.**

simlu:
> Obviously the 3rd choice is not reasonable. That was my point in this thread.

Ah, I'm finally getting it.

Conversely downstream in this example: Even though you updated the marked dependency in semantic-release there is no reason why a project using semantic-release would update semantic-release (and dependencies) unless you do a major version update as well (as they might be using ^).

Lots of food for thought here... Thank you for this discussion!

travi:
> Ideally npm should provide a way to update a sub-dependency via CLI. But it's not the case....

This! We're in the early days of using lockfiles with npm, but I hope things continue in a positive direction to handle things like this.

> In conclusion, if you want to use a package-lock.json regenerate it frequently to stay up to date.

Unfortunately, this doesn't seem to be an action that is enabled very well through the CLI. Do you know of a way to do this effectively? No worries if not since I'm aware you're avoiding their use.

> But GitHub didn't reported the problem to the semantic-release repo. I imagine it's because the fix for it is in the range we declared?

Also a fair point. We're also very early with tools like the vulnerability reporting from tools like github, snyk, nsp, etc. Until tools like that start automatically sending PRs to projects, similar to what greenkeeper does, I think the best we can hope for is for package consumers that get the notifications to open issues to start conversations like this. Thank you for having the patience to work through the conversation with us.

> Obviously the 3rd choice is not reasonable. That was my point in this thread.

I hope its clear that we don't want to suggest that we'd want semantic-release to be responsible to keeping its dependency in our project. This particular dependency simply deserved special consideration around whether the allowed range should be tightened up. Even if we landed on the answer being no in the end, I think the conversation is worthwhile in special cases like this. Sometimes it is worth exceptions to event defined common practices so its doesnt hurt to at least ask :)

> If you use a package-lock.json you give up the ability to get an up to date dependency tree.

I think this is the key point to all of this. Everything is about tradeoffs. Different maintainers will make different decisions about which are most important. For me, predictable but slightly out of date outweighs unpredictably/inconsistently up to date. My use of greenkeeper/greenkeeper-keeper shortens the time that my non-direct dependencies are out of date even further, especially for the direct dependencies that update often. There is no perfect solutions. Dependency management is _hard_. We've all got to find workable solutions that work for us and our teams.

The fact that we are even this deep into this conversation means we are far further in understanding and considering the tradeoffs in this space than most. I really appreciate thinking through a problem like this with you. Projects like semantic-release, greenkeeper, etc have the potential to make these problems far better than vanilla npm for teams that might never get this deep into considering what we're talking about. Thank you for the work you're putting into this space!

pvdlg:
> Unfortunately, this doesn't seem to be an action that is enabled very well through the CLI. Do you know of a way to do this effectively? No worries if not since I'm aware you're avoiding their use.

Sadly no. I'm thinking about building a GitHub app that would run every day, fully update the `yarn.lock` and `package-lock.json` and opens a PR with it. But I didn't had much time to work on that and I'm not really satisfied with opening a PR everyday. The other option would be to push to master directly but that has some constraints as well.

> Even if we landed on the answer being no in the end, I think the conversation is worthwhile in special cases like this. Sometimes it is worth exceptions to event defined common practices so its doesnt hurt to at least ask :)

Absolutely! This is type of conversation is always beneficial. It can be also used or future reference.

> I think this is the key point to all of this. Everything is about tradeoffs. Different maintainers will make different decisions about which are most important. For me, predictable but slightly out of date outweighs unpredictably/inconsistently up to date.

Absolutely it's a tradeoff. My problem is that both yarn and npm, use a lockfile by default and do a terrible job at explaining users what tradeoff they are doing. Most users do not realize the tradeoff they are making.

Actually a lot of users also think that a `yarn.lock` or a `package-lock.json` protect them from experiencing a different tree between their dev environment and their users' execution environment. That's something provided by `npm-shrinkwrap.json` but not by `yarn.lock` and `package-lock.json`. It's so confusing that even the Greenkeeper team, who know a lot about dependency management and npm got confused and introduced greenkeeperio/greenkeeper#506.

Lockfiles are not necessarily a bad idea , but because the concept is complex and their were introduced without much documentation, no warning in the CLI and they are used by default, that create a lot of confusion. The result is that many many projects use them and have dependencies in their dev environment that didn't received any fix/feature for months or years. The more time passes the more their dev environment diverge from the execution environment of their users (as the dev env is frozen but not the exec env) and the more likely it is to have bugs related to that.

I think it's to yarn and npm to:
- make sure the community understand that with lockfiles you need to upgrade them often
- provide tools to do so (better than `rm -rf node_modules && rm -f package-lock.json && npm install`)
- make sure the community understand that lockfile do not impact the exec environment

There is also the notion of speed behind all that. Using a lockfile speed up the installation. Yarn introduced the lockfile first, and their main selling point was speed. They put a lot of emphasis on "we are faster than npm" but not so much on the "at the expense of not having an automatically up to date dependency tree". Then npm had to catch up and introduced `package-lock.json`

> The fact that we are even this deep into this conversation means we are far further in understanding and considering the tradeoffs in this space than most. I really appreciate thinking through a problem like this with you. Projects like semantic-release, greenkeeper, etc have the potential to make these problems far better than vanilla npm for teams that might never get this deep into considering what we're talking about. Thank you for the work you're putting into this space!

Appreciation is shared :-)

travi:
Very well said. All very valid points. I still stand by preferring a lockfile over not for my projects, but that is absolutely because of careful consideration of the tradeoffs that are not well known to many. 

The lack of emphasis on educating the community about the tradeoffs has absolutely caused a ton of problems throughout the community. Unfortunately, it is starting to seem like many conversations, unlike this one, simply land on "lockfiles are hard; give up on them" rather than trying to drive decisions to be educated about the tradeoffs. 

Since communication from npm and yarn falls so short of properly educating, do tools like semantic-release stand in a strong enough position to help fill in the gaps (as if you don't already have enough on your plate)? If nothing else, it seems like the audience of tools in this space is likely to be the most interested in better understanding reality. Educated users of these tools are likely to share knowledge along with the information of the success they've had with these tools.

What else can we do to encourage educated decisions in this area?

pvdlg:
> What else can we do to encourage educated decisions in this area?

I've been working on a Medium post for a while, but did't go time to finish it. I guess that can be a good solution. Based on that maybe try to engage with Yarn / npm people to get their feedback.

Issue #575, opened 2017-12-27
How to set up custom regex for parserOpts ?


jy95:
Hello everyone,

I use the conventional-changelog preset eslint with [sgc ](https://github.com/JPeer264/node-semantic-git-commit-cli) and I would to handle a custom thing introduced by sgc : emoji at the beginning of the commit title.

Now, a build where I use [that ](https://travis-ci.org/jy95/torrent-files-library/jobs/320472340): 

```
[Semantic release]: Call plugin analyze-commits
[Semantic release]: Analyzing commit: :hammer: Refactor: Update description inside package.json
[Semantic release]: The commit should not trigger a release
``` 

and a build without [that ](https://travis-ci.org/jy95/torrent-files-library/jobs/320475836): 

```
[Semantic release]: Call plugin analyze-commits
[Semantic release]: Analyzing commit: Refactor: Remove emoji for sgcrc
[Semantic release]: The release type for the commit is patch 
``` 

My current config in package.json (and the [rules ](https://github.com/jy95/torrent-files-library/blob/master/config/release-rules.js))
```json
"release": {
    "analyzeCommits": {
      "preset": "eslint",
      "releaseRules": "./config/release-rules.js"
    },
    "generateNotes": {
      "preset": "eslint"
    }
  }
``` 

The settings I would like to add (for analyzeCommits and generateNotes) : 

``` 
"parserOpts": {
        "headerPattern": /^(?:\:(\w*)\:)?\s(\w*)\:\s(.*?)(?:\((.*)\))?$/,
        "headerCorrespondence": [
          "emoji",
          "tag",
          "message"
        ]
      }
``` 

The problem is that [JSON ](http://json.org/) file (package.json) allows only primitive values.

So, I think I could be nice to provide a example of **release.config.js** based on this case, isn't it ? (since the current tests doesn't show one )

Thanks for helping.

pvdlg:
Thanks for the feedback and suggestion. Could you just clarify if what you are trying to achieve works and if you are suggesting to add more examples to the doc, or if you are asking for help to make it work?

More generally, the `parserOpts` in `analyzeCommits` and `generateNotes` is there for convenience to allow you to make minor changes to a conventional-changelog preset.

If you want to achieve something more advanced the best option is to implement your own conventional-changelog preset. Here is an example I did a while ago (It's not really up to date but can be of some help): https://github.com/pvdlg/conventional-changelog-metahub.

[...]

PR #593, opened 2018-01-05
Publish a documentation website

pvdlg:
This supersede #591.

This PR redistribute the documentation written in #591 in multiple files to make more accessible and easy to read.

See the version published on [GitBook](https://www.gitbook.com): https://semantic-release.gitbooks.io/semantic-release/content/#highlights

Fix #591


jy95:
It would be a good idea to add this example for release.config.js , no ?
https://github.com/semantic-release/semantic-release/issues/575#issuecomment-354110425

pvdlg:
> It would be a good idea to add this example for release.config.js , no ?

That would be a better fit for the commit-analyzer and note generator plugin.

docs/recipes/gitlab-ci.md
```diff
@@ -0,0 +1,71 @@
...
+
+publish:
+  image: node:8
+  stage: release
+  script:
+    # Only for a local semantic-release installation
+    - npm run semantic-release
+
+    # Only for a global semantic-release installation
+    - npm install semantic-release
```
felixfbecker: Did you mean to add `-g` here?
Why would we want to recommend this approach given it's disadvantages (no Greenkeeper support etc)?
Should in any case have a version requirement so it's not breaking folks on a major release

docs/recipes/travis.md
```diff
@@ -0,0 +1,151 @@
...
+os:
+  - linux
+  - osx
+
+script:
+  # Run tests
+  - go test -v ./...
+
+  # Use nvm to install and use the Node LTS version (nvm is installed on Travis images)
+  - nvm install lts/*
```
felixfbecker: I would always pin this to a specific major version, e.g. 8. Otherwise if a new major version becomes lts this would break the build.

docs/support/FAQ.md
```diff
@@ -0,0 +1,148 @@
...
+```
+
+See the [CI configuration recipes](../recipes/README.md#ci-configurations) for more details on specific CI environments.
+
+## Can I use semantic-release to publish non-JavaScript packages?
+
+Yes, **semantic-release** is a Node CLI application but it can be used to publish any type of packages.
+
+To publish a non-JavaScript package you would need to:
+- Use a [global](../usage/installation.md#global-installation) **semantic-release** installation
```
felixfbecker: I think it makes sense to mention that you can just install it globally in CI if you don't want a package.json, but stating the user _needs_ to is misinformation.

docs/support/FAQ.md
```diff
@@ -0,0 +1,148 @@
...
+
+Yes with the [dry-run options](../usage/configuration.md#dryrun) which prints to the console the next version to be published and the release notes.
+
+## Can I use semantic-release with Yarn?
+
+If you are using a [local](../usage/installation.md#local-installation) **semantic-release** installation and run multiple CI jobs with different versions, the `yarn install` command will fail with Node < 8 as **semantic-release** require [Node >= 8](#why-does-semantic-release-require-node-version--8). See [yarnpkg/rfcs#69](https://github.com/yarnpkg/rfcs/pull/69).
+
+In order to run **semantic-release** with [Yarn](https://yarnpkg.com) you would need to:
+- Use a [global](../usage/installation.md#global-installation) **semantic-release** installation
+- Make sure to install and run the `semantic-release` command only in a CI jobs running with [Node >= 8](#why-does-semantic-release-require-node-version--8)
```
felixfbecker: Wouldn't it be much easier to just use `--ignore-engines`? That makes it behave like npm (only print warnings)

docs/recipes/gitlab-ci.md
pvdlg: > Did you mean to add -g here?

Yes I meant `-g`. Fixed in #594. Thanks.

> Why would we want to recommend this approach given it's disadvantages (no Greenkeeper support etc)?

Because not all projects use a `package.json`.

> Should in any case have a version requirement so it's not breaking folks on a major release

I added a recommendation in #594. Thanks.

docs/recipes/travis.md
pvdlg: We are guaranteeing the support of Node LTS whatever version it is. See [Node Support Policy](https://github.com/semantic-release/semantic-release/blob/caribou/docs/support/node-support-policy.md).

docs/support/FAQ.md
pvdlg: This is what we suggested when the problem was reported, but for some reason it wasn't working due to a bug in `yarn` that was somehow ignoring `--ignore-engines`.

If it's fixed the doc can be changed

docs/recipes/travis.md
felixfbecker: Sure, the _latest_ version of semantic-release guarantees that. But if a project is on an older pinned major version of semantic-release (which it should be for reproducable builds) then a new LTS release can break the build.

docs/support/FAQ.md
pvdlg: By "non-JavaScript package" I implied a project without a `package.json`. I added some clarification in #594

docs/recipes/travis.md
pvdlg: The doc clearly mention is a simplified example.

An example cannot plan for every possible interaction in every situation otherwise it ends up being confusing with a lot of "If this do that".

The person configuring the `.travis.yml` file expected to have a minimum understanding of Travis. The objective of this doc is explain how to use semantic-release, not to explain how to use Travis, or to choose which Node version is best for your project or your company.

The users who configure the `.travis.yml` should be able to analyze by themselve the fact they specifically pin a very old version of semantic-release.

pvdlg:
> How is the website deployed?

With [GitBook](https://www.gitbook.com)

PR #613, opened 2018-01-23
Handle `getLastRelease` with git tags and move to core

pvdlg:
Benefits:
- Prevent the `ENOTINHISTORY` error as the commits is necessarily in the history if associated with a tag
- Prevent the `ENOGITHEAD` error as a commit sha is necessarily associated with a Git tag
- Avoid all the workarounds in npm plugins's `getLastRelease`
- Allow to rebase commits or move them from a branch to another while keeping the version-sha association (the association would be lost when using the npm `gitHead`)
- Simplify recovery solution by simply creating/changing tags
- Simplify plugin development:
  - Do not require plugins to handle Git authentication setup and verification
  - Plugins can now push to the repository without any config or set up
  - Do not require plugins to create a Git tag and handle the situation where the tag already exists

Fix semantic-release/git#15, Ref #193, Ref semantic-release/evolution#1

- [x] Remove the `getLastRelease` plugin type
- [x] Retrieve the last release based on Git tags
- [x] Create the next release Git tag before calling the `publish` plugins
- [x] Remove Git tag creation in GitHub plugin - semantic-release/github#28
- [x] Remove Git tag creation in Git plugin - semantic-release/git#20
- [x] Remove Git tag creation in GitLab plugin - semantic-release/gitlab#8
- [x] Remove `getLastRelease` in Git plugin - semantic-release/git#20
- [x] Remove `getLastRelease` in npm plugin - semantic-release/npm#45
- [x] Remove `getLastRelease` in exex plugin - semantic-release/exec#9
- [x] Remove Git authentication documentation in Git plugin - semantic-release/git#20
- [x] Update integration tests once updated plugins are released

## Breaking changes
- The `getLastRelease` plugins will not be called anymore and the `--get-last-release` CLI option is not allowed anymore
- The Git authentication is now mandatory and must be set via `GH_TOKEN`, `GITHUB_TOKEN`,  `GL_TOKEN`, `GITLAB_TOKEN` or `GIT_CREDENTIALS` as described in [CI configuration](https://github.com/semantic-release/semantic-release/blob/caribou/docs/usage/ci-configuration.md#authentication).
- The `--repositoryUrl` CLI options is replaced by `--repository-url`.

### Make sure the commit associated with the last release is tagged with `v<last_release_version>`

This will already be the case for all semantic-release users using the GitHub (default), Git or GitLab plugins or for any release done manually with `npm publish`.

### Make sure the Git authentication is configured

This will already be the case for all semantic-release users using the GitHub (default), Git or GitLab plugins. See [CI configuration](https://github.com/semantic-release/semantic-release/blob/caribou/docs/usage/ci-configuration.md#authentication).


pvdlg:
Yep we can test on the `@next` branch.

We should also merge #618 before this one.

PR #617, opened 2018-01-24
fix: fix the `--no-ci` arg parsing

### LINKED ISSUE #616: The `--no-ci` option is ignored
## Current behavior

When calling `semantic-release --no-ci`, the parameter is ignored.
`commander` set `options.noCi` as `undefined`, instead the `options.ci` is set to `true`.

In addition the `noCi` options from the config files is loaded only after checking `options.noCi`. So `noCi` from a config file is effectively ignored.

## Expected behavior

`semantic-release --no-ci` should be parsed properly.

pvdlg:
The `noCi` is no properly set based on the `--no-ci` CLI arg and it is overwritten by the `noCi` option.

Fix #616

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -48,8 +48,11 @@ module.exports = async () => {
       program.outputHelp();
       process.exitCode = 1;
     } else {
+      const opts = program.opts();
+      // Set the `noCi` options as commander.js sets the `ci` options instead (becasue args starts with `--no`)
+      opts.noCi = opts.ci === false ? true : undefined;
       // Remove option with undefined values, as commander.js sets non defined options as `undefined`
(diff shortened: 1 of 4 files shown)


cli.js
```diff
@@ -48,8 +48,11 @@ module.exports = async () => {
       program.outputHelp();
       process.exitCode = 1;
     } else {
+      const opts = program.opts();
+      // Set the `noCi` options as commander.js sets the `ci` options instead (becasue args strats with `--no`)
```
travi: typo: "strats" => "starts"

pvdlg: Thanks!

PR #622, opened 2018-01-27
Recommend to execute semantic-release via npx

boennemann:
Historically using an npm script to execute semantic-release has always been a stopgap solution for the rather ugly `semantic-release pre && npm publish && semantic-release post` hack.

Now that this script is no longer necessary (excellent work 👏) I felt like it would be desirable to get rid of the npm script in the first place. This becomes especially true with the differentiation between local and global installs, which was referenced throughout the docs, and makes it necessary to write every configuration guide (or at least parts of it) twice.

Luckily `npx` makes this extremely easy to do. `semantic-release` requires node version 8, which already ships with npx. (node@8.0.0 ships with npm@5.2.0 ships with npx@9.0.3).

`npx` will pickup the binary from the local `./node_modules/.bin` folder, the $PATH, or install semantic-release if necessary. This makes `npx semantic-release` compatible with all the current scenarios.

1. Local: Add to devDeps, npx will pick up the local './node_modules/.bin` 
2. Global: Install globally, npx will pick up the globally installed version
3. *new* Intermediate: npx will do it's thing to install and run semantic-release from its cache folder

This way the decision whether to use a local or a global install doesn't carry through to the execution command. The only difference is now whether semantic-release is installed locally in the package.json or not, which is the only place where this decision should be manifested.

I applied these changes to all scenarios in the docs to illustrate my point. I hope you like these changes. Making the `semantic-release-cli` behave this way would be another to-do then.

Best,
Stephan

### DIFF
--- a/docs/usage/installation.md
+++ b/docs/usage/installation.md
@@ -2,32 +2,28 @@
 
 ## Local installation
 
-For [Node modules projects](https://docs.npmjs.com/getting-started/creating-node-modules) we recommend to install **semantic-release** locally and to run the `semantic-release` command with a [npm script](https://docs.npmjs.com/misc/scripts):
+For [Node modules projects](https://docs.npmjs.com/getting-started/creating-node-modules) we recommend to install **semantic-release** locally and to run the `semantic-release` command with [npx](https://www.npmjs.com/package/npx):
 
 ```bash
 $ npm install --save-dev semantic-release
 ```
 
-In your `package.json`:
-
-```json
-"scripts": {
-  "semantic-release": "semantic-release"
-}
-```
-
 Then in the CI environment:
 
 ```bash
-$ npm run semantic-release
+$ npx semantic-release
 ```
+
+**Note:**: `npx` is a tool bundled with `npm@>=5.2.0`. It is used to conveniently find the semantic-release binary and to execute it. See [What is npx](../support/FAQ.md#what-is-npx) for more details.
 
 ## Global installation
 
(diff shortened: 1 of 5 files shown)


pvdlg:
Thanks! I think it's a great change!

We just need to make sure that `npx` is available where we recommend to use it. On Travis it's always available independently of the Node version that the job runs. But on CI it might not be the case.
On all the modifications in this PR, `npx` should be available, so we're good to go!

I just have one remark in the Global install (see inline comment).

docs/usage/installation.md
```diff
@@ -2,32 +2,32 @@
...
-For other type of projects we recommend to install **semantic-release** globally directly in the CI environment:
+For other type of projects we recommend to install **semantic-release** directly in the CI environment:
 
 ```bash
+# Option 1: npx
+$ npx semantic-release
+```
+
+```bash
+# Option 2: npm global install
```
pvdlg: IT seems there is a problem with option 2 as we would install semantic-release globally and run it twice (once with `semantic-release` and once with `npx semantic-release`).

Overall we probably don't need two options but just `npx semantic-release` as `npx` would take care of the install.

boennemann:
@pvdlg @gr2m Thanks for your feedback!

I removed the `npm i -g semantic-release` part completely, as per @pvdlg's comment.

I added an FAQ entry "What is npx" and linked it up, also in the installation section where we recommend npx to install node@8.

Hope this works now :)

boennemann:
I'm so eager to get this out, I even added Yarn instructions 🙈

PR #625, opened 2018-01-28
Fix image path

newyork-anthonyng:
Before:
screen shot 2018-01-28 at 1 15 46 pm

After:
screen shot 2018-01-28 at 1 16 09 pm


gr2m:
Thank you @newyork-anthonyng 💐

Issue #629, opened 2018-01-31
Remove remote tag remotely fails due to invalid shorthand


crash7:
Seems that this part of the code it's not compatible with the git-version in the image circleci/node:8.9:
https://github.com/semantic-release/semantic-release/blob/39536fa34e64234fe530adb37b19380d5fa6a699/lib/git.js#L105-L113

Using --delete directly should fix this!

## Current behavior
```
semantic-release:get-version-head delete remote tag { Error: Command failed: git push -d https://[secure]@github.com/secret/more-secret.git v1.0.0
error: unknown switch `d'
usage: git push [<options>] [<repository> [<refspec>...]]
```

## Expected behavior

The tag it's removed correctly

## Environment

- **semantic-release** version: 13.1.0
- CI environment: circleci
- GIT version: git version 2.1.4
- Plugins used: exec, git and github
- **semantic-release** configuration: 
- CI logs: added relevant part in current behavior
- Help of git push 
```
circleci@f:~/project$ git push -h
usage: git push [<options>] [<repository> [<refspec>...]]

    -v, --verbose         be more verbose
    -q, --quiet           be more quiet
    --repo <repository>   repository
    --all                 push all refs
    --mirror              mirror all refs
    --delete              delete refs
    --tags                push tags (can't be used with --all or --mirror)
    -n, --dry-run         dry run
    --porcelain           machine-readable output
    -f, --force           force updates
    --force-with-lease[=<refname>:<expect>]
                          require old value of ref to be at this value
    --recurse-submodules[=<check>]
                          control recursive pushing of submodules
    --thin                use thin pack
    --receive-pack <receive-pack>
                          receive pack program
    --exec <receive-pack>
                          receive pack program
    -u, --set-upstream    set upstream for git pull/status
    --progress            force progress reporting
    --prune               prune locally removed refs
    --no-verify           bypass pre-push hook
    --follow-tags         push missing but relevant tags

```

crash7:
I changed the line `shell = await execa('git', ['push', '-d', origin, tagName], {reject: false});` to `shell = await execa('git', ['push', '--delete', origin, tagName], {reject: false});` and everything solved!

pvdlg:
Thanks for reporting! Would you like to open a PR to fix it?

crash7:
@pvdlg sure! 😄

PR #631, opened 2018-01-31
fix: use long option to delete tag in git push

crash7:
Fixes https://github.com/semantic-release/semantic-release/issues/629


pvdlg:
Thanks!

PR #634, opened 2018-02-02
fix: always transform `git+https` URLs to `https`

pvdlg:
Git url with format `git+https` are not supported when used in `git push <URL> ref`. It errors out with `fatal: Unable to find remote helper for 'git+https'`.

In some rare cases `get-git-auth-url.js` is producing a `git+https` URL.
It happens if:
- The repository is a GitHub `https` URL as `normalize-package-data` transforms it to a `git+https` URL
- No Git authentication is set (for example when the user uses ssh keys) as in that case `get-git-auth-url.js` returns the URL as is

### DIFF
--- a/lib/get-git-auth-url.js
+++ b/lib/get-git-auth-url.js
@@ -19,11 +19,12 @@ module.exports = repositoryUrl => {
     ? `gitlab-ci-token:${process.env[envVar]}`
     : process.env[envVar];
 
-  if (!gitCredentials) {
-    return repositoryUrl;
-  }
-
   const {protocols} = gitUrlParse(repositoryUrl);
   const protocol = protocols.includes('https') ? 'https' : protocols.includes('http') ? 'http' : 'https';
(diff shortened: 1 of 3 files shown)


gr2m:
can you double check if this works with private repositories?

pvdlg:
Do you mean private GitHub repository? What would be the difference between public and private repo regarding?
The problem is related to the `git` CLI not handling the `git+https` type of URL when specified in a `git push` command.

gr2m:
just some wage problems I run into in the past, I can recall it exactly ... but sounds like they wouldn’t apply. Let me review the code :)

brettstack:
Thanks, I'm running into this. Any ETA on next NPM release?

gr2m:
@brettstack it’s all automatically released using `@next` dist-tag, this one in https://github.com/semantic-release/semantic-release/releases/tag/v13.1.3

PR #637, opened 2018-02-04
feat: add `success` and `fail` notification plugins

### LINKED ISSUE #627: Proposal: Success/Fail notification
This is a proposal to add the `success` and `fail` plugin hook to alert users when a release is successful or failed.

Benefits:
- Alert the user when a problem happen during the release. Currently the only way for user to know a problem happened is to rely on the CI job failure notifications.
- Allow to give more detailed information to users when a release fails. Currently users have to look at the CI logs to figure out what happened.
- Allow to deliver notifications for successful releases that contains information about all the `publish` plugin that ran, therefore allowing to send a message that contains the link to the github release, the npm package etc...

**Add a return value to `publish` hook**

The `publish` hooks could return an object with information related to the release that the plugin did, such as an URL, or additional information.

**`succes` hook**
Accept an Array of plugins (similar to the `publish` hook). Each one would be called by the core, when the release is successful an would be passed the same information as the `publish` hook + an Array of releases. Each release in the Array would be the return value of the `publish` hooks + some additional info such as the plugin that did the release.

**`fail` hook**
Would be called if the release fails and would be passed an Array of SemanticReleaseErrors.

**Add `detailedMessage` to `@sematnic-release/error`**
When throwing a `SemanticReleaseError` the core and plugins would add the `detailedMessage` that would contains:
- A detailed and user friendly explanation of the problem
- Possible solutions
- Links to documentation and help

This detailed error could be used by the `fail` hook to generate a user friendly and helpful message.

**GitHub plugin `fail` hook**
Open an issue with the detailedMessage of each error that happened

**GitHub plugin `success` hook**
Add a comment on each issue/PR that was closed by the release, with a link to each release (on npm, GitHub, Atom etc...).
Add a list of links to the GitHub releases (i.e the npm package or Atom package link, or info about the compiled files pushed to the repo by the `git` plugin)

**Gitter plugin with `fail` and `success` hook**
Post a message on a Gitter chat with information about the releases.

**Slack plugin with `fail` and `success` hook**
Post a message on a Slack chat with information about the releases.

pvdlg:
Fix #627

- Allow `publish` plugins to return an `Object` with information related to the releases
- Add the `success` plugin hook, called when all `publish` are successful, receiving a list of release
- Add the `fail` plugin hook, called when an error happens at any point, receiving a list of errors
- Add detailed message for each error


pvdlg:
Regarding the error `details` message defined in [lib/definitions/errors.js](https://github.com/semantic-release/semantic-release/blob/e8a0a57a785a523861d0316d0733e3a56c196b2d/lib/definitions/errors.js#L1) an alternative would be to use [handlebars.js](https://github.com/wycats/handlebars.js) to generate them.
This way we would have a directory that would contain markdown files with `handlebars.js` variable. The benefit would be a better experience when editing the detailed error message as the IDE would help formatting proper markdown.

But as the message as fairly simple it might be overkill to use `handlebars.js`.

@gr2m do you have thoughts on that?

gr2m:
no opinions either way. I don't think that it's too complex to need a templating language, and I don't think it will change a lot, so probably we are fine without mustache?

pvdlg:
Ok. Let's go. Hopefully #627 should receive a comment after the merge and release.

travi:
nice work! this is a big win. really excited for this, both in my own projects and those i'm a consumer of or try to find time to contribute to :)

PR #645, opened 2018-02-07
fix: get npm global node_modules to resolve global dependencies

trollr:
We need the option to use global installed node_modules e.g. for dockerized semantic-release independent of tech stack (java, go, node, ..).

This is not the smartest solution and just a suggestion which functionality is needed.

### DIFF
--- a/lib/plugins/normalize.js
+++ b/lib/plugins/normalize.js
@@ -3,6 +3,7 @@ const {inspect} = require('util');
 const SemanticReleaseError = require('@semantic-release/error');
 const {isString, isObject, isFunction, noop, cloneDeep} = require('lodash');
 const resolveFrom = require('resolve-from');
+var childProcess = require('child_process');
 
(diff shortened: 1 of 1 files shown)


pvdlg:
After the analysis we did together on Gitter it's due to the way you call `npx`.

You were installing this way:
```bash
npm install -g semantic-release/semantic-release#git-debug
npm install -g @semantic-release/git @semantic-release/changelog
npm install -g @semantic-release/git @semantic-release/git
```
and then calling `npx` with `npx semantic-release/semantic-release#git-debug --dry-run`.

Doing so make `npx` reinstall `semantic-release` rather than using the global install. And when doing so `npx` was installing `semantic-release` in its own `node_modules` which is different than the global one.

If you do `npx semantic-release --dry-run` then `npx` will run the globally installed version as expected.

I'm closing this PR as everything works as expected. Do not hesitate to ping here or on Gitter if you have any other issue.

PR #659, opened 2018-02-11
Add Travis Build Stages and CircleCI recipes

pvdlg:
Fix #573

### DIFF
--- a/docs/recipes/travis-build-stages.md
+++ b/docs/recipes/travis-build-stages.md
@@ -1,1 +1,95 @@
 # Using semantic-release with [Travis CI build stages](https://docs.travis-ci.com/user/build-stages)
+
+## Environment variables
+
+The [Authentication](../usage/ci-configuration.md#authentication) environment variables can be configured in [Travis Repository Settings](https://docs.travis-ci.com/user/environment-variables/#defining-variables-in-repository-Settings) or with the [travis env set CLI](https://github.com/travis-ci/travis.rb#env).
+
+Alternatively, the default `NPM_TOKEN` and `GH_TOKEN` can be easily [setup with semantic-release-cli](../usage/ci-configuration.md#automatic-setup-with-semantic-release-cli).
+
+## Multiple Node jobs configuration
+
+### `.travis.yml` configuration for multiple Node jobs
+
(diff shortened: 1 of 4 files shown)


docs/recipes/travis-build-stages.md
```diff
@@ -1 +1,114 @@
...
+  - 4
+
+os:
+  - linux
+  - osx
+
+# Run the default test stage first, then the release one only when all the test jobs are successful
+stages:
+  - test
+  - release
```
gr2m: I'm not sure we need this, I think that is the default?

docs/recipes/travis-build-stages.md
```diff
@@ -1 +1,114 @@
...
+language: node_js
+
+node_js:
+  - 8
+  - 6
+  - 4
+
+os:
+  - linux
+  - osx
```
gr2m: I'd leave out the `os:` setting for simplicity

docs/recipes/travis-build-stages.md
```diff
@@ -1 +1,114 @@
...
+
+**Advanced configuration**: Running the tests in the `script` step of the `release` stage is not necessary as the previous stage(s) already ran them. To increase speed, the `script` step of the `release` stage can be overwritten to skip the tests. Note that other commands such as build or compilation might still be required.
+
+```yaml
+language: node_js
+
+node_js:
+  - 8
+  - 6
+  - 4
```
gr2m: Maybe we can leave out `4` already? Either way okay with me

docs/recipes/travis-build-stages.md
pvdlg: Indeed it's the default and we don't need it. I added it for more clarity.
But we can remove it.
Do you think it's more confusing to keep it or to remove it?

docs/recipes/travis-build-stages.md
pvdlg: I removed them in both Travis and Travis build stage

docs/recipes/travis-build-stages.md
pvdlg: I removed Node 4. But we should keep 6 and 8 as the point of the recipe is to show how to test on multiple Node versions

docs/recipes/travis-build-stages.md
pvdlg: I removed it

docs/recipes/travis-build-stages.md
gr2m: good point

PR #660, opened 2018-02-12
fix: fix `EPLUGINCONF` error details

pvdlg:


lib/definitions/errors.js
```diff
@@ -53,13 +53,13 @@ Your configuration for the \`tagFormat\` option is \`${stringify(tagFormat)}\`.`
...
-    details: `The [${pluginName} plugin configuration](${linkify(
-      `docs/usage/plugins.md#${toLower(pluginName)}-plugin`
+  EPLUGINCONF: ({pluginType, pluginConf}) => ({
+    message: `The \`${pluginType}\` plugin configuration is invalid.`,
+    details: `The [${pluginType} plugin configuration](${linkify(
+      `docs/usage/plugins.md#${toLower(pluginType)}-plugin`
     )}) if defined, must be a single or an array of plugins definition. A plugin definition is either a string or an object with a \`path\` property.
 
-    Your configuration for the \`${pluginName}\` plugin is \`${stringify(pluginConf)}\`.`,
+    Your configuration for the \`${pluginType}\` plugin is \`${stringify(pluginConf)}\`.`,
```
gr2m: how did that change any behavior? Isn't it just a variable rename?

pvdlg: I was passing an object with `pluginType` to the error function but I was using `pluginName` instead. I resulted in a message with `undefined`.

pvdlg: The parameter is passed as `{pluginName, pluginConf}` which is one Object parameter with `pluginName` and `pluginConf` properties.

Issue #649, opened 2018-02-08
Multiple commits tagged when using `git` plugin


effervescentia:
## Current behavior

In a configuration with custom `publish` plugins, a tag is created and pushed for the initial commit hash before the `@semantic-release/github` and `@semantic-release/git` plugins are being used to create and tag a new commit with an updated changelog from `@semantic-release/changelog`
Because of this my CI now gets 2 webhook notifications for new tags being pushed and has to run 2 deployment workflows 

## Expected behavior

When overriding the `publish` plugins nothing else should create or push tags to github

## Environment

- **semantic-release** version: `13.1.3`
- CI environment: Circle CI
- Plugins used: `@semantic-release/changelog`, `@semantic-release/git`, `@semantic-release/github`
- **semantic-release** configuration:
```
"release": {
    "tagFormat": "v${version}-dev",
    "verifyConditions": [
      "@semantic-release/changelog",
      "@semantic-release/git",
      "@semantic-release/github"
    ],
    "publish": [
      {
        "path": "@semantic-release/changelog",
        "changelogFile": "DEV_CHANGELOG.md"
      },
      {
        "path": "@semantic-release/git",
        "message": "chore(dev release): ${nextRelease.version}-dev [skip ci]\n\n${nextRelease.notes}",
        "assets": "DEV_CHANGELOG.md"
      },
      "@semantic-release/github"
    ]
  },
```
- CI logs: (sorry, private CI)
```
[Semantic release]: Load plugin verifyConditions from @semantic-release/git in shareable config ./release.dev
[Semantic release]: Load plugin verifyConditions from @semantic-release/github in shareable config ./release.dev
[Semantic release]: Load plugin analyzeCommits from @semantic-release/commit-analyzer
[Semantic release]: Load plugin generateNotes from @semantic-release/release-notes-generator
[Semantic release]: Load plugin publish from @semantic-release/changelog in shareable config ./release.dev
[Semantic release]: Load plugin publish from @semantic-release/git in shareable config ./release.dev
[Semantic release]: Load plugin publish from @semantic-release/github in shareable config ./release.dev
[Semantic release]: Run automated release from branch master
[Semantic release]: Call plugin verify-conditions
[Semantic release]: Verify GitHub authentication
[Semantic release]: Found git tag v0.2.1-dev associated with version 0.2.1
[Semantic release]: Found 1 commits since last release
[Semantic release]: Call plugin analyze-commits
[...]

PR #663, opened 2018-02-12
feat: add the `prepare` plugin hook

pvdlg:
Proof of concept for #649.
Related to semantic-release/npm#52.

This PR introduce the `prepare` plugin hook, that will be called before create the Git Tag and the `publish` hook.

Currently plugin that create a file that can be committed (e.g. a `CHANGELOG.md`), create a commit or publish to a registry all operate in the `publish` hook. As some plugins require to have the Git tag created in the remote repo before publishing we currently have to create the tag before calling the `publish` plugins and update the tag whenever a plugin create a commit.
This is a quite inelegant solution...

That leads to creation/deletion/re-creation of Git tags for each plugin doing a commit as described in #649.

The addition of the `prepare` hook make things more organized and cleaner:
- `prepare` => Do all the things required before pushing the Git tag, which can include creating a `CHANGELOG.md`, update the `package.json`, create compiled files, make a commit etc...
- `publish` => Do all the things required after pushing the Git tag, which is mostly publish the release

BREAKING CHANGE: Committing or creating files in the `publish` plugin hook is not supported anymore and now must be done in the `prepare` hook

Plugins with a `publish` hook that makes a commit or create a file that can be committed must use the `prepare` hook, otherwise the commit will be done after the Git tag is pushed.

TODO:
- [x] Change the `@semantic-release/npm` dependency once semantic-release/npm#52 is released
- [x] Change `@semantic-release/changelog` plugin `publish` to `prepare`=> merge semantic-release/changelog#12
- [x] Change `@semantic-release/git` plugin `publish` to `prepare`=> merge semantic-release/git#35
- [x] Add `prepare` to `@semantic-release/exec` plugin => merge semantic-release/exec#14

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,7 @@ After running the tests the command `semantic-release` will execute the followin
 | Verify release    | Verify the release conformity with the [verify release plugins](docs/usage/plugins.md#verifyrelease-plugin).                                                          |
 | Generate notes    | Generate release notes with the [generate notes plugin](docs/usage/plugins.md#generatenotes-plugin) for the commits added since the last release.                     |
 | Create Git tag    | Create a Git tag corresponding the new release version                                                                                                                |
+| Prepare           | Prepare the release with the [prepare plugins](docs/usage/plugins.md#prepare-plugin).                                                                                 |
 | Publish           | Publish the release with the [publish plugins](docs/usage/plugins.md#publish-plugin).                                                                                 |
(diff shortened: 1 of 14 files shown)


pvdlg:
> Code changes look good! Just to be clear, this is not committing the version to package.json in the main branch on GitHub, right?

No. Nothing change in that regards.
We just split `publish` into `publish` and `prepare` so it makes it easier to use the `git` plugin to push the updated `package.json`. But it's still optional and has to be set up by using the `git` plugin.

PR #668, opened 2018-02-13
fix: remove the github plugin from default success and fail hooks

### LINKED ISSUE #664: Inclusion of default github implementation for new 'success' and 'failure' hooks may be a breaking change
If a user is not using any github plugin features in their project in semantic-release version 13.2, the release will suddenly fail when upgrading to 13.3. If the user doesn't have GitHub authentication configured for the project, semantic-release will fail with ENOGHTOKEN.

pvdlg:
Fix #664 

TODO:
- [ ] Deprecate `13.3.0`, `13.3.1` and `13.4.0` on npm
- [ ] Make a breaking change commit that re-enable `@semantic-release/github` as the default for `success` and `fail`
- [ ] Update the release note of `14.0.0` with a migration guide, indicating to users who don't use `@semantic-release/github` to set `success` and `fail` to `false`

No need to change the `peerDependencies` of any plugin.

### DIFF
--- a/docs/usage/configuration.md
+++ b/docs/usage/configuration.md
@@ -166,3 +166,27 @@ CLI argument: `--publish`
 Define the list of [publish plugins](plugins.md#publish-plugin). Plugins will run in series, in the order defined in the `Array`.
 
 See [Plugins configuration](plugins.md#configuration) for more details.
+
+### success
+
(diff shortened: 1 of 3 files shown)


pvdlg:
Ok all done!

PR #671, opened 2018-02-13
fix: do not transform repositoryUrl if it allow to push

### LINKED ISSUE #667: git repository url forcefully converted to https
# Bug report

## Current behavior

I'm trying to release a project hosted on gitlbab. 
I'm using ssh agent for the git operations and I'm providing a GL_TOKEN env variable for the gitlab release.
the repository.url is in the package.json is in the form `git@hostname:group/project.git`

Unluckily semantic-release always convert the repo url to `https://hostname/group/project.git` unless it's explicitly using the `http` protocol (https://github.com/semantic-release/semantic-release/blob/caribou/lib/get-git-auth-url.js#L23) and tries to use the GL_TOKEN for authentication. 

As my gitlab is using http protocol the `https://hostname/group/project.git` endpoint is invalid and the release fails at the `verifyAuth` step (https://github.com/semantic-release/semantic-release/blob/caribou/lib/git.js#L89) as that will call `git  push --dry-run https://hostname/group/project.git HEAD:${branch}`

I don't see the reason why the semantic-release should forcefully use an http/https endpoint when git protocol (with ssh credential) is available

If i change the repository.url to `http://hostname/group/project.git` then everything works correctly (but ssh agent is not used)

## Expected behavior

The git operations should be performed suing the git url (hence using the ssh key) while still be able to call the gitlab api using the provided GL_TOKEN

## Environment

- **semantic-release** version: 13.4
- CI environment: jenkins but can reproduce locally
- Plugins used: npm, git, gitlab
- **semantic-release** configuration: 
```
"release": {
    "branch": "master",
    "debug": true,
    "verifyConditions": [
      "@semantic-release/git",
      "@semantic-release/gitlab"
    ],
    "publish": [
      {
        "path": "@semantic-release/npm",
        "npmPublish": false,
        "tarballDir": "artifacts"
      },
      "@semantic-release/git",
      "@semantic-release/gitlab"
    ],
    "success": [],
    "fail": [],
    "gitlabUrl": "http://someip"
  },
```

pvdlg:
Fix #667

Use the unmodified configured repositoryUrl (without adding the credentials) if it works, otherwise do as of now and convert the URL to `http`/`https` and add the credentials.

This allow users to push tags and commits via ssh while still using the GitHub/GitLab API.

### DIFF
--- a/lib/get-config.js
+++ b/lib/get-config.js
@@ -3,7 +3,7 @@ const readPkgUp = require('read-pkg-up');
 const cosmiconfig = require('cosmiconfig');
 const resolveFrom = require('resolve-from');
 const debug = require('debug')('semantic-release:config');
-const {repoUrl} = require('./git');
+const {repoUrl, verifyAuth} = require('./git');
 const PLUGINS_DEFINITIONS = require('./definitions/plugins');
 const plugins = require('./plugins');
 const getGitAuthUrl = require('./get-git-auth-url');
@@ -51,7 +51,9 @@ module.exports = async (opts, logger) => {
     ...pickBy(options, option => !isUndefined(option) && !isNull(option)),
   };
 
-  options.repositoryUrl = options.repositoryUrl ? getGitAuthUrl(options.repositoryUrl) : options.repositoryUrl;
+  if (!await verifyAuth(options.repositoryUrl, options.branch)) {
+    options.repositoryUrl = options.repositoryUrl ? getGitAuthUrl(options.repositoryUrl) : options.repositoryUrl;
+  }
 
   debug('options values: %O', options);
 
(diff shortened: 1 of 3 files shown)


lib/get-config.js
```diff
@@ -51,7 +51,10 @@ module.exports = async (opts, logger) => {
     ...pickBy(options, option => !isUndefined(option) && !isNull(option)),
   };
 
-  options.repositoryUrl = options.repositoryUrl ? getGitAuthUrl(options.repositoryUrl) : options.repositoryUrl;
+  console.log(options.repositoryUrl, options.branch);
```
brokenmass: do you want this `console.log` ? shouldn't be a `debug()` ?

pvdlg: Oops that was for dev debugging and I forgot to remove it. Good catch!

brokenmass:
I tested in my environment and now it's working correctly so all good for me, you can merge.

pvdlg:
Thanks !

Issue #699, opened 2018-03-13
git verifyAuth failed on public Github repository


escalate:
## Current behavior

During the CI run I receive a failure of the verifyAuth function (lib/git.js).
The git push in dry-run mode to my public Github repository does not work.
The call does not use the GH_TOKEN environemt variable to authenticate.
The failure seems to be not very crtitcal because the following steps are running without errors.
The step "Verify GitHub authentication" works correct so there is no real problem.

## Expected behavior

1. No simple verification test of the git repository because the connectivity to Github will be checked later.

OR

2. The verification test should use the GH_TOKEN environment variable to authenticate.

## Environment

- **semantic-release** version: latest
- CI environment: Travis CI https://github.com/escalate/loxone-ha-bridge-importer/blob/master/.travis.yml
- Plugins used: github, commit-analyzer, release-notes-generator
- **semantic-release** configuration: https://github.com/escalate/loxone-ha-bridge-importer/blob/master/.releaserc
- CI logs:

```
...
  py36: commands succeeded
  flake8: commands succeeded
  congratulations :)

travis_time:end:19ce6194:start=1520925358526575323,finish=1520925376365937399,duration=17839362076
[0K
[32;1mThe command "tox" exited with 0.[0m
travis_fold:start:after_success.1
[0Ktravis_time:start:079149b4
[0K$ nvm install lts/*
Downloading and installing node v8.10.0...
Downloading https://nodejs.org/dist/v8.10.0/node-v8.10.0-linux-x64.tar.xz...
Computing checksum with sha256sum
Checksums matched!
Now using node v8.10.0 (npm v5.6.0)

travis_time:end:079149b4:start=1520925376373571227,finish=1520925379187824204,duration=2814252977
[0Ktravis_fold:end:after_success.1
[0Ktravis_fold:start:after_success.2
[0Ktravis_time:start:02d84422
[0K$ npx travis-deploy-once "npx semantic-release --debug"
npx: installed 246 in 8.829s
[Travis Deploy Once]: There is only one job for this build.
npx: installed 387 in 13.296s
[Semantic release]: Running semantic-release version 15.0.3
2018-03-13T07:16:43.735Z semantic-release:git Error: Command failed: git push --dry-run https://github.com/escalate/loxone-ha-bridge-importer.git HEAD:master
remote: Invalid username or password.
fatal: Authentication failed for 'https://github.com/escalate/loxone-ha-bridge-importer.git/'


[...]

PR #703, opened 2018-03-16
fix: prevent git CLI to prompt user/password on CI

pvdlg:
Per https://github.com/semantic-release/semantic-release/issues/699#issuecomment-373445979

### DIFF
--- a/index.js
+++ b/index.js
@@ -22,6 +22,10 @@ async function run(options, plugins) {
   if (!isCi && !options.dryRun && !options.noCi) {
     logger.log('This run was not triggered in a known CI environment, running in dry-run mode.');
     options.dryRun = true;
+  } else {
+    // When running on CI, prevent the `git` CLI to prompt for username/password. See #703.
+    process.env.GIT_ASKPASS = 'echo';
+    process.env.GIT_TERMINAL_PROMPT = 0;
   }
 
   if (isCi && isPr && !options.noCi) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -22,6 +22,10 @@ async function run(options, plugins) {
   if (!isCi && !options.dryRun && !options.noCi) {
     logger.log('This run was not triggered in a known CI environment, running in dry-run mode.');
     options.dryRun = true;
+  } else {
+    // When running on CI, prevent the `git` CLI to prompt for username/password
```
gr2m: I would suggest we add the link to this PR here for easier reference in future: https://github.com/semantic-release/semantic-release/issues/699#issuecomment-373445979

pvdlg: Should the comment and [GitHub git blame](https://help.github.com/articles/tracing-changes-in-a-file/) be enough?
Personally I use Git blame really often, but not sure if it's a well known feature.

I'm not sure if in general it's better to reference PR/issues as comment versus using GitHub/Git feature. Or at least I don't know in which cases it worst it to reference an PR/issue in a comment. Any suggestion?

AVVS:
@pvdlg @gr2m It doesnt seem to solve the issue

I'm not sure if its some sort of weirdness with env not being passed down with execa, but thats what I get when manually testing, notice that when I pass env variables manually, its all good, but process.env thats been modified inside the nodejs process is actually not being treated by execa anyhow. Also the timeout option itself doesnt trigger for some reason

screen shot 2018-03-18 at 9 28 13 pm

gr2m:
Could you please create a follow up issue? It will be easier for us to track

AVVS:
@gr2m done :)

PR #704, opened 2018-03-16
Remove unused dependency

honzajavorek:
It doesn't seem Semantic Release still uses commander anywhere


pvdlg:
Good catch! Thanks!

PR #710, opened 2018-03-19
fix: prevent git prompt before permissions verification

### LINKED ISSUE #707: git CLI still prompts user/password on CI
Copy from https://github.com/semantic-release/semantic-release/pull/703#issuecomment-374290461

I'm not sure if its some sort of weirdness with env not being passed down with execa, but thats what I get when manually testing, notice that when I pass env variables manually, its all good, but process.env thats been modified inside the nodejs process is actually not being treated by execa anyhow. Also the timeout option itself doesnt trigger for some reason

screen shot 2018-03-18 at 9 28 13 pm

pvdlg:
Fix #707 

This also remove the `execa` timeout as it doesn't work with a prompt. I'll investigate more for an alternative solution to implement a timeout in such case.

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,8 +11,10 @@ const getNextVersion = require('./lib/get-next-version');
 const getCommits = require('./lib/get-commits');
 const getLastRelease = require('./lib/get-last-release');
 const {extractErrors} = require('./lib/utils');
+const getGitAuthUrl = require('./lib/get-git-auth-url');
 const logger = require('./lib/logger');
-const {unshallow, gitHead: getGitHead, tag, push} = require('./lib/git');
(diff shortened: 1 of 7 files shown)


pvdlg:
I tested there: https://semaphoreci.com/vanduynslagerp/test-semaphore/branches/master/builds/45
It seems to work ok now and the git auth check doesn't ask for the password and fail as expected.

Testing unreleased version on Semaphore is a pain as when installing a package from a branch, Semaphore just ignores it and install from master. So I had to make a release in `@pvdlg/semantic-release`. Also it seems there is some sort of cache on Semaphore so even if the last release is `1.0.3` on semaphore the version `1.0.0` is installed, so I had to specify `npm install -g @pvdlg/semantic-release@1.0.2`.

PR #730, opened 2018-04-04
Update plugins.md (for consistency)

bertho-zero:


pvdlg:
Thanks for the PR!

In order to be consistent with the rest of the docs, the change should be the other way around: `github` => `@semantic-release/github`, `npm` => `@semantic-release/npm`, etc...

As plugins are npm packages it make more sense to reference them by their npm name (which include the scope). In addition it allows to differentiate `npm` (as in https://github.com/npm/npm) from `@semantic-release/npm`.

Could you make the change to your PR?

pvdlg:
Thanks!

Issue #741, opened 2018-04-10
Some issue of VSTS url while using PAT


roger199585:
## Current behavior
- Wrong repository while running `verifyAuth`
- `verifyAuth` check the verification by `git push --dry-run https://companyname.visualstudio.com/DefaultCollection/MyRepo.git HEAD:BUILD`
- The log is below
- The issue is in `get-git-auth-url.js`

## Expected behavior
On VSTS the <repositoryUrl> we need is like `https://companyname.visualstudio.com/DefaultCollection/_git/MyRepo`,
I have try it and it work when I change the `<repositoryUrl>` into the format above.
I'll send the PR later.

## Environment

- **semantic-release** version: 15.1.5
- CI environment: VSTS
- CI logs: 
```
2018-04-10T12:00:43.9592602Z fatal: Authentication failed for 'https://[secure]@joueapp.visualstudio.com/Joue/NgxSidebar.git/'
2018-04-10T12:00:43.9618081Z 
2018-04-10T12:00:43.9632042Z 
2018-04-10T12:00:43.9660310Z     at makeError (/vsts/agent/_work/1/s/node_modules/semantic-release/node_modules/execa/index.js:172:9)
2018-04-10T12:00:43.9691120Z     at Promise.all.then.arr (/vsts/agent/_work/1/s/node_modules/semantic-release/node_modules/execa/index.js:277:16)
2018-04-10T12:00:43.9719535Z     at <anonymous>
2018-04-10T12:00:43.9749730Z     at process._tickCallback (internal/process/next_tick.js:188:7)
2018-04-10T12:00:43.9777288Z [Semantic release]: EGITNOPERMISSION The push permission to the Git repository is required.
2018-04-10T12:00:43.9805753Z semantic-release cannot push the version tag to the branch BUILD on remote Git repository.
2018-04-10T12:00:43.9820889Z 
2018-04-10T12:00:43.9849821Z Please refer to the authentication configuration documentation (https://github.com/semantic-release/semantic-release/blob/caribou/docs/usage/ci-configuration.md#authentication) to configure the Git credentials on your CI environment.
2018-04-10T12:00:43.9861303Z 
2018-04-10T12:00:43.9869573Z 
2018-04-10T12:00:43.9886799Z { SemanticReleaseError: The push permission to the Git repository is required.
2018-04-10T12:00:43.9903773Z     at module.exports (/vsts/agent/_work/1/s/node_modules/semantic-release/lib/get-error.js:6:10)
[...]

PR #742, opened 2018-04-10
Fix: issue#741

roger199585:
This is the way that I fix the issue

### DIFF
--- a/lib/get-git-auth-url.js
+++ b/lib/get-git-auth-url.js
@@ -1,8 +1,8 @@
-const {parse, format} = require('url');
-const {isUndefined} = require('lodash');
+const { parse, format } = require('url');
+const { isUndefined } = require('lodash');
 const gitUrlParse = require('git-url-parse');
 const hostedGitInfo = require('hosted-git-info');
-const {verifyAuth} = require('./git');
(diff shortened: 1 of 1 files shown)


pvdlg:
Hardcoding a specific behavior based on the presence of a specific string of character in the URL is not the proper fix. This is too brittle (what if CSTS offer different or private domaine names?). In addition it would open the door to fixes for each specific case with more hardcoded conditions that make the maintenance a lot harder.

See https://github.com/semantic-release/semantic-release/issues/741#issuecomment-380355402 for a better solution.

In addition, when opening a PR, please:
- Do not modifying the formatting of unrelated code
- Format the code you change according to the project formatting
- Add unit tests for the case you add

I would invite you to read our [contribution guidelines](https://github.com/semantic-release/semantic-release/blob/caribou/CONTRIBUTING.md) in particular the [lint](https://github.com/semantic-release/semantic-release/blob/caribou/CONTRIBUTING.md#lint) and [tests](https://github.com/semantic-release/semantic-release/blob/caribou/CONTRIBUTING.md#tests) sections.

pvdlg:
No worries. We wrote the contributing guidelines to help new contributor.

Issue #747, opened 2018-04-18
Fails on NPM publish with valid token


schalkneethling:
## Current behavior

When running semantic release, a new tag is written to Github and the release is published but, the step to publish to NPM fails with

```
[Semantic release]: Publishing version 1.0.2 to npm registry
[Semantic release]: An error occurred while running semantic-release: { Error: Command failed: npm publish ./. --registry https://registry.npmjs.org/
npm ERR! publish Failed PUT 403
npm ERR! code E403
npm ERR! You do not have permission to publish "bob". Are you logged in as the correct user? : bob
npm ERR! A complete log of this run can be found in:
npm ERR!     /home/travis/.npm/_logs/2018-04-18T08_25_57_953Z-debug.log
```

I am using tokens for both Github and NPM configured as environmental variables in Travis-CI.

One question though, does the last line of the error suggest that the user being used for NPM is bob? That is the repo name so, if that is the case, I am not sure how it is inferring that from the token.

Thank you in advance for your assistance.

## Expected behavior

It should publish to Github and NPM successfully.

## Environment

- **semantic-release** version: ^15.1.5
- CI environment: Travis-CI
- Plugins used: Just the defaults
- **semantic-release** configuration: default
- CI logs: https://travis-ci.org/mdn/bob/builds/368034044

pvdlg:
This is not a semantic-release issue. It's related to npm.
See https://github.com/npm/npm/issues/8561

It's probably because the package name you are trying to publish has already been used by someone else.

The npm error message is not really explicit though... It would be a good idea to add that in [troubleshooting](https://github.com/semantic-release/semantic-release/blob/caribou/docs/support/troubleshooting.md)

boennemann:
> It's probably because the package name you are trying to publish has already been used by someone else.

Note that also publishes fail if the names are too similar. E.g. if bob.js exists, you can't publish bobjs

schalkneethling:
Thanks so much for the feedback. Change the name I must ;) Well slightly, but not to slightly.

schalkneethling:
And it works! Thanks so much for the awesome tool and the exceptional support! https://www.npmjs.com/package/mdn-bob

PR #748, opened 2018-04-19
docs: add details on NPM error 403 to troubleshooting doc

schalkneethling:
Adds clarifying information with regards to NPM error 403 as discussed in this issue: https://github.com/semantic-release/semantic-release/issues/747

### DIFF
--- a/docs/support/troubleshooting.md
+++ b/docs/support/troubleshooting.md
@@ -29,3 +29,18 @@ You can recover from that issue by identifying the commit in your branch history
 $ git tag -f v<version of the last release> <commit sha1 corresponding to last release>
 $ git push -f --tags origin <your release branch>
 ```
+
+## You do not have permission to publish 'package-name'
+
+When running semantic-release you may encounter the following error:
+
+```
+An error occurred while running semantic-release: { Error: Command failed: npm publish ./. --registry https://registry.npmjs.org/
+npm ERR! publish Failed PUT 403
+npm ERR! code E403
+npm ERR! You do not have permission to publish "<package-name>". Are you logged in as the correct user? : <package-name>
+```
+
+This message is a little unclear, and might not have anything to with your `NPM_TOKEN` or authentication method. It might instead be related to the package name itself. If there is already a package with the same name as yours or, there is a very close match, it could trigger this error.
+
+Best way to be sure, is to search [npmjs.org](https://www.npmjs.com/)) using your package name. If there is a name conflict, rename your package in your `package.json`


docs/support/troubleshooting.md
```diff
@@ -5,8 +5,9 @@
 This error happens when the commit associated with the last release cannot be found in the branch history.
 
 Multiple situation can cause this issue:
-- The package name configured in your `package.json` already exists and **semantic-release** obtains the last release of that package, which is unrelated to yours
```
pvdlg: Can remove this change of syntax that is unrelated to this PR?

docs/support/troubleshooting.md
```diff
@@ -29,3 +30,18 @@ You can recover from that issue by identifying the commit in your branch history
...
+
+## You do not have permission to publish 'package-name'
+
+When running semantic-release you may encounter the error below:
+
+```
+An error occurred while running semantic-release: { Error: Command failed: npm publish ./. --registry https://registry.npmjs.org/
+npm ERR! publish Failed PUT 403
+npm ERR! code E403
+npm ERR! You do not have permission to publish "package". Are you logged in as the correct user? : package
```
pvdlg: Maybe use `package-name` in the error message instead of `package`? That would be clearer.

docs/support/troubleshooting.md
```diff
@@ -29,3 +30,18 @@ You can recover from that issue by identifying the commit in your branch history
...
+```
+
+This has in fact nothing to do with your `NPM_TOKEN` or authentication method. It is actually related to this NPM issue: [https://github.com/npm/npm/issues/8561](https://github.com/npm/npm/issues/8561) and [Improve message thrown from Error code 'E403'](https://github.com/npm/registry/issues/240). The problem is that there is already a package with the same name as the one you are trying to publish. The resolution therefore is to rename your package in your `package.json`
```
pvdlg: I wouldn't be that categorical here. There is situation in which this error message is actually correct and caused by a permission. So I would soften the language here.

pvdlg: No need to reference to `https://github.com/npm/registry/issues/240` here.
Just explain the error might be due to the package name (or a very similar one) is already used.

docs/support/troubleshooting.md
```diff
@@ -29,3 +30,18 @@ You can recover from that issue by identifying the commit in your branch history
...
+
+NOTE: If your package is called for example bob.js and there is already a package with that name. Changing it to bobjs will not resolve the problem as NPM will also reject names that are too similar.
```
pvdlg: No need for the note here. Also it's difficult to know what npm actually consider similar or not.
Just mention in the previous verbiage something like "the package name, or a very similar one, is already taken".

docs/support/troubleshooting.md
schalkneethling: Oh, looks like my editor made a formatting change i.e. changed `-` to `*` So, the copy is not actually removed.

pvdlg: I don't understand what you mean... Please remove your modification so `-` stays `-`.

schalkneethling: I will revert. More specifically, I believe it might be prettier-atom that did the reformatting.

schalkneethling:
@pvdlg Updated, thanks for the review.

docs/support/troubleshooting.md
```diff
@@ -29,3 +30,18 @@ You can recover from that issue by identifying the commit in your branch history
...
+```
+An error occurred while running semantic-release: { Error: Command failed: npm publish ./. --registry https://registry.npmjs.org/
+npm ERR! publish Failed PUT 403
+npm ERR! code E403
+npm ERR! You do not have permission to publish "<package-name>". Are you logged in as the correct user? : <package-name>
+```
+
+This message is a little unclear, and might not have anything to with your `NPM_TOKEN` or authentication method. It might instead be related to the package name itself. If there is already a package with the same name as yours or, there is a very close match, it could trigger this error.
+
+Best way to be sure, is to search [npmjs.org](npmjs.org) using your package name. If there is a name conflict, rename your package in your `package.json`
```
gr2m: I think `[npmjs.org](npmjs.org)` should be `[npmjs.org](https://www.npmjs.com/)`, otherwise this is good to go, thanks :)

docs/support/troubleshooting.md
```diff
@@ -5,6 +5,7 @@
 This error happens when the commit associated with the last release cannot be found in the branch history.
 
 Multiple situation can cause this issue:
+
```
pvdlg: Sorry to nitpick, but can you also remove this additional new line as it's not related to your addition?

schalkneethling:
Thanks for the reviews everyone. This has been updated.

pvdlg:
Thanks!

PR #759, opened 2018-04-29
fix: verify the local branch is up to date with the remote one

### LINKED ISSUE #709: release fails because "git push --dry-run" fails with  "HEAD -> master (non-fast-forward)"
## Current behavior

`semantic-release` is checking if it has write access with 

```
git push --dry-run https://[secure]@github.com/octokit/routes.git HEAD:master
```

It currently fails the release if the server rejects because it’s fast forward

```
! [rejected]        HEAD -> master (non-fast-forward)
error: failed to push some refs to 'https://[secure]@github.com/octokit/routes.git'
hint: Updates were rejected because a pushed branch tip is behind its remote
hint: counterpart. Check out this branch and integrate the remote changes
hint: (e.g. 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
````

## Expected behavior

I think if we can take this error as a success, as we wouldn’t get it if we wouldn’t have write access, I assume?

## Environment

- **semantic-release** version: 15.1.2
- CI environment: Travis
- CI logs: https://travis-ci.org/octokit/routes/jobs/355627978#L2104

### PR DESCRIPTION
**Pull request by pvdlg**
Fix #709

### DIFF
--- a/lib/git.js
+++ b/lib/git.js
@@ -131,6 +131,17 @@ async function verifyTagName(tagName) {
   }
 }
 
+/**
+ * Verify the local branch is up to date with the remote one.
+ *
+ * @param {String} branch The repository branch for which to verify status.
+ *
+ * @return {Boolean} `true` is the HEAD of the current local branch is the same as the HEAD of the remote branch, `false` otherwise.
+ */
+async function isBranchUpToDate(branch) {
+  return isRefInHistory(await execa.stdout('git', ['rev-parse', `${branch}@{u}`]));
+}
+
 module.exports = {
   gitTagHead,
   gitTags,
@@ -143,4 +154,5 @@ module.exports = {
   tag,
   push,
   verifyTagName,
+  isBranchUpToDate,
 };
--- a/lib/git.js
+++ b/lib/git.js
@@ -45,7 +45,7 @@ async function isRefInHistory(ref) {
 /**
  * Unshallow the git repository (retriving every commits and tags).
  *
- * @param {String} repositoryUrl The remote repository URL with credentials.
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: lib/git.js [hunk 1]
```diff
@@ -129,6 +129,17 @@ async function verifyTagName(tagName) {
   }
 }
 
+/**
+ * Verify the local branch is up to date with the remote one.
+ *
+ * @param {String} branch The repositoru branch for which to verify status.
```
gr2m: /s/repositoru/repository

test/helpers/git-utils.js
```diff
@@ -212,6 +209,15 @@ export async function gitCommitTag(gitHead) {
  * @param {String} branch The branch to push.
  * @throws {Error} if the push failed.
  */
-export async function push(origin, branch) {
+export async function push(origin = 'origin', branch = 'master') {
```
gr2m: can you please explain why you set defaults now? Could this be  problem if `master` is not the default branch on remote?

And unrelated to this PR: should the `origin` argument be called `remote`? `"origin"` is the default remote name, or does the parameter refer to something else here?

pvdlg: This code is not used in semantic-release. It's just code used in the test.
I set the defaults out of convenience.

Yes I can change the arg name to `remote`

pvdlg: Well actually in each other function on this util file similar args are named `origin`. Should I change it to `remote` everywhere? Or maybe `repositoryUrl` would be clearer?

gr2m: `repositoryUrl` is more clear I think 👍

PR #763, opened 2018-04-30
unshallow with credentials

botcheddevil:


pvdlg:
Can you fix the test?

I think you have to pass the remote URL to `unshallow()` in the test.
You might have to change some tests in `index.test.js` to make sure they use a cloned repo.

PR #781, opened 2018-05-07
docs(recipe): update gitlab usage configuration

cdaringe:
# problem

- it wasn't clear to me how to setup for usage with GitLab

# solution

- fill the missing documentation gap in the GitLab recipe


pvdlg:
This recipes is part of CI recipes that covers the CI configuration only, not specifics related to a particular Git host. Note it's named `GitLab CI` and not `GitLab`.
So information related to GitLab itself (the Git host) shouldn't be in that recipe.

All the information you added are already in the documentation of the GitLab shareable config: https://github.com/semantic-release/gitlab-config. If something is unclear or missing it should be added over there.

The shareable config is also mentioned here: https://github.com/semantic-release/semantic-release/blob/caribou/docs/extending/shareable-configurations-list.md

pvdlg:
Closing per previous comment and due to no response.
Please re-open if you disagree or want to discuss/refine a possible doc improvement.

Issue #782, opened 2018-05-07
SSH Guide should be more detailed


ev1stensberg:
## New feature motivation

When following along the SSH key guide with github + travis, I did everything as expected, but it didn't work. In the guide it says "one of these is needed to push", so I assumed that a SSH key alone would make it work.

What defaults can I use, what defaults work out of the box? From intuition, I thought that if SSH keys were present, without the GH_TOKEN, then the release would trigger, but it didn't. 

## New feature description

A more detailed documentation on pushing with SSH keys and what is needed for SSH keys to work. 

( A small paragraph on what keys you need with SSH ) ( i.e GH_TOKEN + SSH keys)

## New feature implementation

N/A

pvdlg:
Fixed in #783

PR #783, opened 2018-05-07
docs: add comment about publishing with ssh keys

ev1stensberg:
As stated in #782


pvdlg:
The note added is not accurate, as the SSH keys are sufficient for the base functionalities of semantic-release core. An API is required only when using a plugin that requires it and would be mentioned in the plugin doc.

Could you replace the note with:

```md
**Note:** SSH keys allow to push the [Git tag](https://git-scm.com/book/en/v2/Git-Basics-Tagging) associated to the released version. Some plugins might also require an API token. See each plugin documentation for additional information.
```

ev1stensberg:
gotcha, updated now

ev1stensberg:
There we go 👍

pvdlg:
Thanks

PR #788, opened 2018-05-09
fix: use `git rev-parse origin/${branch}` to verify origin head

### LINKED ISSUE #786: Error when no upstream is configured for active branch
## Current behavior

Using semantic-release from Jenkins I'm experiencing the next error because Jenkins doesn't add an upstream branch for the checked out branch:

```
[Semantic release]: An error occurred while running semantic-release: { Error: Command failed: git rev-parse master@{u}
fatal: no upstream configured for branch 'master'
```

The problem is introduced in d15905c0d5acaa0847173cc3588deb40d2848301

## Expected behaviour

Skip verifying if the local branch is up to date with the remote one when there is no upstream configured. This way new check will still be useful but without breaking scenarios where upstream is not configured.

## Environment

- **semantic-release** version: 15.3.1
- CI environment: Jenkins

pvdlg:
Fix #786.

Seems to fix the issue according to https://github.com/semantic-release/semantic-release/issues/786#issuecomment-387830848

### DIFF
--- a/lib/git.js
+++ b/lib/git.js
@@ -143,10 +143,10 @@ async function verifyTagName(tagName) {
  *
  * @param {String} branch The repository branch for which to verify status.
  *
- * @return {Boolean} `true` is the HEAD of the current local branch is the same as the HEAD of the remote branch, `false` otherwise.
+ * @return {Boolean} `true` is the HEAD of the current local branch is the same as the HEAD of the remote branch, falsy otherwise.
  */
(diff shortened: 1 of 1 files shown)


felixfbecker:
What if `origin` is not the upstream remote (or there is no remote named `origin`)?

gr2m:
It’s a problem when it’s a problem :) I’d suggest if someone runs into it in a production setting then we will address it

PR #794, opened 2018-05-18
docs: Add details about how semantic-release handles merges.

jwalton:
Especially how it handles squash merges.  See https://github.com/semantic-release/commit-analyzer/issues/65.

### DIFF
--- a/docs/support/FAQ.md
+++ b/docs/support/FAQ.md
@@ -1,5 +1,25 @@
 # Frequently Asked Questions
 
+## How does semantic-release handle merges?
+
+First, a primer on git commits.  Every commit in git has a "parent".  If you create a new git repo and make a commit `A`, since this is the first commit it will not have a parent.  If you make a second commit `B`, this commit will have `A` as the parent.  When you run `git log`, it starts at the commit you're currently on, and walks up the parents to show you the history of your branch.
+
+A "merge commit" is a special kind of commit with two parents.  To continue our example, if you create a new branch and commit `C` and `D`, then merge it into the master, master will now have five commits - `A` and `B` from master, `C` and `D` from the branch, and the merge commit itself.
+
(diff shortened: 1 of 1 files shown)


pvdlg:
The `How does semantic-release handle merges?` is really generic and unrelated to semantic-release.
semantic-release simply analyze the commits present on the release branch when it runs. It doesn't matter if those commits have been push, rebased or merged from another branch. When you merge and a Merge commit is created, semantic-release just analyze the merged (note the **d** in merge**d**) and ignore the Merge commit.
So all merge consideration are not relevant for how semantic-release works so I would remove this section entirely.

The `How does semantic-release handle squash merges?` section is not completely accurate and is not clear due to explanation about breaking change. The fact that breaking change are parsed is just a side-effect of an invalid commit message. See https://github.com/semantic-release/commit-analyzer/issues/65#issuecomment-390505020

I would replace this entire section by this paragraph in the [Troubleshooting](https://github.com/semantic-release/semantic-release/blob/caribou/docs/support/troubleshooting.md) doc instead:
```md
## Squashed commits are ignored by semantic-release

**semantic-release** parses commits according to a [commit message convention](https://github.com/semantic-release/semantic-release#commit-message-format) to figure out how they affect the codebase. Commits that doesn't follow the project's commit message convention are simply ignored.

When [squashing commits](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#_squashing) most Git tools will by default generate a new commit message with a summary of the squashed commits. This commit message will most likely not be compliant with the project's commit message convention and therefore will be ignored by **semantic-release**.

When squashing commits make sure to rewrite resulting commit message to be compliant with the project's commit message convention.

**Note**: if the resulting squashed commit would encompasses multiple changes (for example multiple unrelated features or fixes) then it's probably not a good idea to squash those commits together. A commit should contain exactly one self-contained functional change and a functional change should be contained in exactly one commit. See [atomic commits](https://en.wikipedia.org/wiki/Atomic_commit).
```

PR #797, opened 2018-05-20
docs: clarify which commit types trigger a release

### LINKED ISSUE #787: Unclear what "relevant code changes" means
The [Readme file](https://github.com/semantic-release/semantic-release/blob/caribou/README.md) mentions that:

> ... runs the `semantic-release` command to make a release if there are **relevant codebase changes** since the last release.

and:

> By default a release will be done for each push to the release branch that contains **relevant code changes**.

But it is never explained what "relevant code changes" means exactly. After reading the [FAQ](https://github.com/semantic-release/semantic-release/blob/caribou/docs/support/FAQ.md#is-it-really-a-good-idea-to-release-on-every-push) it seems to hint that a release will happen with every single push to the release branch, and not only when there are "relevant code changes". Of course this doesn't apply if the commit has a `[skip release]` in the message.

Given these somewhat contradictory statements I'm still not sure when a release is going to happen. Of course I'm a new user, otherwise I would probably know the answer through trial and error, but I'd prefer to be sure instead of having to go through that.

pvdlg:
Fix #787

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,9 +68,9 @@ Here is an example of the release type that will be done based on a commit messa
 
 ### Triggering a release
 
-When pushing new commits to the release branch (i.e. `master`) with `git push` or by merging a pull request or merging from another branch, a CI build is triggered and runs the `semantic-release` command to make a release if there are relevant codebase changes since the last release.
+For each new commits added to the release branch (i.e. `master`) with `git push` or by merging a pull request or merging from another branch, a CI build is triggered and runs the `semantic-release` command to make a release if there are codebase changes since the last release that that affect the package functionalities.
 
-By default a release will be done for each push to the release branch that contains relevant code changes. If you need more control over the timing of releases you have a couple of options:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -68,9 +68,9 @@ Here is an example of the release type that will be done based on a commit messa
 
 ### Triggering a release
 
-When pushing new commits to the release branch (i.e. `master`) with `git push` or by merging a pull request or merging from another branch, a CI build is triggered and runs the `semantic-release` command to make a release if there are relevant codebase changes since the last release.
+For each new commits added to the release branch (i.e. `master`) with `git push` or by merging a pull request or merging from another branch, a CI build is triggered and runs the `semantic-release` command to make a release if there are codebase changes since the last release that that affect the package functionalities.
```
ricardograca: Typo: "...that that affect..."

pvdlg: Thanks. See https://github.com/semantic-release/semantic-release/pull/798

PR #805, opened 2018-05-30
docs(plugins): remove version warning about semantic-release-docker

felixfbecker:
semantic-release-docker is compatible with latest semantic-release now,
see peerDependency: https://david-dm.org/felixfbecker/semantic-release-docker?type=peer


gr2m:
Thanks!

PR #822, opened 2018-06-16
fix: prevent exception during dryRun if generateNotes is disabled

sirlantis:
This was failing with `Error: marked(): input parameter is undefined or null` when setting `generateNotes` to `[]`.


sirlantis:
I hope my use-case of disabling changelog generation is not esoteric.

Should I test for `== null` instead?

pvdlg:
Can you add a test to keep 100% coverage? Thanks.

PR #823, opened 2018-06-16
Add SemVer compatibility badge

greysteil:
Would you be up for adding a badge that shows how SemVer compliant / bug free new releases are? I was looking through the data we gather at Dependabot and realised we could put one together, so threw together the below:

[SemVer](https://dependabot.com/compatibility-score.html?dependency-name=semantic-release&package-manager=npm_and_yarn&version-scheme=semver)

Would love your opinion - is it useful? Anything you'd change about it (e.g., keep / lose the logo)? If you click it there's an explanation of how it's calculated. I guess for `semantic-release` there's no reason why this dependency would ever break tests, so your 100% score is pretty much guaranteed.

Any feedback super appreciated!

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,9 @@
   <a href="https://greenkeeper.io">
     <img alt="Greenkeeper" src="https://badges.greenkeeper.io/semantic-release/semantic-release.svg">
   </a>
+  <a href="https://dependabot.com/compatibility-score.html?dependency-name=semantic-release&package-manager=npm_and_yarn&version-scheme=semver">
+    <img alt="semver-compatibility" src="https://api.dependabot.com/badges/compatibility_score?dependency-name=semantic-release&package-manager=npm_and_yarn&version-scheme=semver">
+  </a>
   <a href="#badge">
     <img alt="semantic-release" src="https://img.shields.io/badge/%20%20%F0%9F%93%A6%F0%9F%9A%80-semantic--release-e10079.svg">
   </a>


pvdlg:
Thanks for the proposal.
We decided a while ago to keep the badges on the readme to the minimum, so we are not going to add a new one as suggested in this PR.

I checked quickly the page associated with this badge. I might have missed something, but it's not clear for me what "semver compatibility" means. Does it check the released version are named with a semver tag? If it's the case it's not really useful as npm forces that anyway.

greysteil:
👍 - if you're not keen on badges then that makes perfect sense.

On the confusion on the page, I'd love to get to the bottom of that and change it to be clearer. Was it this line that was confusing?

> Dependabot has performed 245 updates of semantic-release between SemVer compatible versions, and 100% of those updates passed CI.

By SemVer compatible versions I'm trying to refer to changes like 2.1.0 -> 2.2.0, rather than incompatible version updates like 2.2.0 -> 3.0.0. Was it that text that confused you or something else?

pvdlg:
This sentence doesn't make much sense to me. How `2.1.0 -> 2.2.0` is valid and `2.2.0 -> 3.0.0` not? 
It's perfectly ok to release a version `3.0.0` if the last release is `2.2.0` and the new change is a breaking change.
I really don't understand what you are measuring.

greysteil:
Ah, OK - so it's measuring the percentage of users' test runs that pass when upgrading between versions that, according to SemVer, shouldn't break anything. So that would include `2.1.0 -> 2.2.0` but not necessarily `2.2.0 -> 3.0.0`. Does that make sense now? I'll try to make it clearer on the page.

pvdlg:
Ok it's clearer now.

I'm not sure you'll a reliable number though, because you are going to get a lot of false positive in cases where the CI just fail for unrelated reason, the user tests are flappy, the user tests fails for unrelated reason etc...

So someone maintaining a library and displaying your badge might end up showing something that is not 100% even though their releases perfectly respect semver.
And the score will actually depends on the quality of tests written by dependents libraries that also use  Dependabot. So if I'm unlucky and the few libraries using mine and having Dependabot configured have flappy tests then I'll get a bad score even though I fully respect semver.

greysteil:
There's certainly a problem with flaky tests, but from my experience it's not that big a one. I'm filtering out all of the projects whose tests were initially failing, and in future it would be relatively straightforward for me to filter out flaky tests, too (since Dependabot does updates across lots of dependencies for a given user we can identify whether their tests are flaky).

Even without that optimisation we see really high scores for most packages that use SemVer well - check out:
- [jest](https://dependabot.com/compatibility-score.html?dependency-name=jest&package-manager=npm_and_yarn&version-scheme=semver)
- [eslint](https://dependabot.com/compatibility-score.html?dependency-name=eslint&package-manager=npm_and_yarn&version-scheme=semver)

For others, where a breaking change or bug was accidentally introduced in a minor version, the scores are lower:
- [prettier](https://dependabot.com/compatibility-score.html?dependency-name=prettier&package-manager=npm_and_yarn&version-scheme=semver) (click ... see more to see the score for `1.12.1 -> 1.13.0`)

Anyway, thanks for your feedback!

pvdlg:
Ah ok that make sense.
I think that's an interesting feature indeed! I imagine the more people use Dependabot the more accurate the results will be.

PR #825, opened 2018-06-18
Fix documented explanation in plugins.md

richardson-trevor:
`verifyConditions` explanation was wrong previously.


pvdlg:
Thanks!

PR #859, opened 2018-07-16
docs: add new semantic-release-gcr plugin link

carlos-cubas:
Update docs to add a new community plugin semantic-release-gcr. This plugin publishes a docker image to google cloud registry.


pvdlg:
Thanks for the PR!
Looking at the code I see that you are using a [a bash script](https://github.com/carlos-cubas/semantic-release-gcr/blob/master/scripts/gcloudlogin.sh). Because of that any CI based on windows will not be able to run your plugin.

Do you think you can change that to make sure your plugin would work for everyone?

Thanks

carlos-cubas:
I roughly have an idea of how to change it, but I don't have a windows machine to test it on.  in the meantime i'll update the readme to explain support only for *nix

pvdlg:
For info, [Appveyor](https://www.appveyor.com/) offer free Windows CI environment for OSS.

carlos-cubas:
Hi Pierre, 

I ended up updating the docs with a no windows support disclaimer.  I unfortunately don't have the time to do this now. 

So i have created an issue on the repo to track it: https://github.com/carlos-cubas/semantic-release-gcr/issues/2

In general I believe there's some value, for a large portion of the community, to use the plugin in its current state. So i would encourage a merge of the PR and addition to the docs.

But it is your call.

PR #879, opened 2018-08-01
Unnecessary use of package url

JureSotosek:
While looking trough the code I noticed that there were two different packages doing the same job, parsing the urls, which seamed unnecessary. So i removed one of the packages and moved all the function to the other one, optimising the necessary number of packages to be imported.

Also, the only other place where the removed package is used is in `lib/definitions/errors.js` , if you want, I can also replace the use of the package there with the other package('git-url-parse) that is now used in the files and changes I made and remove the package('url) completely.

### DIFF
--- a/lib/get-git-auth-url.js
+++ b/lib/get-git-auth-url.js
@@ -1,4 +1,3 @@
-const {parse, format} = require('url');
 const {isUndefined} = require('lodash');
 const gitUrlParse = require('git-url-parse');
 const hostedGitInfo = require('hosted-git-info');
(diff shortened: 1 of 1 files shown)


pvdlg:
This is done like that on purpose.
In that code, we parse the URL, add the basic auth part and format it. `git-url-parse` can achieve that but it also does a bunch of other stuff that we don't necessary control. For example, when reformatting the URL from the parsed object, it used to always remove `.git` at the end of the URL.
In other terms `gitUrlParse(repositoryUrl).toString() === repositoryUrl` doesn't always return `true`.

Using `url.parse` and `url.format` guarantees that not transformation will be done in the process or parsing/formatting. So the resulting URL will be identical to the original one, with the basic auth added.

Also the `url` package is part of Node (it's not a `package.json` dependency) so using it doesn't have any impact.

Issue #827, opened 2018-06-19
Proposal: use shareable configs instead of default plugins


pvdlg:
Currently we set the `@semantic-release/npm` and `@semantic-release/github` plugins by default.
Those are good defaults but it gets quite complex and counter-intuitive for users who do not want to use one of those plugin, as their configuration will be merged with the default one.

For example, if one do not want to use the `@semantic-release/github` they would have to:
- Check the `@semantic-release/github` doc to figure out on which step it is used
- For each steps in which only `@semantic-release/github` is used, override the config with `false`
- For each steps in which `@semantic-release/github` is used with another plugin, override with the other plugins only

In that example, the config would be:
```json
{
  "verifyConditions": ["@semantic-release/npm"],
  "publish": "@semantic-release/npm",
  "fail": false,
  "success": false
}
```

To make things more simple we could remove all the default plugins and create specific [shareable configurations](https://github.com/semantic-release/semantic-release/blob/caribou/docs/usage/shareable-configurations.md). For example:
- `@semantic-release/npm-github-config` (identical to current defaults)
- `@semantic-release/npm-gitlab-config` (release on npm, use GitLab repo)
- `@semantic-release/npm-git-config` (release on npm, push `package.json` and `CHANGELOG.md` to the repo)
- `@semantic-release/github-config` (release on GitHub only)
- `@semantic-release/gitlab-config` (release on GitLab only)
- `@semantic-release/git-config` (push `package.json` and `CHANGELOG.md` to the repo. Would work for Bitbucket and custom Git servers)
- `@semantic-release/npm-config` (release on npm only)

By default, if the user provide no value for `extends` we would use `@semantic-release/npm-github-config` which is equivalent to the current default. So that would be transparent for users currently relying on the default config.

The config for the previous example (using npm but not github) would simply be:
```json
{
  "extends": ["@semantic-release/npm-config"],
}
```

For users who want something more custom and need to configure each hook, things would be simpler as they wouldn't need to care about overriding each step default value, they could just set what they need.

pvdlg:
//cc @gr2m

gr2m:
sounds good to me! I’m a bit concerned about all the new packages, it would be good to find other folks take ownership over them?

pvdlg:
It's always good to get more people taking ownership. However it's really simple packages as they contains only a few lines of JSON.

pvdlg:
[...]

PR #936, opened 2018-10-01
Simplify plugins configuration

pvdlg:
The plugins configuration is currently quite complex and counter intuitive. That generate a lot of support burden on the issue tracker or the Gitter channel, create some frustration for users and probably discourage some potential new users who try out semantic-release for the first time.

Here is a list of friction points and difficulties I encounters helping other to configure semantic-release:
1. Using plugins require to understand the concept of [plugin step/hooks](https://github.com/semantic-release/semantic-release#release-steps)
2. Using plugins require to understand which step/hooks are implemented by the plugins one want to use
3. Using plugins require to understand which step/hooks are implemented by the default plugins in order to overwrite them
4. Plugins options have to be passed to each step/hooks implemented by the plugin (for example with `@semantic-release/npm`, `pkgRoot` has to be defined 3 times: in `verifyConditions`, `prepare` and `publish` which makes no sense for a user)
5. Passing options to a plugin is confusing. It's currently done by using an object with a `path` prop along with the desired options. The word `path` seems to refer to a file system path, however a module name is expected. In addition mixing the plugin name and the options in the same objet is an uncommon practice.
6. Some plugins have to be installed while others don't
7. A breaking change in one of the default plugin introduce a breaking change in semantic-release as a whole for some users

This PR aims at solving most of those issues by introducing a new way to configure plugins, while **preserving complete backward compatibility** with the current configuration method.

This is a better fix for #827.

### New `plugins` option
The new recommended way to configure plugins is to use the `plugins` option which is an Array of plugin names. For example:
```json
 "plugins": ["@semantic-release/commit-analyzer", "@semantic-release/release-notes-generator", "@semantic-release/npm"]
```
No need to deal with the concept of step/hooks anymore, semantic-release will run each step implemented by each plugin defined, in the order of definition.

That should solves issues 1, 2 and 3.

**Note**: Only plugins returning an Object with a function for each step are supported in the `plugins` options. Old plugin format that just return a function are still supported when used with the current step/hook configuration.

**Note**: The current way of defining plugins (with the options `verifyConditions`, `analyzeCommits`, `publish` etc...) is still supported but mentioned in the documentation anymore.

### Defining options by wrapping by wrapping the name and an options object in an array

In order to pass options to a plugin, it's name can be wrapped in an Array along with the options to pass. This is a practice implemented by several well known project such as [Babel plugins](https://babeljs.io/docs/en/plugins#plugin-options) or [eslint rules](https://eslint.org/docs/user-guide/configuring#configuring-rules).

Global plugins options are still supported and can be set at the root of the semantic-release config and will be passed to all plugins.

```json
{
  "plugins": [
    "@semantic-release/release-notes-generator",
    ["@semantic-release/github", {
      "assets": ["dist/**"]
      }],
    "@semantic-release/git"
  ],
  "preset": "angular"
}
```
In this example `assets` will be passed to each steps of `@semantic-release/github` but not to `@semantic-release/git`. `preset` will be passed to both `@semantic-release/commit-analyzer` and `@semantic-release/release-notes-generator`.

That should solves issues 4 and 5.

### Documentation improvements

Plugins must be installed, except for the default ones (`@semantic-release/commit-analyzer`, `@semantic-release/release-notes-generator`, `@semantic-release/npm`, `@semantic-release/github`) which might be confusing. For backward compatibility reason we have to continue to use default plugins and to embed them.
However, the main documentation and each plugin README now mention to install the plugins.
Ideally users should install all plugins, so they can choose the version used and they would be aware of a breaking change in a default plugin.

That should solves issues 7 and 8.

TODO:
- [ ] Update `@semantic-release/commit-analyzer` dependency once new plugin version released
- [ ] Update `@semantic-release/release-notes-generator` dependency once new plugin version released


felixfbecker:
Sometimes I like to compose plugins for different steps in config, e.g. run only `@semantic-release/npm` for `prepare` to update a package.json, but then publish the package in a different way with `@semantic-release/exec`. Or not run the `fail` and `success` plugins of `@semantic-release/github`. How would that be supported in a world with the `plugins` option?

pvdlg:
That's still supported. This change is fully backward compatible with existing configurations.

By the way for the cases you mention you also have the following options:
- Set the `npmPublish` option to `false` to disable publishing
- Set `successComment`/`successTitle` and/or `failTitle` to `false` (See semantic-release/github#111)

pvdlg:
As it's a pretty big feature that involve some plugins changes, we are going to keep this feature (and the associated plugin PR) on `@next` dist-tag for a couple weeks. I would rather avoid to keep it there longer as we have a discrepancy between the docs and what's on `@latest`.

That'd be great if users could test that new version on their projects, in particular:
- Making sure existing configurations still works
- Making sure the new `plugins` option works as expected

PR #944, opened 2018-10-05
feat: added --no-tag configuration option

anru:
Sometimes I don't need to add and push git tag directly, but I want do it through plugins instead.
So I added option for such cases

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -32,6 +32,7 @@ Usage:
     .option('publish', {...stringList, group: 'Plugins'})
     .option('success', {...stringList, group: 'Plugins'})
     .option('fail', {...stringList, group: 'Plugins'})
+    .option('no-tag', {describe: 'Skip creating git tag for repository', type: 'boolean', group: 'Options'})
     .option('debug', {describe: 'Output debugging information', type: 'boolean', group: 'Options'})
     .option('d', {alias: 'dry-run', describe: 'Skip publishing', type: 'boolean', group: 'Options'})
     .option('h', {alias: 'help', group: 'Options'})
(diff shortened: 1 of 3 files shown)


pvdlg:
semantic-release cannot work without push tags because on each run it will check the previous tags to figure what is the last release and what are the new commits.
So it cannot be optional.

anru:
> semantic-release cannot work without pushing tags

right, but where is restriction that forbids third-party plugins to push tags ?

pvdlg:
btw, what is the use case? Why would you not want to create a tag?

anru:
In my case runners doesn't have access to git repository, and only way to push tags is using gitlab api

Diokuz:
Probably, the better solution would be an option: how to push a tag – via `git push` or via `api`. But no-push-access-for-runners is definitely the case.

pvdlg:
> In my case runners doesn't have access to git repository, and only way to push tags is using gitlab api

It the runner doesn't have access to the git repo, how the repo was cloned on the CI in the first place? How semantic-release was able to read and analyze the commits?
And how is it possible to have access via API but not with git? As far as I can tell Gitlab expose the API and the git interface on the same host/port and the permissions are the same for both.

pvdlg:
Closing as no answer was provided. If you want to propose a feature please use the corresponding [issue template](https://github.com/semantic-release/semantic-release/issues/new?template=02_feature_request.md) and explain the in which circumstances the feature would be useful.
In that case include the answers to the questions asked in https://github.com/semantic-release/semantic-release/pull/944#issuecomment-427886267

fortesp:
> btw, what is the use case? Why would you not want to create a tag?

To do a QA testing snapshot release .. or kind of a pre-release... and the core could just validate the commits, calculate change the version and publish with no tag, because its not an official release.

PR #955, opened 2018-10-18
docs: add semantic-release-expo to plugins list

byCedric:
I created an [Expo](https://expo.io/) plugin for Semantic Release, see it [here](https://github.com/bycedric/semantic-release-expo). 😄


pvdlg:
Thanks!

PR #982, opened 2018-11-10
docs(plugins): clarify the `analyzeCommits` step

mdluo:
### Affected document page

https://semantic-release.gitbook.io/semantic-release/usage/plugins

### Motivation

In the following option case,

```json
{
  "plugins": [
    "@semantic-release/exec"
  ]
}
```

 The "analyzeCommits" step of plugin `commit-analyzer` will never be executed. This kind of behaviour is not covered by the document, while the explanation can only be found in the [test case](https://github.com/semantic-release/semantic-release/blob/caribou/test/plugins/plugins.test.js#L108-L126)


pvdlg:
Thanks for the PR. Instead I opened #983 which allow multiple definitions for `analyzeCommits` plugins. That will make things clearer and more consistent across plugins.

PR #997, opened 2018-11-23
Add maven-semantic-release to list of community plugins

evansiroky:
This change adds [maven-semantic-release](https://github.com/conveyal/maven-semantic-release) to the list of community plugins.


pvdlg:
Can you remove the "warning" section from there? That should be in the Readme of your plugin.
Also it's not accurate as now the preferred configuration methods is to use the [`plugins` option](https://github.com/semantic-release/semantic-release/blob/caribou/docs/usage/plugins.md).

I quickly checked your plugin code and there is several issues I found:
- Hardcoded dependency to GitHub [there](https://github.com/conveyal/maven-semantic-release/blob/390da6825193a722bb6f2ee689045da160324e51/lib/maven.js#L40). Use the option `repositoryUrl` instead, which is an authenticated URL that semantic-release passed to you
- Hardcoded dependency to Travis [there](https://github.com/conveyal/maven-semantic-release/blob/390da6825193a722bb6f2ee689045da160324e51/lib/maven.js#L41). Use the `branch` option that semantic-release passed to you
- Instead of creating and pushing your commit, you should just update the `pom.xml` and let users use `https://github.com/semantic-release/git` to commit it
- What your are doing in [verifyRelease](https://github.com/conveyal/maven-semantic-release/blob/dev/lib/verify-release.js) should be done in `verifyConditions`. The point of `verifyRelease` is to determine if the release type determined based on commits is correct (e.g. if based on the commits semantic-release determine a `patch` verifies that there is no breaking change by running previous version test or checking the public API didn't change, etc...)
- It's quite surprising to change the `pom.xml` to a snapshot version in [publish.js#L18](https://github.com/conveyal/maven-semantic-release/blob/390da6825193a722bb6f2ee689045da160324e51/lib/publish.js#L18). You should not need to do that with automated version releases.
- You should not [set `user.email` and `user.name`](https://github.com/conveyal/maven-semantic-release/blob/390da6825193a722bb6f2ee689045da160324e51/lib/git.js#L25), [semantic-release does that already](https://github.com/semantic-release/semantic-release/blob/caribou/docs/usage/configuration.md#git-environment-variables)
- You should not [verify the current CI branch](https://github.com/conveyal/maven-semantic-release/blob/390da6825193a722bb6f2ee689045da160324e51/lib/git.js#L18), [semantic-release does that already](https://github.com/semantic-release/semantic-release/blob/2b082acc7306aa94f9b7ddb4c7460914109bd6ef/index.js#L47)
- You should not [checkout a branch](https://github.com/conveyal/maven-semantic-release/blob/390da6825193a722bb6f2ee689045da160324e51/lib/git.js#L23) as it can creates many issues on CIs for code that can run after semantic-release
- You should not [define semantic-release as a dependency](https://github.com/conveyal/maven-semantic-release/blob/390da6825193a722bb6f2ee689045da160324e51/package.json#L15) but as a `peerDependency`. See [example](https://github.com/semantic-release/git/blob/39c660e608f66b078197f35cdfac8b9157920444/package.json#L77).
- Your doc [Step 5](https://github.com/conveyal/maven-semantic-release#step-5--create-a-travisyml-file-thatll-run-maven-semantic-release-after-success) and [Step 6](https://github.com/conveyal/maven-semantic-release#step-6--add-a-github-token-to-travis) are not necessry as that's already part of semantic-release doc

Would you mind solving those issues before we add it to the list? Feel free to ask questions on on the [Gitter chat](https://gitter.im/semantic-release/semantic-release).

evansiroky:
Thanks for the tips.  I believe we will be able to address most of these comments.  However, the part about making commits and pushing them is an important part of our internal workflow.  These commits fulfill 2 purposes.  First, we need to make a commit with the release so that when the code is tarred up for a github release, the correct version number is included in the pom.xml and also the commit listed in the release contains these changes.  Second, we need to update to a snapshot version because we create these intermediate versions that we don't want to create a github release and maven central release for, but do use for internal development by uploading jar files to s3 that we can then use elsewhere.  Furthermore, we're about to add more code that would merge these commits back into our own development branch.  Perhaps we could make these extra commits an optional thing that could be enabled by setting a cli flag or something.  Would that suffice?

pvdlg:
I was just wondering about those 2 commits. I think the one done in `prepare` can be done with `@semantic-release/git`. For the one done after `publish` that can be done in your plugin. If it's really specific to your workflow it can be behind an option. It's it's a general practice that any user would use it can be the default behavior without option.

To clarify, it's perfectly fine to create a plugin very specific to your workflow. But in such case it doesn't make much sense to add it to the list of plugins in this repo. The goal of this list is to provide plugins that can be used by other. This is why it's important to not restrict those to one CI for example.

evansiroky:
I think it may be best to add in an option to make the commits optional so that this can still be listed as a community plugin so we can get more community feedback and collaboration.

Issue #1074, opened 2019-01-18
Release skipped on CircleCI?


simlu:
Unfortunately I'm having trouble publishing. My setup is as follows:

(1) Only one master branch on main repo and a forked repo. PRs should be merged from fork into master and then directly published.
(2) The CircleCI option "Build forked pull requests" is set to "off"
(3) The CircleCI build are now triggered once the PR is merged in. That part works great. However the environment variables [CIRCLE_PULL_REQUEST, CIRCLE_PR_NUMBER, etc](https://circleci.com/docs/2.0/env-vars/) are still set.

This now causes semantic-release to stall with the error message `This run was triggered by a pull request and therefore a new version won't be published.`

Any idea how I could fix this?

Thank you very much! Cheers, L~

PS: I'm pretty sure on TravisCI this behaviour is different.

pvdlg:
What does trigger the CircleCI build, pushing to the PR branch or merging to master? On which branch CircleCI is running?

If the job was triggered by the merge to master and the build actually run on master branch then `CIRCLE_PULL_REQUEST` and `CIRCLE_PR_NUMBER` should not be set. If they are that's an issue with CircleCI that you should report over there.

simlu:
The build is triggered by merging into master. I'll try circle ci and see what they have to say and get back here. Cheers!

simlu:
Got a response when describing the above behavior to them:

*[...] The merge is part of the PR so these variables are expected to be set. The `CIRCLE_PULL_REQUEST` is specific to a forked PR and `CIRCLE_PR_NUMBER` references the specific PR or one chosen at random if there are multiples associated. More on this can be found in the [built-in environment](https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables) variables documentation.*

So these are expected to be set. Do we need a flag `--allow-on-pr`? How should this best be solved?

Cheers, L~

pvdlg:
Last time I tried CircleCI those variable were set only on the merge build triggered by opening/updating the PR. After merging the PR in GitHub a regular build was triggered on master.

Can you provide a link to the problematic CircleCI job?

simlu:
Unfortunately the repo is private atm. I can probably set up a test case. But this might have to do with (2)?

simlu:
@pvdlg I've put a pr up for now: https://github.com/semantic-release/semantic-release/pull/1076

[...]

Issue #753, opened 2018-04-24
Add an option to output the version that would be published


BenoitAverty:
# Feature request

## New feature motivation
I am currently implementing automatic releases for my project, which is a web application. The project currently satisfies my needs as I can disable the npm plugins and replace it with gitlab plugins.

The problem is that I need to put the version of my project into the built application (via Webpack's DefinePlugin), but the version field of my package.json is always set to 0.0.0.

I would like to be able to use semantic-release to print the version that would be published and use it in webpack DefinePlugin.

Currently, the only way to do that would be to parse the output of a dry-run, but it is complicated and not stable.

## New feature description

I see two ways this could happen : 
 * The simple one is to provide a cli option to instruct semantic release to just print the version. For example `semantic-release --print-next-version`
 * Allow a more extensive programmatic usage of semantic-release, with a JS api that would (among other things) allow to retrieve the next version.

I could open a PR for the first solution if you'd like. The second one is a bigger task (probably several tasks)

gr2m:
I think you can run a custom script in the `prepublishOnly` script, see https://semantic-release.gitbooks.io/semantic-release/docs/support/FAQ.html#how-can-i-use-a-npm-build-script-that-requires-the-packagejsons-version- 

At this point the version in package.json is set. Your custom script can read it and write it into your build files

BenoitAverty:
I would but unfortunately I don't publish with npm. My build script is executed way before the semantic release.

pvdlg:
You can also use the [@semantic-release/exec](https://github.com/semantic-release/exec) to run a script at any step you'd like that can be passed the new version number. Something like that:
```json
{
  "release": {
    "prepare": [
      "@semantic-release/npm",
      {
        "path": "@semantic-release/exec",
        "cmd": "my-script ${nextRelease.version}",
      },
    ],
  }
}
```

With this config the npm plugin is going to update your `package.json` first, then `my-script` will be called with the new version as first argument.

BenoitAverty:
The problem with all these solutions is that I need to have two configurations : One that outputs the version or sets it somewhere, and one for the actual release in gitlab (tag, etc.) which doesn't seem to be possible easily.

For now I think i'll parse the output of a dry-run, but this is really brittle if the output changes.

pvdlg:
> The problem with all these solutions is that I need to have two configurations : One that outputs the version or sets it somewhere, and one for the actual release in gitlab (tag, etc.) which doesn't seem to be possible easily.

[...]

PR #1076, opened 2019-01-19
feat: added '--skip-tag', '--skip-pr-check' and `--branch=*`

simlu:
This pr is not final. I'm putting this up since it solves my use case nicely. I do want to add more tests once we have agreed on changes here.

There are different reasons why options are added:

### skipTag
See discussion here: https://github.com/semantic-release/semantic-release/issues/753#issuecomment-455241656
Basically we can disable all plugin execution, however there is no way to prevent the tagging. So if we just want to output the version to a file without changing anything in the repo, this option is necessary.

### skipPrCheck
See discussion here: https://github.com/semantic-release/semantic-release/issues/1074
Basically CircleCI has weird behaviour. I can provide show case of the behaviour if that is desired.

### DIFF
--- a/index.js
+++ b/index.js
@@ -21,7 +21,7 @@ marked.setOptions({renderer: new TerminalRenderer()});
 
 async function run(context, plugins) {
   const {cwd, env, options, logger} = context;
-  const {isCi, branch: ciBranch, isPr} = envCi({env, cwd});
+  const {isCi, branch: ciBranch, isPr, prBranch} = envCi({env, cwd});
 
   if (!isCi && !options.dryRun && !options.noCi) {
     logger.warn('This run was not triggered in a known CI environment, running in dry-run mode.');
@@ -39,12 +39,12 @@ async function run(context, plugins) {
     });
   }
 
-  if (isCi && isPr && !options.noCi) {
+  if (isCi && isPr && !options.noCi && !options.skipPrCheck) {
     logger.log("This run was triggered by a pull request and therefore a new version won't be published.");
     return false;
   }
 
-  if (ciBranch !== options.branch) {
+  if ((isPr ? prBranch : ciBranch) !== options.branch) {
     logger.log(
       `This test run was triggered on the branch ${ciBranch}, while semantic-release is configured to only publish from ${
         options.branch
@@ -102,8 +102,8 @@ async function run(context, plugins) {
 
   await plugins.prepare(context);
 
-  if (options.dryRun) {
-    logger.warn(`Skip ${nextRelease.gitTag} tag creation in dry-run mode`);
+  if (options.dryRun || options.skipTag) {
+    logger.warn(`Skip ${nextRelease.gitTag} tag creation in dry-run / skip-tag mode`);
   } else {
     // Create the tag before calling the publish plugins as some require the tag to exists
     await tag(nextRelease.gitTag, {cwd, env});
--- a/docs/usage/configuration.md
+++ b/docs/usage/configuration.md
@@ -115,6 +115,20 @@ CLI arguments: `-d`, `--dry-run`
 
 Dry-run mode, skip publishing, print next version and release notes.
 
+### skipTag
+Type: `Boolean`<br>
+Default: `false`
+CLI arguments: `--skip-tag`
+
+Prevent tagging of the git commit on release.
+
+### skipPrCheck
+Type: `Boolean`<br>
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -39,19 +39,21 @@ async function run(context, plugins) {
...
-  if (ciBranch !== options.branch) {
+  if (![ciBranch, '*'].includes(options.branch)) {
     logger.log(
       `This test run was triggered on the branch ${ciBranch}, while semantic-release is configured to only publish from ${
         options.branch
       }, therefore a new version won’t be published.`
     );
     return false;
   }
+  options.branch = isPr ? prBranch : ciBranch;
```
simlu: Ensure branch is correctly set for further use.

simlu: This should probably be moved to [L46](https://github.com/semantic-release/semantic-release/pull/1076/files#diff-168726dbe96b3ce427e7fedce31bb0bcR46)

index.js
```diff
@@ -39,19 +39,21 @@ async function run(context, plugins) {
...
   }
 
-  if (isCi && isPr && !options.noCi) {
+  if (isCi && isPr && !options.noCi && !options.skipPrCheck) {
     logger.log("This run was triggered by a pull request and therefore a new version won't be published.");
     return false;
   }
 
-  if (ciBranch !== options.branch) {
+  if (![ciBranch, '*'].includes(options.branch)) {
```
simlu:
```suggestion
  if (![isPr ? prBranch : ciBranch, '*'].includes(options.branch)) {
```
Is probably better.

pvdlg:
Before you spend more time on the PR, this is not something we will implement.

As mentioned in #753, outputting the version to be released without making a release is a feature that we are still evaluating.
So far the only use cases mentioned there that  are commons enough and explained in detail are supported via the pre-release feature in version 16.0.0-beta. This use case is to publish release for integration tests purpose only then once the test are successful to publish on the main release channel. This use case is detailed there: https://github.com/semantic-release/semantic-release/blob/beta/docs/recipes/pre-releases.md.
Other comments in #753 are either:
- of the type "I want it because that's my workflow" without explanation justifying why that workflow is better than the proposed alternative
- or are not common enough to justify the implementation of the feature

Regarding skipping the PR check this is not something we will implement as we do not want to make releases from a PRs for obvious reasons. If we fail to detect properly that a CI job has been triggered by PR, then will fix the detection.

simlu:
@pvdlg Sure. I've mostly spend time on this since I'm now deploying using my fork. That works great for now.

pongells:
> Before you spend more time on the PR, this is not something we will implement.
> 
> As mentioned in #753, outputting the version to be released without making a release is a feature that we are still evaluating.
> So far the only use cases mentioned there that are commons enough and explained in detail are supported via the pre-release feature in version 16.0.0-beta. This use case is to publish release for integration tests purpose only then once the test are successful to publish on the main release channel. This use case is detailed there: https://github.com/semantic-release/semantic-release/blob/beta/docs/recipes/pre-releases.md.
> Other comments in #753 are either:
> 
> * of the type "I want it because that's my workflow" without explanation justifying why that workflow is better than the proposed alternative
> * or are not common enough to justify the implementation of the feature
> 
> Regarding skipping the PR check this is not something we will implement as we do not want to make releases from a PRs for obvious reasons. If we fail to detect properly that a CI job has been triggered by PR, then will fix the detection.

I have a use case and a motivation for the `--skip-tag`. 

Use case: I want to use this module locally to generate the CHANGELOG and update the version(s) of my package(s). I want to be able to check the CHANGELOG and eventually manually modify it before merging to master for release. 

Motivation: we can customize and/or skip almost everything, I don't see why the tagging should be forced upon us when it would be so easy to just add the suggested flag.

rplotkin:
So, please look at the git-flow and one-flow models. In these models, the release branch _is what is tagged_. That release branch undergoes QA review (ideally), and is merged from a PR. I have to work around the refusal to integrate --skip-pr-check despite our git pattern being one commonly utilized in enterprise.

Ridermansb:
Yes.. I agree with @rplotkin .. I have the same issue with `--skip-pr-check`

JoA-MoS:
Use Cases (any language / framework that requires the build number at build time) 
* C#
* Java / Gradle (gradle.properties)
* There are probably a lot more
* Bundling your app into a docker image and using semantic release to publish the docker image

PR #1083, opened 2019-01-27
Add a reference to semantic-release-ado on the plugins list

lluchmk:
Adds a reference to the semantic-release-ado plugin on docs/extending/plugins-list.md.


lluchmk:
Commit message not following convention, will reopen with a proper message.

Issue #214, opened 2016-04-09
Signed tags


Conaclos:
Hi!

I'm currently signing each tag (GPG signing) by default.

Is there any solution to force semantic-release to use the command `git tag -s` instead of a github token?
Of course, I run semantic-release on local.

Thanks.

destroyerofbuilds:
`semantic-release` would still need your GitHub token to push an update to your project's Releases page. 

Otherwise, it would be nice to have signed tags. Would go really well with the recent release of signature verification released by GitHub.

Conaclos:
@destroyerofbuilds Thanks for your relevant reply ;)

PR #1106, opened 2019-02-26
feat(publish): add flag to enable git tag signing

nicolasbrugneaux:
Hello! I noticed it was impossible to sign tags (even using `git config tag.forceSignAnnotated` or `commit.gpgsign`). So I digged into the code and added the option to do so.

I hope this is okay and should be completely backwards-compatible 😄 

[Related Issue semantic-release/semantic-release#214]

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -34,6 +34,7 @@ Usage:
     .option('success', {...stringList, group: 'Plugins'})
     .option('fail', {...stringList, group: 'Plugins'})
     .option('debug', {describe: 'Output debugging information', type: 'boolean', group: 'Options'})
+    .option('sign', {alias: 's', describe: 'Enable GPG signing of the tag', type: 'boolean', group: 'Options'})
     .option('d', {alias: 'dry-run', describe: 'Skip publishing', type: 'boolean', group: 'Options'})
     .option('h', {alias: 'help', group: 'Options'})
     .option('v', {alias: 'version', group: 'Options'})
--- a/docs/usage/configuration.md
+++ b/docs/usage/configuration.md
@@ -135,6 +135,14 @@ Output debugging information. This can also be enabled by setting the `DEBUG` en
 
 **Note**: The `debug` is used only supported via CLI argument. To enable debug mode from the [JS API](../developer-guide/js-api.md#javascript-api) use `require('debug').enable('semantic-release:*')`.
 
+### sign
+
+Type: `Boolean`<br>
+Default: `false`<br>
+CLI argument: `--sign` / `-s`
+
+Set to `true` to enable [git tag signatures](https://git-scm.com/book/id/v2/Git-Tools-Signing-Your-Work).
+
 ## Git environment variables
 
 | Variable              | Description                                                                                                                                                                                                                    | Default                              |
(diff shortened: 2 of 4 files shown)


pvdlg:
You can use a GPG key to sign your commits. If you use the `@semantic-release/git` plugin to push a release commit you can https://github.com/semantic-release/git#gpg-signature.

What value does signing the tag itself (in addition of the commits) would provide?
Is there a use for it?

I quite reluctant adding yet another option...That create more complexity, more maintenance and more confusion for users...

nicolasbrugneaux:
Thanks for the quick reply! The link you provided was the first thing I looked at, unfortunately it's only for CI environement as it relies on the passphrase being available in the CI environment of 
eg: 
```yml
before_install:
  # Decrypt the git_gpg_keys.asc.enc key into /tmp/git_gpg_keys.asc
  - openssl aes-256-cbc -K $encrypted_AAAAAAAAAAAA_key -iv $encrypted_BBBBBBBBBBBB_iv -in git_gpg_keys.asc.enc -out /tmp/git_gpg_keys.asc -d
  # Make sure only the current user can read the keys
  - chmod 600 /tmp/git_gpg_keys.asc
  # Import the gpg key
  - gpg --batch --yes --import /tmp/git_gpg_keys.asc
  # Create a script that pass the passphrase to the gpg CLI called by git
  - echo '/usr/bin/gpg2 --passphrase ${GPG_PASSPHRASE} --batch --no-tty "$@"' > /tmp/gpg-with-passphrase && chmod +x /tmp/gpg-with-passphrase
  # Configure git to use the script that passes the passphrase
  - git config gpg.program "/tmp/gpg-with-passphrase"
  # Configure git to sign the commits and tags
  - git config commit.gpgsign true
  # Configure git to use your GPG key
  - git config --global user.signingkey ${GPG_KEY_ID}
```

This option would allow to use semantic release with tag signing when doing local releases.

As for why, it's yet an additional layer of authenticity to the releases on the repo, which is appreciable.

While I understand the concern of adding another option , I just haven't found a suitable solution thus far 😞

pvdlg:
If you want to sign your commit locally it's even more simple. A simple Google search will give you a lot of results. For example: https://git-scm.com/book/en/v2/Git-Tools-Signing-Your-Work

I understand the "additional layer of authenticity" argument. However if you don't even sign your commit, I don't really understand the need to sign the tags. Start by configuring your environment to sign your commits.

I don't think signing tags is a feature important enough to justify the addition of a new option. Especially considering you can already sign commits.

Maybe you can find a way to configure Git to always sign tags: https://stackoverflow.com/a/50647394/2238351

travi:
also, keep in mind that you would have to make your GPG key available on your CI server in order to do the signing. you at minimum would want a separate key from your normal one so that you dont put your personal key at more risk of exposure. at that point, things start to get a little confusing around who actually signed it since its not your normal trusted key. 

personally, if i were to ever consider this, i would create a key that is associated with a bot account rather than me and use it only for this. would the signature of a key for a bot be as trusted as a maintainer, though? the risk of accidentally exposing the key somehow from CI would probably outweigh the value in my mind. plus, in order to use the key, you would either need to create a key with no passphrase (which has no protection if exposed), or also expose the passphrase to the CI (which would then also be possible to steal if anyone gained enough access to steal the key).

nicolasbrugneaux:
Thank you both for your feedback! 😃

PR #1133, opened 2019-03-25
docs: add a plugin for Gradle project

KengoTODA:
Thanks for sharing a great product, that helps me to reduce CI/CD related tasks! I believe that we can kill all human in near future :)

I want to add one more plugin that works for [Gradle](https://gradle.org/) project. Please have a review.
Thanks in advance!


KengoTODA:
Thank you!

PR #1135, opened 2019-03-26
Update getting-started.md

jvalley91:
Fixin' docs

### DIFF
--- a/docs/usage/getting-started.md
+++ b/docs/usage/getting-started.md
@@ -1,10 +1,10 @@
 # Getting started
 
 In order to use **semantic-release** you must follows this steps:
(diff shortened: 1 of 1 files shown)


jvalley91:
To be specific: In "Getting started", changed the unorganized list to an organized one, assuming that these should be done in order.

pvdlg:
Ok to change the list to an ordered one. However it's not the proper syntax, the number should be followed by a period like this:
```
1. xxxxx
2. yyyyy
```

pvdlg:
Closing as no response was provided.
Also those steps doesn't necessarily have to be done in a specific order.

PR #1140, opened 2019-04-04
fix(cli): use unmodified PATH when checking git version

patrickleet:
See conversation at end of the issue:
https://github.com/semantic-release/npm/issues/143

And:
https://github.com/sindresorhus/execa/issues/196

### DIFF
--- a/bin/semantic-release.js
+++ b/bin/semantic-release.js
@@ -23,7 +23,7 @@ See https://github.com/semantic-release/semantic-release/blob/master/docs/suppor
 
 execa
   .stdout('git', ['--version'], {
-    preferLocal: false
+    preferLocal: false,
   })
   .then(stdout => {
     var gitVersion = findVersions(stdout)[0];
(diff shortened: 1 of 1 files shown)


patrickleet:
I worry this means other parts of semantic-release could potentially be using the wrong git version, aside from a version check where version 2 is actually required unless `preferLocal: false` is added in other places as well.

patrickleet:
still using my temp package I published to get around this when using semantic-release with jenkins x

mateodelnorte:
Experiencing the same issue. An approval and merge on this would be great.

patrickleet:
Yea, this affects everyone using Jenkins X who wants to use semantic-release

gr2m:
Can you resolve the conflict please?

pvdlg:
Changing that value only there won't fix the problem...It will just use the correct version when doing the version check not when using `git` anywhere else.

execa@2.0.0 use `false` as the default so that should solve the problem.
We'll move to execa@2.0.0 in v16.0.0 (as it require us to upgrade the minimum version of Node that we support.)

pvdlg:
execa@2.0.0 updated here: https://github.com/semantic-release/semantic-release/pull/1217

PR #1155, opened 2019-04-18
docs: Add support for versions that contain zeroes in sample regex

jeremypress:
example: 10.x.x is not currently supported as the regex only looks for 1-9


pvdlg:
Fixed by #1176

jeremypress:
Much appreciated @pvdlg!

Issue #1171, opened 2019-05-03
allow plugin to overwrite what env-ci does


cyberhck:
## New feature motivation
I'm configuring teamcity and it's very stubborn, versions are different, I can set branch name on environment variable, but env-ci won't let us do anything.

## New feature description

It'd be awesome, if we could do either of two things,
- totally allow plugin to replace envCi (not awesome)
- allow user replace envCi with their function, but pass in envCi as parameter (they could simply install themselves, but semantic-release could give them env and cwd, so it's easy)
- allow user to overwrite values passed, kind of like afterCi method which receives whatever envCi returns, then plugin can update values and return.

## New feature implementation
well, this should be non breaking change, as far as I see, it should be fairly simple as well.

Plugins will have a one more place where we could do this, maybe a new step, for example (of course better naming later), envCi, then on run function, after envCi is done, (or instead of), depending on what we decide, we can simply check if the plugin is there, and that could overwrite any info from envCi.

If you come up with what path to choose, I can send a PR :)

cyberhck:
I've tried to come up with #1172 can someone see if that's something we'd go with?

cyberhck:
for now I've changed it to overwrite branch name

gr2m:
Is this a problem that could be fixed in https://github.com/pvdlg/env-ci instead? [The support matrix](https://github.com/pvdlg/env-ci#supported-ci) lists teamcity and claims it reads out the `branch` configuration. What exactly is not working? Maybe open an issue there instead if you don’t get the branch configuration you would expect?

cyberhck:
thing is, I'm running my tests and semantic-release on a docker container, by mounting volume, so there's no other way of getting branch name because git branch doesn't give anything, and env-ci looks at some file which isn't present in my case.

It'd be awesome if we could convert env-ci usage into a plugin instead, by default we can use env-ci, but if someone wants to overwrite, they should be able to.

gr2m:
@pvdlg put the most thought into this, let’s wait to see what he thinks.

Maybe `env-ci` could allow to override the properties with special `ENV_CI_*` environment variables?

cyberhck:
I don't think they will https://github.com/pvdlg/env-ci/pull/83 I tried to give a PR, but it seems like they might not be aon board.

gr2m:
[...]

PR #1172, opened 2019-05-03
feat(envCi): plugin support for determining branch name

cyberhck:
This is just a proof of concept of what I think it would seem like, this way it shouldn't have any breaking change, adding some naming convention, and adding some tests should work, this PR doesn't have to be merged, it's just in reference to #1171


pvdlg:
Closing per https://github.com/semantic-release/semantic-release/issues/1171#issuecomment-501541725

Issue #1170, opened 2019-05-01
getBranches does not use git auth url


dotfold:
## Current behavior

I'm running semantic-release in a `node:11` docker container, with `GL_URL` and `GL_TOKEN` values provided.

During the setup phase after [`verify(context)`](https://github.com/semantic-release/semantic-release/blob/beta/index.js#L54), [`getBranches`](https://github.com/semantic-release/semantic-release/blob/beta/index.js#L54) is then called which eventually results in the invocation of the git command `git ls-remote --heads origin`. This is made without any auth and results in an authentication error from the remote.

I've followed the code path through, and the command is invoked in [`lib/branches/expand.js`](https://github.com/semantic-release/semantic-release/blob/beta/lib/branches/expand.js#L6)

Should `getGitAuthUrl` be run first, and then the `options.repositoryUrl` be passed through to `getBranches` and subsequently also the call to `fetch` in [`lib/branches/index`](https://github.com/semantic-release/semantic-release/blob/beta/lib/branches/index.js#L20)?

I've added that change in locally and `getBranches` works when using the `gitAuthUrl` instead of `origin`.

I'd be happy to submit a PR if this approach is valid.

## Expected behavior

There should not be an authentication error when a token is supplied.

## Environment

- **semantic-release** version: 16.0.0-beta.18
- CI environment: docker
- Plugins used: 
- **semantic-release** configuration: 
```
---
branches:
  - master
  - name: next
    prerelease: 'rc'
  - name: beta
    prerelease: 'beta'
plugins:
  - '@semantic-release/commit-analyzer'
  - '@semantic-release/release-notes-generator'
  - '@semantic-release/gitlab'
success: false
fail: false
```
- CI logs: 

Note: this is running in a `node:11` docker container.

```
root@94e0a71c0f6d:/src# GL_URL=<REDACTED> GL_TOKEN=<REDACTED> npm run release:dry

> beta-channel-example@1.0.0 release:dry /src
> semantic-release --debug --dry-run

[10:38:19 PM] [semantic-release] › ℹ  Running semantic-release version 16.0.0-beta.18
  semantic-release:config load config from: /src/.releaserc.yaml +0ms
  semantic-release:config options values: { branches:
  semantic-release:config    [ 'master',
  semantic-release:config      { name: 'next', prerelease: 'rc' },
  semantic-release:config      { name: 'beta', prerelease: 'beta' } ],
  semantic-release:config   repositoryUrl:
[...]

Issue #1156, opened 2019-04-18
Git issues with `16.0.0-beta.18`: "Host key verification failed. fatal: Could not read from remote repository."


fernandoespinosa:
## Current behavior

We're getting some rather frustrating Git connection issues with `16.0.0-beta.18` that do not occur on the stable release `15.13.3`. We need to start using your beta because we need to begin publishing pre-releases from a `beta` branch.

Basically after plugins load, it seems to want to use SSH and ignores the config setting:

```
repositoryUrl: 'https://github.com/PolymathNetwork/polymath-sdk.git'
```

Certainly on the CI server side (TeamCity Build Agent) I am using SSH to fetch and build the sources, but I have the `GITHUB_TOKEN` environment variable properly setup and passed to the build process as well. Indeed, this all works perfectly fined with `15.13.3` but it only breaks once I start using `16.0.0-beta.18`!

## Expected behavior

Everything works perfectly fined with `15.13.3` but it only breaks once I start using `16.0.0-beta.18`!

## Environment

- **semantic-release** version:
```
[18:32:27]	[Step 4/4] yarn run v1.15.2
[18:32:27]	[Step 4/4] $ /opt/buildagent/work/f6a9fa9badf552f8/node_modules/.bin/semantic-release --debug
[18:32:28]	[Step 4/4] [6:32:28 PM] [semantic-release] › ℹ  Running semantic-release version 16.0.0-beta.18
```

- CI environment:

```
root@d9ea98f77ce9:/# lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.1 LTS
Release:	18.04
Codename:	bionic
root@d9ea98f77ce9:/# node --version
v11.14.0
root@d9ea98f77ce9:/# yarn --version
1.15.2
```

- Plugins used & **semantic-release** configuration:
```
[18:32:28]	[Step 4/4] 2019-04-18T18:32:28.329Z semantic-release:config load config from: /opt/buildagent/work/f6a9fa9badf552f8/package.json
[18:32:28]	[Step 4/4] 2019-04-18T18:32:28.338Z semantic-release:config options values: { branches:
[18:32:28]	[Step 4/4]    [ '+([1-9])?(.{+([1-9]),x}).x',
[18:32:28]	[Step 4/4]      'master',
[18:32:28]	[Step 4/4]      'next',
[18:32:28]	[Step 4/4]      'next-major',
[...]

PR #1186, opened 2019-05-20
fix(repositoryUrl): on beta repositoryUrl needs auth for pre-release flows

jmodjeski75:
Fixing issue with authentication for repositoryUrl not being used for prerelease workflow

### DIFF
--- a/index.js
+++ b/index.js
@@ -53,7 +53,8 @@ async function run(context, plugins) {
   // Verify config
   await verify(context);
 
-  context.branches = await getBranches(context);
+  options.repositoryUrl = await getGitAuthUrl(context);
+  context.branches = await getBranches(options.repositoryUrl, context);
   context.branch = context.branches.find(({name}) => name === ciBranch);
 
   if (!context.branch) {
@@ -69,13 +70,11 @@ async function run(context, plugins) {
     `Run automated release from branch ${ciBranch}${options.dryRun ? ' in dry-run mode' : ''}`
   );
 
-  options.repositoryUrl = await getGitAuthUrl(context);
-
   try {
     try {
       await verifyAuth(options.repositoryUrl, context.branch.name, {cwd, env});
     } catch (error) {
-      if (!(await isBranchUpToDate(context.branch.name, {cwd, env}))) {
+      if (!(await isBranchUpToDate(options.repositoryUrl, context.branch.name, {cwd, env}))) {
         logger.log(
           `The local branch ${
             context.branch.name
--- a/lib/branches/expand.js
+++ b/lib/branches/expand.js
@@ -2,8 +2,8 @@ const {isString, remove, omit, mapValues, template} = require('lodash');
 const micromatch = require('micromatch');
 const {getBranches} = require('../git');
 
-module.exports = async ({cwd}, branches) => {
-  const gitBranches = await getBranches({cwd});
+module.exports = async (repositoryUrl, {cwd}, branches) => {
+  const gitBranches = await getBranches(repositoryUrl, {cwd});
 
   return branches.reduce(
     (branches, branch) => [
--- a/lib/branches/index.js
+++ b/lib/branches/index.js
@@ -8,16 +8,17 @@ const expand = require('./expand');
 const getTags = require('./get-tags');
 const normalize = require('./normalize');
 
-module.exports = async context => {
+module.exports = async (repositoryUrl, context) => {
   const {cwd, env} = context;
 
   const remoteBranches = await expand(
+    repositoryUrl,
     context,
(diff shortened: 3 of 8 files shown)


jmodjeski75:
possible fix for #1170 #1156

pvdlg:
I'm not sure those changes were necessary. It seems the change is to use the authenticated URL to call `git ls-remote`. However this command should work without authentication, as the the CI was able to clone the repo.

My understanding is that if `git clone` works then `git ls-remote` should work as well.
How come the repo was cloned successfully but the branches couldn't be read on the remote?

pvdlg:
Also this commit make the build fail.

gr2m:
Let’s revert and then try again

pvdlg:
The build error happens only on Travis though, not locally.
I'll investigate a bit more and if I can't find a solution I'll revert and work on it later.

jmodjeski75:
> I'm not sure those changes were necessary. It seems the change is to use the authenticated URL to call `git ls-remote`. However this command should work without authentication, as the the CI was able to clone the repo.
> 
> My understanding is that if `git clone` works then `git ls-remote` should work as well.
> How come the repo was cloned successfully but the branches couldn't be read on the remote?

@pvdlg 
Are you using a private repository? `git ls-remote` for a private repository requires auth

pvdlg:
> Are you using a private repository? git ls-remote for a private repository requires auth

Yes but that auth should already be provided one way or another by your CI (ssh key or environment variable or other).

How could your CI run `git clone` otherwise?
What I understand is how `git clone` could work while `git ls-remote` doesn't?

jmodjeski75:
`git clone <with-auth-url>`

pvdlg:
The `git clone` is done by your CI way before semantic-release run.

Where does the `<with-auth-url>` come from? If your CI uses an authenticated URL to clone why doesn't it set that URL in the `origin` of the repo.
How do you end up in the situation of your CI running `git clone <with-auth-url>` then setting the non authenticated URL as the `origin` of the clone?

Before this PR, the `git ls-remote` was called with `origin` in order to use whatever remote URL the CI uses which should at least gives read permission.

jmodjeski75:
> why doesn't it set that URL in the `origin` of the repo.

Where is this documented?

> Before this PR, the `git ls-remote` was called with `origin` in order to use whatever remote URL the CI uses which should at least gives read permission.

no it _assumed_ there was a remote named `origin`

pvdlg:
> Where is this documented?

What do you mean? In the your CI doc? I don't know.
It's the way that every CI I tested works. It doesn't make much sense for the CI to set the `origin` as an URL that doesn't works.

What CI are you using?

> no it assumed there was a remote named origin

I don't know is *assumed* is the right term here. Every repo on which semantic-release would run is a local clone and calling `git clone <some_url>` setts the `origin` to <some_url>.

It seems your CI, for some reason calls `git clone <auth_url>` and then `git remote set-url origin <non_auth_url>`. I don't see what's the point of doing that.

jmodjeski75:
> What do you mean? In the your CI doc? I don't know.

Here in this repository, no where in the docs does it indict having to call `git remote add origin <auth_url>` if this were true, this would not be [necessary](https://github.com/semantic-release/semantic-release/blob/master/index.js#L6)2. If the consumer, CI or not, should explicitly have a `remote` named `origin` and set to the appropriate url for both `git fetch` and `git push` it should be clearly documented.
But based on the documentation [here](https://github.com/semantic-release/semantic-release/blob/master/docs/recipes/git-auth-ssh-keys.md#git-authentication-with-ssh-keys) that convenience of figuring out which auth scheme to use and what the correct repository url with auth for `git fetch` and `git push` should be is provided but if that is incorrect, remove, close the loop.

pvdlg:
I tested semantic-release on something like 20 different CI. All of them set the `origin` of the cloned repo as a URL that has read access to the repo. Some use and ssh key, some other use use an URL formatted with credentials, but they all make sure you can run git commands that read the repo without having to do anything (if you don't specify a URL `git` will use `origin` by default).

It seems your CI behave differently. Again, what CI are you using?

jmodjeski75:
> (if you don't specify a URL `git` will use `origin` by default).

sure but the repository **may not** have a `remote` called `origin` but I digress. See [git remove](https://git-scm.com/docs/git-remote) `git remove remove origin` or `git remove rm origin`

> It seems your CI behave differently. Again, what CI are you using?

I am using Jenkins
#1156 - TeamCity
#1170 - a docker based CI of some kind

alexan:
yes jenkins uses different git authentication, for checkout and for what is available, when you use git binary directly

alexan:
I don't know if these changes are connected. But after updating and usage of beta-20 in a existing project I got another error: https://github.com/semantic-release/semantic-release/issues/1199

PR #1189, opened 2019-05-24
doc: clarify that branch names use globbing

bryanlarsen:


bryanlarsen:
found the right docs.  should probably point to https://github.com/semantic-release/semantic-release/blob/beta/docs/usage/workflow-configuration.md instead

gr2m:
What is the setting that you use for your setup now? I think many folks won’t understand what "base-style pattern matching" means. It’s a good idea to link to it, but I wonder if we can make the description more clear by adding another example?

pvdlg:
Fixed by #1311

PR #1198, opened 2019-06-05
test: fix parsing of `MODULE_NOT_FOUND` errors message in unit tests

pvdlg:


index.js
```diff
@@ -53,7 +53,9 @@ async function run(context, plugins) {
   // Verify config
   await verify(context);
 
+  console.log(options.repositoryUrl)
   options.repositoryUrl = await getGitAuthUrl(context);
+  console.log(options.repositoryUrl)
```
gr2m: forgot `console.log` statements?

pvdlg: I added them to debug what's going on on Travis as the build doesn't fail locally.
I'll remove those after.

PR #1200, opened 2019-06-06
fix: revert to the correct refspec in fetch function

pvdlg:


jmodjeski75:
@pvdlg I don't have write access but being responsible for the initial change. This should not have changed and was unnecessary.

PR #1216, opened 2019-06-25
chore: document that plugins config is an override

robcresswell:
This patch documents that the `plugin` config option is an override, not a merge with the default. This is significant, for example, if you customise the github plugin as it will then prevent npm publishing by default unless the npm plugin is also defined manually.


robcresswell:
@gr2m No worries! Thanks for the maintenance work and merging 😇

PR #1247, opened 2019-07-30
docs(contributing): added instructions on how to run gitbook locally

esciara:


CONTRIBUTING.md
```diff
@@ -249,3 +250,10 @@ $ npm run test
 All the [semantic-release](https://github.com/semantic-release) repositories use [Commitizen](https://github.com/commitizen/cz-cli) to help you create [valid commit messages](#commit-message-guidelines).
 
 After staging your changes with `git add`, run `npm run cm` to start the interactive commit message CLI.
+
+## Working with the doc
+
+Working with the doc is like [working with the code](#working-with-the-code), with the difference that you might find useful to see your changes locally before submitting.
+
+You can do that by following the instructions on [how to use gitbook locally](https://til.secretgeek.net/gitbook/use_gitbook_locally.html).
```
gr2m: Can you copy the instructions here? Just in case the content at the URL disappears. Also if we have the docs here we can improve it if we find a way :)

esciara: done. But I must admit I don't really understand the part:

>Infact....
>```
>You need to install "gitbook-cli" to have access to the gitbook command anywhere on your system.
>If you've installed this package globally, you need to uninstall it.
> >> Run "npm uninstall -g gitbook" then "npm install -g gitbook-cli"
>```

I am not knowledgeable on this... If someone could help clarify...

gr2m: Honestly, it confuses me, too. I wonder if the `gitbook` module is needed to be installed globally at all? 🤔 

Maybe all that is needed to work is `npm install --global gitbook-cli`?

I just uninstalled `gitbook` with `npm uninstall -g gitbook` and it still works. I think all we need is

```
npm install --global gitbook-cli
gitbook build . # build assets into _book/
gitbook serve . # serve gitbook locally
```

gr2m: can you confirm?

esciara: Yep it works... Will change the doc accordingly... and squash the last two commits.

esciara: done

gr2m: Thanks a lot for confirming 👍

Issue #1192, opened 2019-05-31
Allow to create annotated git tag


vibhanshuc:
## New feature motivation

I am utililizing [`git describe`](https://git-scm.com/docs/git-describe) to find out the latest tag somewhere in my project. 

> The command finds the most recent tag that is reachable from a commit. If the tag points to the commit, then only the tag is shown.

For example if last released [annotated git tag](https://git-scm.com/book/en/v2/Git-Basics-Tagging) is `v1.1.0` and it is also associated with latest commit, then running `git describe` will return `v1.1.0`. When other commits are made on top of `v1.1.0` then running `git describe` returns something like this `v1.1.0-42-g0888269` where `0888269` is short sha of latest commit.

At this point if we create a new annotated git tag `v1.2.0` the `git describe` returns `v1.2.0`, and if it  is non-annotated tag then `git describe` returns `v1.1.0-g0888269` which is not reflecting the correct latest tag.

Since, semantic-release doesn't allow to create annotated tag this way of getting latest tag using `git describe` is unusable.

## New feature description

It will  allow to create [annotated tags](https://git-scm.com/book/en/v2/Git-Basics-Tagging) rather than lightweight tags when a new release is created.

## New feature implementation

We would need to change the [`tag`](https://github.com/semantic-release/semantic-release/blob/master/lib/git.js#L138) function so that it creates `annotated tag`, something like this

``` javascript
async function tag(tagName, execaOpts) {
  await execa('git', ['tag', tagName, '-a', '-m', `release ${tagName}`], execaOpts);
}
```

Not sure if it impacts other things or any change in any of the plugings is required.

gr2m:
I think tags on GitHub are currently created in https://github.com/semantic-release/github using the API, I’m not sure if what you need is supported.

Maybe it’s something you can implement using the [exec plugin](https://github.com/semantic-release/exec)?

pvdlg:
The tags are created in the core.

For information you can achieve with `git describe --tags`. See https://git-scm.com/docs/git-describe#Documentation/git-describe.txt---tags.

The `--tags` makes lightweight tags to behave like annotated tags for the purpose of `git describe`.

I'm not against using annotated tags, however I don't know what would be the impact.
Is there any other reasons to use annotated tags?

vibhanshuc:
[...]

PR #1266, opened 2019-08-15
Possibility to create annotated tags

fhuonder:
This PR contains the functionality to create annotated tags.
We want to create such tags in our pipeline. Would be great if you could pull this one in.

Regards,
Florian

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -22,6 +22,7 @@ Usage:
     .option('b', {alias: 'branch', describe: 'Git branch to release from', type: 'string', group: 'Options'})
     .option('r', {alias: 'repository-url', describe: 'Git repository URL', type: 'string', group: 'Options'})
     .option('t', {alias: 'tag-format', describe: 'Git tag format', type: 'string', group: 'Options'})
+    .option('a', {alias: 'annotate-tag', describe: 'Annotate git tag', type: 'string', group: 'Options'})
     .option('p', {alias: 'plugins', describe: 'Plugins', ...stringList, group: 'Options'})
     .option('e', {alias: 'extends', describe: 'Shareable configurations', ...stringList, group: 'Options'})
     .option('ci', {describe: 'Toggle CI verifications', type: 'boolean', group: 'Options'})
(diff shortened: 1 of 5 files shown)


pvdlg:
That feature as already been implemented in v16-beta: #1203

However it seems it create issues: https://github.com/semantic-release/semantic-release/issues/1192#issuecomment-506263912. So I'm considering reverting it.

What's the reason to create annotated tags?

fhuonder:
I see.
My version would be configurable and by default the annotated tags are disabled.

pvdlg:
ok, but what's the reason to create annotated tags?

fhuonder:
Hi,

Git supports two types of tags (lightweight and annotated).
_semantic-release_ currently only supports lightweight tags. Users who want to create annotated tags cannot create such.
Annotated tags provide more information of a tag:

- tagger information
- the date the tag was committed
- annotation message

A lightweight tag in contrast is like a unmodifiable branch.

Regards,
Florian

pvdlg:
I know what annotated tags are.

My question is: what's the benefit of using those in the context of semantic-release? What value does it bring? What does it allow you to do that you are currently not able to do?

fhuonder:
You mean because the one that creates the tag is the same person/bot that committed the version?
If so I agree that it does not make sense to create an annotated tag.

fhuonder:
Hi,
I would like to reopen this request.
Above you asked why annotated tags should be supported. Now I have a reason:
In our project we use the plugin https://github.com/danielflower/maven-gitlog-plugin to generate the release notes.
This plugin only works when the tags are annotated, otherwise it says "Light-weight tags not supported. Skipping".

Would be great if you could reconsider...

Regards,
Florian

kennethjor:
There a [Gradle pugin](https://github.com/DanySK/git-sensitive-semantic-versioning-gradle-plugin) I'm using for generating semantic versions off tags. That only supports annotated tags as well. It would be super nice if this was configurable at least, as I understand there are some who do not want to use annotation tags. FWIW [standard-version](https://github.com/conventional-changelog/standard-version) uses annotated tags as well.

sheerlox:
@travi I think we should close this in favor of https://github.com/semantic-release/semantic-release/pull/1871.

Issue #1194, opened 2019-06-03
Make semantic-release compatible with GitHub Actions


gr2m:
## New feature motivation

[GitHub Actions](https://developer.github.com/actions/) come with a token out of the box: `GITHUB_TOKEN` (see [Environment variables](https://developer.github.com/actions/creating-github-actions/accessing-the-runtime-environment/#runtime-environment-resources)).

Unfortunately that token cannot be used to authenticate git commands, not directly anyway. 

I want semantic-release to work out-of-the box with GitHub Actions. Especially in combination with GitHub’s new npm registry, this would remove the need to create any tokens by hand.

## New feature description

With a personal access token, this is possible:

```
git push https://${TOKEN}@github.com/gr2m/sandbox.git
```

But the `GITHUB_TOKEN` provided to GitHub actions is internally not a private access token, but a GitHub App installation token (🤷‍♂)

But there is a way, by prefixing it with `x-access-token` (mentioned [somewhere deep in GitHub’s developer documentation](https://developer.github.com/apps/quickstart-guides/creating-ci-tests-with-the-checks-api/#step-22-cloning-the-repository)).

So this _does_ work, but only with installation tokens

```
git push https://x-access-token:${TOKEN}@github.com/gr2m/sandbox.git
```

Installation tokens can be easily identified, they have a `v1.` prefix, so we could add a check like this

```
if (/^v1\./.test(TOKEN)) {
  // use "https://x-access-token:${TOKEN}@github.com/gr2m/sandbox.git" schema
}
```

Alternatively we can check for an [environmental variable that is always set in a GitHub Action](https://developer.github.com/actions/creating-github-actions/accessing-the-runtime-environment/#environment-variables), such as `GITHUB_ACTION`

## New feature implementation

I think the change would need to be implemented in `lib/get-git-auth-url.js`

https://github.com/semantic-release/semantic-release/blob/bcf3a853f2ad988a34edca0f484bd7df07d3c7ac/lib/get-git-auth-url.js#L46-L54

Instead of returning `auth: gitCredentials`, we could add an additional check if `gitCredentials` is a GitHub App installation access token, in which case we prefix it with `x-access-token:`

pvdlg:
We already have something in place to add a token prefix:
https://github.com/semantic-release/semantic-release/blob/7e824b2c1831d3e0226cdaa5caf4533d99e34693/lib/get-git-auth-url.js#L6-L14

[...]

PR #1273, opened 2019-08-21
fix: prefix git auth url with "x-access-token:" when run in a GitHub Action

activescott:
As mentioned in a comment at https://github.com/semantic-release/github/issues/182#issuecomment-523257373, the fix for GitHub Actions support (semantic-release/semantic-release#1194) that was made in master as part of semantic-release/semantic-release#1197 hasn't yet made it to the `beta` branch as of v16.0.0-beta.22. This PR adds the fix to the `beta` branch with the necessary minor changes since these files have slightly diverged.

I tested this and demonstrated the fix working at https://github.com/activescott/agentmarkdown/runs/198973327#step:7:61 which is directly pulling it's semantic-release dependency from this PR's branch.

### DIFF
--- a/lib/get-git-auth-url.js
+++ b/lib/get-git-auth-url.js
@@ -3,16 +3,6 @@ const {isNil} = require('lodash');
 const hostedGitInfo = require('hosted-git-info');
 const {verifyAuth} = require('./git');
 
-const GIT_TOKENS = {
-  GIT_CREDENTIALS: undefined,
-  GH_TOKEN: undefined,
-  GITHUB_TOKEN: undefined,
-  GL_TOKEN: 'gitlab-ci-token:',
-  GITLAB_TOKEN: 'gitlab-ci-token:',
-  BB_TOKEN: 'x-token-auth:',
-  BITBUCKET_TOKEN: 'x-token-auth:',
-};
-
 /**
  * Determine the the git repository URL to use to push, either:
  * - The `repositoryUrl` as is if allowed to push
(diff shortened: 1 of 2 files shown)


activescott:
I can look into the failing test but any pointers would be welcomed. I suspect a dependency is submitting? returning? different payload. Looks like it was expecting `body.type` to be `JSON` but the actual is `STRING` 🤔

gr2m:
Thanks for tackling this, I've ran into the same problem myself and had to downgrade. I'm just back from vacation and need to catch up with a lot before I can help out, please ping me again if you can't figure it out in a day or two. Good luck!

activescott:
@gr2m I could use a hand. I've looked at this quite a bit and can't put my finder on the issue. I _think_ the root of the issue is that in the run of the semantic-release cli at https://github.com/semantic-release/semantic-release/blob/3e4bb40990c4849912ff7eddf595d0823cdfcbb1/test/integration.test.js#L283 it only finds the merge commit and doesn't "update" the release as the integration test seems to think it should according to the assertion at https://github.com/semantic-release/semantic-release/blob/3e4bb40990c4849912ff7eddf595d0823cdfcbb1/test/integration.test.js#L284.

I don't mind investigating a bit more but I could use a pointer, I hate to admit it, but I'm a bit stumped on this one 😕

activescott:
@gr2m Figured this one out: Only the semantic-release/beta branch builds are compatible with semantic-release@v16. So semantic-release/beta@5.4.0-beta.1 worked, but subsequently v5.4.3 was released from master but didn't include the changes from the beta branch and 5.4.0-beta.1. But the semver carat operator does allow `5.4.3` to satisfy `^5.4.0-beta.1`. So the build started breaking after semantic-release/github@5.4.3 was released. I updated the package.json to pull in an explicit beta build of semantic-release/github at https://github.com/semantic-release/semantic-release/blob/d58d2b84fe129afdf5966bee3fe27634f72a5b53/package.json#L24

I think that's the right fix, considering these long-lived beta branches but let me know if you prefer another method.

gr2m:
That should work for the time being, until we get past the `@beta` phase of the upcoming semantic-release. Thoughts @pvdlg?

activescott:
@pvdlg @gr2m Please let me know if there is anything else I can be doing to keep this moving. I'd love to see this through to completion.

activescott:
Sorry guys that was my PR. I see your point on the cherry pick though. Will prefer that next time. @pvdlg Happy to do the "fix it later" for you if you want. I assume you just want those commits cherry picked and the merge conflicts resolved on top of the current beta branch?

PR #1249, opened 2019-07-31
docs: made doc files org clearer and augmented content

esciara:
Hi guys. Being a newbie, I struggled a fair bit getting a grip on `semantic-release` by reading the doc.

This is a first step to improving the doc: 
- renamed directories; 
- augmented a fair bit of content. (`usage` for the time being)

To be continued, but please merge this already (after reviewing it and after eventual corrections)

PR #1246, opened 2019-07-30
Doc cleanup

esciara:
Includes:

- docs(recipes): cleaned doc and navigation
  - Corrected inconsistancies between the sidebar and the actual pages. Remove non existing pages. 
  - Clarified status of "Package managers and languages" to "to be completed".
- docs: repared broken links to "CI configuration recipes"
- docs: cleaned "Developer guide" section navigation
- docs: synched README.md and SUMMARY.md

Also added my (Jetbrains) IDE files to `.gitignore`.

Issue #1250, opened 2019-08-01
docs: `gitbook serve .` does not necessarily display end result


esciara:
## Current behavior

Numbered list not rendered properly. See screenshot of a `gitbook.io` site:
<img width="763" alt="Screenshot 2019-08-01 at 06 48 37" src="https://user-images.githubusercontent.com/1336689/62266396-71a76c80-b428-11e9-84da-1f2e27c8baf2.png">

## Expected behavior

Properly rendered numbered list, as tested locally with `gitbook serve .`. See screenshot:

<img width="804" alt="Screenshot 2019-08-01 at 06 50 59" src="https://user-images.githubusercontent.com/1336689/62266484-c3e88d80-b428-11e9-8d0a-9d4a52c626ef.png">

## Environment

- **semantic-release** version: N/A
- CI environment: N/A
- Plugins used: N/A
- **semantic-release** configuration: N/A
- CI logs: N/A

esciara:
Goops... forgot to ping @gr2m . 😄

gr2m:
I don't know gitbook very well. I'll add a help wanted label, if you or someone else want to look into that, we'd appreciate it

esciara:
Had a quick look, and could not find tricks about templates... Will try a fix, but can't test it locally. Suggest to accept it and see whether it fixes the issue.

PR #1275, opened 2019-08-22
Docs improvements

pvdlg:
This PR revert all the commits from #1249, #1246, #1247. Those PR introduced several issues:
- Wrongly implied that documentation has to be read in a certain order
- Broke several links in other projects
- Wrongly removed the `README.md` in each directory that were used as a table of content when browsing directly on GitHub
- Added some inaccurate information
- Added some repetitive details that were already present in other page that were linked instead and unnecessarily complexified the doc
- Added explanation to use GitBook that is released to GitBook V2 while we use V3. The doc generated locally with this method would be different than the published one.

There was some valuable changes tough. I went through all the commits and preserved the improvements.

Fix #1275


gr2m:
Sorry that it caused so much trouble, I should have caught that.

esciara:
Sorry too. Just back online and saw all the trouble this caused... 

I noticed that some reverted changes brought back some issues (#1250 for instance), and also removed some elements, which I think don't fall into the grievances @pvdlg mentions. 

Do you mind if I propose a PR to mend this? If you agree @pvdlg and under your supervision...

Issue #691, opened 2018-03-08
npm release channel no longer considered when determining last version?


travi:
I've been working on several breaking changes that i've been mergine together before publishing as a single major release. I've been publishing pre-releases along the way and merged the branch to master tonight, which happened to be after `semantic-release` was updated to the latest version.

## Current behavior

It appears that `semantic-release` only analyzed the commits since the last tag, rather than the last release to the `latest` npm channel.

## Expected behavior

I expected that the `latest` npm dist-tag would be found as the previous version so that all commits since that release would be analyzed. 

I can't link to travis-pro logs publicly, but i just merged a similar pre-released branch today that did this successfully, showing "Found version 15.3.1 of package ... _with dist-tag latest_". 

I know I saw some of the tagging behavior moved to core, but does the npm plugin not still handle looking up the last version that was actually published as `latest`?

## Environment

- **semantic-release** version: v15.0.1
- CI environment: Travis
- Plugins used: none beyond the defaults (should be defaulting to github and npm)
- **semantic-release** configuration: https://github.com/greenkeeper-keeper/hapi-greenkeeper-keeper
- CI logs: https://travis-ci.org/greenkeeper-keeper/hapi-greenkeeper-keeper/jobs/350688820#L2612-L2639

pvdlg:
> It appears that semantic-release only analyzed the commits since the last tag, rather than the last release to the latest npm channel.

That's right we retrieve the last release with git tags now instead of using the npm registry.

However if you release some version from another branch and merge it later on the branch managed by semantic-release, it should work as long as you create tags on your other branch.
This would work because the tags would "follow" the commits when you merge.

For example if you have the branch `master` with the last release being `1.0.0` and you create a new branch `next`. Then you do some work on `next`, releasing `1.1.0` and `1.2.0` by creating the tags `v1.1.0` and `v1.2.0` and releasing on `@next` dist-tag.

Then you merge `next` into `master` and semantic-release runs. The last release will be determined as `1.2.0`, because the `v1.2.0` is in the `master` history since the merge.

If you are making prerelease that's a different story as they are outside the normal semver rules.
From the above example, if you release `2.0.0-beta.1` and `2.0.0-beta.2`, then after the merge will find the last release as `1.0.0`, which is intended, and release `2.0.0`.

Long story short, it should work as you expect as long as:
- You create tags for each release
- You don't rewrite the git history in a way that would disassociate the tags created during the release from their commits

travi:
[...]

PR #1276, opened 2019-08-23
feat: add optional support for prereleases

remie:
# Feature proposal
The fix for #691 was to exclude pre-releases when determining the last release version. However, this breaks support for using semantic-release to also manage pre-releases.

The suggestion solution is to optionally allow inclusion of pre-releases when determining the last release version by adding a `--prereleases` flag to the cli.


travi:
Prereleases are available and working quite well in the v16 beta version of semantic-release. Take a look at the docs in that branch for more details

remie:
@travi when is v16 expected to become GA?

travi:
i'm not a maintainer, so i can't say with any certainty when it will be promoted, but i've been using it myself on many projects and have found it to be very stable and effective. if you are wanting pre-release support, i don't think you should hesitate to use the beta as-is.

Issue #974, opened 2018-11-05
semantic-release+conventionalcommits.org GitHub action


bcoe:
## New feature motivation

I've been playing a bit with GitHub actions, my motivation being to show off how an automated tagging/CHANGELOG generation workflow can chain together with [actions/npm](https://github.com/actions/npm).

So far I've been prototyping with `standard-version`, which is a CLI tool I wrote that handles tagging/CHANGELOG generation via-[Conventional Commit Messages](https://www.conventionalcommits.org/en/v1.0.0-beta.2/).

_The prototype is [here](https://github.com/conventional-commits/conventional-commits-action)..._

It feels like it would potentially make more sense for the action to use some combination of `semantic-release` bins...

## New feature description

I was thinking perhaps `semantic-release` could be broken out into individual actions, each running as individual steps in a GitHub workflow ... here's my dream world:

1.  run a Travis CI action that fans out tests on a variety of platforms.

_if the build is successful ..._

1. run a `semantic-release` action that handles tagging.
2. run a `semantic-release` action that handles CHANGELOG generation. (optionally).
3. run a `semantic-release` action that mints the GitHub release (optionally).
4. run the `actions/npm` action to handle publishing to npm.

## New feature implementation

Curious what other folks think, CC: @hbetts, @gajus, @joshk, @nickvanw, @zeke

pvdlg:
What would be the point of breaking down semantic-release into multiple pieces that could run individually?

semantic-release is already modular thanks to plugins: https://github.com/semantic-release/semantic-release/blob/multi-branches/docs/usage/plugins.md

Instead a GitHub action could be one that runs when all CI pass (is that possible by the way?) and that just run `npx semantic-release`.

Breaking down semantic-release into multiple piece as you suggest would be quite complex to do and make us loose a lot of features. Currently the each plugins are orchestrated and called by the core and there is some dependencies, for example:
- If the `publish` step of a plugin fails we call the `fail` step with the corresponding `Error`, that allow to open an issue on GitHub for example
- For each successful `publish` we call the `success` plugins with information related to each release (URL, name, plugin that made the release, commits, etc...), that allow for example to add comments on issue and PR included in a release
- The `generateNotes` plugins (the default one can use conventional commits or other presets) are called and the resulting notes are passed to all `publish` plugins
- There is a `prepare` plugin step that allow to generate or modify files, that can be published by the `publish` plugins

[...]

PR #1317, opened 2019-10-13
Add GitHub actions to recipes

alissonperez:
related https://github.com/semantic-release/semantic-release/issues/974

Just add a simple [GitHub Actions](https://github.com/features/actions) recipe.

### DIFF
--- a/docs/recipes/github-actions.md
+++ b/docs/recipes/github-actions.md
@@ -2,19 +2,19 @@
 
 ## Environment variables
 
-The [Authentication](../usage/ci-configuration.md#authentication) environment variables can be configured with [Secret variables](https://help.github.com/en/articles/virtual-environments-for-github-actions#creating-and-using-secrets-encrypted-variables).
+The [Authentication](../usage/ci-configuration.md#authentication) environment variables can be configured with [Secret Variables](https://help.github.com/en/articles/virtual-environments-for-github-actions#creating-and-using-secrets-encrypted-variables).
 
-For this example, you'll need to setup a [`NPM_TOKEN`](https://docs.npmjs.com/creating-and-viewing-authentication-tokens) to publish your package to NPM registry. GitHub Actions [automatically adds](https://help.github.com/en/articles/virtual-environments-for-github-actions#github_token-secret) [`GITHUB_TOKEN`](https://help.github.com/en/articles/creating-a-personal-access-token-for-the-command-line) to be used in workflows.
+In this example an [`NPM_TOKEN`](https://docs.npmjs.com/creating-and-viewing-authentication-tokens) is required to publish a package to the npm registry. GitHub Actions [automatically populate](https://help.github.com/en/articles/virtual-environments-for-github-actions#github_token-secret) a [`GITHUB_TOKEN`](https://help.github.com/en/articles/creating-a-personal-access-token-for-the-command-line) environment variable which can be used in Workflows.
 
 ## Node project configuration
 
-[GitHub Actions](https://github.com/features/actions) supports [Workflows](https://help.github.com/en/articles/configuring-workflows) allowing to test on multiple Node versions and publishing a release only when all test pass.
+[GitHub Actions](https://github.com/features/actions) support [Workflows](https://help.github.com/en/articles/configuring-workflows), allowing to run tests on multiple Node versions and publish a release only when all test pass.
 
-**Note**: The publish pipeline must run a [Node >= 8 version](../support/FAQ.md#why-does-semantic-release-require-node-version--83).
+**Note**: The publish pipeline must run on [Node version >= 8.3](../support/FAQ.md#why-does-semantic-release-require-node-version--83).
 
 ### `.github/workflows/release.yml` configuration for Node projects
 
-This example is a minimal configuration for [`semantic-release`](https://github.com/semantic-release/semantic-release) with a build running Node 12 when receives a new commit at `master` branch. See [Configuring a workflow](https://help.github.com/en/articles/configuring-a-workflow) for additional configuration options.
+The following is a minimal configuration for [`semantic-release`](https://github.com/semantic-release/semantic-release) with a build running on Node 12 when a new commit is pushed to a `master` branch. See [Configuring a Workflow](https://help.github.com/en/articles/configuring-a-workflow) for additional configuration options.
 
 ```yaml
 name: Release
@@ -23,30 +23,34 @@ on:
     branches:
       - master
 jobs:
-  install:
-    name: "Generate release"
+  release:
+    name: Release
     runs-on: ubuntu-18.04
     steps:
-      - uses: actions/checkout@v1
-      - uses: actions/setup-node@master
+      - name: Checkout
+        uses: actions/checkout@v1
+      - name: Setup Node.js
+        uses: actions/setup-node@v1
         with:
           node-version: 12
-      - name: Install
+      - name: Install dependencies
         run: npm ci
-      - name: Generate release
+      - name: Release
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
         run: npx semantic-release
 ```
 
-## Pushing `package.json` changes to `master` branch
+## Pushing `package.json` changes to a `master` branch
 
-If you want to keep your `package.json` updated in your code versioning with your released version you could use [`@semantic-release/git`](https://github.com/semantic-release/git) plugin. To use it you'll need to [generate a personal access token](https://help.github.com/en/articles/creating-a-personal-access-token-for-the-command-line) with [permission to push changes to `master` branch](https://help.github.com/en/articles/enabling-branch-restrictions). Store the personal access token as a secret, then set the `GITHUB_TOKEN` environment variable from it (instead of from the provided `GITHUB_TOKEN` secret)
+To keep `package.json` updated in the `master` branch, [`@semantic-release/git`](https://github.com/semantic-release/git) plugin can be used.
+
+**Note**: Automatically populated `GITHUB_TOKEN` cannot be used if branch protection is enabled for the target branch. It is **not** advised to mitigate this limitation by overriding an automatically populated `GITHUB_TOKEN` variable with a [Personal Access Tokens](https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line), as it poses a security risk. Since Secret Variables are available for Workflows triggered by any branch, it becomes a potential vector of attack, where a Workflow triggered from a non-protected branch can expose and use a token with elevated permissions, yielding branch protection insignificant. One can use Personal Access Tokens in trusted environments, where all developers should have the ability to perform administrative actions in the given repository and branch protection is enabled solely for convenience purposes, to remind about required reviews or CI checks.
 
 ## Trigger semantic-release on demand
 
-There is a way to trigger semantic-relase on demand, we can just use [`repository_dispatch`](https://help.github.com/en/articles/events-that-trigger-workflows#external-events-repository_dispatch) to have control when you want to generate a release only making an HTTP request, e.g.:
+There is a way to trigger semantic-relase on demand. Use [`repository_dispatch`](https://help.github.com/en/articles/events-that-trigger-workflows#external-events-repository_dispatch) event to have control on when to generate a release by making an HTTP request, e.g.:
 
 ```yaml
 name: Release
@@ -57,10 +61,8 @@ jobs:
 # ...
 ```
 
-So just call (with your personal `GITHUB_TOKEN` or from a generic/ci user):
+To trigger a release, call (with a [Personal Access Tokens](https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line) stored in `GITHUB_TOKEN` environment variable):
 
 ```
 $ curl -v -H "Accept: application/vnd.github.everest-preview+json" -H "Authorization: token ${GITHUB_TOKEN}" https://api.github.com/repos/[org-name-or-username]/[repository]/dispatches -d '{ "event_type": "semantic-release" }'
 ```
-
-And `Release` workflow will be triggered.
--- a/docs/recipes/github-actions.md
+++ b/docs/recipes/github-actions.md
@@ -47,14 +47,6 @@ A `package.json` is required only for [local](../usage/installation.md#local-ins
 
 ```json
 {
-   "release":{
-      "plugins":[
-         "@semantic-release/commit-analyzer",
-         "@semantic-release/github",
-         "@semantic-release/npm",
-         "@semantic-release/release-notes-generator"
-      ]
-   },
    "devDependencies":{
       "semantic-release":"^15.13.18"
    }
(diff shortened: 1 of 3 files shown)


docs/recipes/github-actions.md
```diff
@@ -0,0 +1,96 @@
...
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@master
+        with:
+          node-version: '12.4.0'
+      - name: Setup NPM Token
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" >> ~/.npmrc
+      - name: Setup NPM registry
+        run: npm config set registry https://registry.npmjs.org/
```
gr2m: Are you sure the two `Setup NPM` steps are necessary? It works just fine for me without it, e.g. see https://github.com/octokit/core.js/blob/master/.github/workflows/release.yml

docs/recipes/github-actions.md
```diff
@@ -0,0 +1,96 @@
+# Using semantic-release with [GitHub Actions](https://help.github.com/en/categories/automating-your-workflow-with-github-actions)
+
+## Environment variables
+
+The [Authentication](../usage/ci-configuration.md#authentication) environment variables can be configured with [Secret variables](https://help.github.com/en/articles/virtual-environments-for-github-actions#creating-and-using-secrets-encrypted-variables).
+
+For this example, you'll need to setup a `NPM_TOKEN` and a `GH_TOKEN`, this last one is recommended to use [`@semantic-release/git`](https://github.com/semantic-release/git) with a **CI user who had permission to push to master branch** to update `package.json` file.
```
gr2m: The `GH_TOKEN` is only needed if you use the `git` plugin and have branch protection enabled as far as I know. I use semantic-release with just `NPM_TOKEN` and the provided `GITHUB_TOKEN`, and we should probably document that as the standard case as it's easier to setup?

docs/recipes/github-actions.md
alissonperez: @gr2m Oh, yes! You are right... I removed it.. please, take a look again.
Thanks!

docs/recipes/github-actions.md
alissonperez: Hum, I changed a little bit the documentation. Please, take a look again @gr2m .
Thanks!

gr2m:
Hey I appreciate you squashing your commits to allow for rebase and merge, but while we still discuss it would actually be helpful to add commits so I can see your changes since my last review.

docs/recipes/github-actions.md
```diff
@@ -0,0 +1,92 @@
+# Using semantic-release with [GitHub Actions](https://help.github.com/en/categories/automating-your-workflow-with-github-actions)
+
+## Environment variables
+
+The [Authentication](../usage/ci-configuration.md#authentication) environment variables can be configured with [Secret variables](https://help.github.com/en/articles/virtual-environments-for-github-actions#creating-and-using-secrets-encrypted-variables).
+
+For this example, you'll need to setup a `NPM_TOKEN` to publish your package to NPM registry and a `GH_TOKEN` to generate a release at GitHub.
+
+If you want to push changes in your `package.json` to your `master` branch before generate release, what is recommended, you may need to add [`@semantic-release/git`](https://github.com/semantic-release/git). With this plugin your `GH_TOKEN` must have permission to push commits to `master` branch.
```
gr2m: We never said it's recommended to push changes to `package.json` back to `master` using the git plugin. Many folks do it, but many don't, it's not a recommendation. It's a good idea to document the special case that requires a custom `GH_TOKEN`, but that should be a separate section addressing this use case specifically, not the default setup instruction

docs/recipes/github-actions.md
```diff
@@ -0,0 +1,92 @@
...
+  install:
+    name: "Generate release"
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@master
+        with:
+          node-version: '12.4.0'
+      - name: Install
+        run: npm i
```
gr2m: please use `run: npm ci`

docs/recipes/github-actions.md
alissonperez: Good, thanks! I removed this section.

alissonperez:
@gr2m Changed. Could you review again, please?

What do you think about that section about `repository_dispatch` usage?

Tks!

gr2m:
I would really appreciate if you would write a description without mentioning `GH_TOKEN` and the `git` plugin at first, and then add a section about using the `git` plugin with GitHub actions and why you need to create a separate token, and how to do that. Could you do that?

> What do you think about that section about repository_dispatch usage?

I like it. You could add a separate section and call it something like `Trigger semantic-release on demand`. You could add even another section called `Trigger semantic-release on a schedule`, but we can add that later

alissonperez:
@gr2m yeah, indeed. I made your suggestions, please, take a look when possible.
Tks!

docs/recipes/github-actions.md
```diff
@@ -40,51 +38,45 @@ jobs:
...
-            {
-               "assets":[
-                  "package.json"
-               ],
-               "message":"chore(release): ${nextRelease.version} [skip ci]\n\n${nextRelease.notes}"
-            }
-         ]
+         "@semantic-release/release-notes-generator"
       ]
    },
```
gr2m: You can remove the `"release"` settings altogether, as these are now the default settings. In fact, you can remove the section on `package.json` altogether, all you need is to install `semantic-release` as devDependency

docs/recipes/github-actions.md
```diff
@@ -4,8 +4,6 @@
 
 The [Authentication](../usage/ci-configuration.md#authentication) environment variables can be configured with [Secret variables](https://help.github.com/en/articles/virtual-environments-for-github-actions#creating-and-using-secrets-encrypted-variables).
 
-For this example, you'll need to setup a `NPM_TOKEN` to publish your package to NPM registry and a `GH_TOKEN` to generate a release at GitHub.
-
```
gr2m: I would leave the first part, the `NPM_TOKEN` still needs to be created (It can be generated at `https://www.npmjs.com/settings/[USERNAME]/tokens/create`)

docs/recipes/github-actions.md
alissonperez: Indeed.. makes sense! I'll change it!

docs/recipes/github-actions.md
alissonperez: Good! I changed it. I also added links to documentation about how to generate these tokens.

alissonperez:
No problem! Thank you for your review.. If you see any other change, please LMK..

docs/recipes/github-actions.md
```diff
@@ -0,0 +1,64 @@
...
+name: Release
+on: [repository_dispatch]
+jobs:
+# ...
+```
+
+So just call (with your personal `GH_TOKEN` or from a generic/ci user):
+
+```
+$ curl -v -H "Accept: application/vnd.github.everest-preview+json" -H "Authorization: token ${GH_TOKEN}" https://api.github.com/repos/[org-name-or-username]/[repository]/dispatches -d '{ "event_type": "any" }'
```
gr2m: Instead of `"event_type": "any"`, I'd suggest this configuration

```yaml
name: Release
on: 
  repository_dispatch:
    types: [semantic-release]
jobs:
# ...
```

And then make the `curl` call with `-d '{ "event_type": "semantic-release" }'`

docs/recipes/github-actions.md
```diff
@@ -0,0 +1,64 @@
...
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@master
+        with:
+          node-version: '12.4.0'
+      - name: Install
+        run: npm ci
+      - name: Generate release
+        env:
+          GH_TOKEN: ${{ secrets.GH_TOKEN }}
```
merlinnot: GitHub actions provide a default token, which has required permissions: `GITHUB_TOKEN`, see https://help.github.com/en/articles/virtual-environments-for-github-actions#creating-and-using-secrets-encrypted-variables

docs/recipes/github-actions.md
alissonperez: Oh, good point! I didn't implement in this way because doc about action says that `repository_dispatch` has no `types` an attr. But, I tested it now and works perfectly, I'll changed it. Tks!

docs/recipes/github-actions.md
alissonperez: Tks @merlinnot! I didn't know about that. I changed it, take a look again when possible, please.

docs/recipes/github-actions.md
```diff
@@ -0,0 +1,66 @@
...
+      - name: Generate release
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: npx semantic-release
+```
+
+## Pushing `package.json` changes to `master` branch
+
+If you want to keep your `package.json` updated in your code versioning with your released version you could use [`@semantic-release/git`](https://github.com/semantic-release/git) plugin. To use it you'll need to generate a [`GITHUB_TOKEN`](https://help.github.com/en/articles/creating-a-personal-access-token-for-the-command-line) with [permission to push changes to `master` branch](https://help.github.com/en/articles/enabling-branch-restrictions).
```
merlinnot: We can change the note about `GITHUB_TOKEN` here. Users don't have to generate it, it's populated automatically in all workflows (although you still need to pass it to a job using "env"). This will simplify the setup for everyone and promote better behavior of using non-personal accounts for automation.

To see that it works just fine, you can look at this repo: https://github.com/ridedott/dependabot-auto-merge-action

docs/recipes/github-actions.md
```diff
@@ -0,0 +1,66 @@
...
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@master
+        with:
+          node-version: '12.4.0'
+      - name: Install
+        run: npm ci
+      - name: Generate release
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
merlinnot: I'd consider changing `GH_TOKEN` to `GITHUB_TOKEN` here. I believe Semantic Release supports it. Why? To keep naming a little simpler, so users don't have to wonder why it's name one way on the left and other way on the right. It's just a small suggestion tho, maybe we can collect some more feedback here.

docs/recipes/github-actions.md
```diff
@@ -0,0 +1,66 @@
...
+      - master
+jobs:
+  install:
+    name: "Generate release"
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@master
+        with:
+          node-version: '12.4.0'
```
merlinnot: It should be safe to just use 12 here? Is there any reason to specify minor and patch here?

```suggestion
          node-version: 12
```

gr2m: Agree

gr2m:
> Overall, it might be worth noting a caveat that often makes developer pull hair out of their head, that is: GitHub Actions can't trigger GitHub Actions.
> 
> To understand better why that's so painful, imagine the following scenario: there's a merge bot of any kind, that merges PRs with minimal number of approvals, generate docs, does whatever you can automate. Once this PR is merged, releases will not be triggered. Below you can find a link to an Action that merges dependency updates in a repo that uses Semantic Release. No dependency updates are released (see history).

That is a very good point, I've lost some hair over this myself although I know about. We should add a warning

alissonperez:
@gr2m and @merlinnot , I made your suggestions..

About `GitHub Actions can't trigger GitHub Actions`, I didn't understand in which flow where it's not possible to happens, because I have a flow here, in a private repo, that is triggered by a http request (using `repository_dispatch`), this flow calls semantic release that makes a commit in `master` branch and generate a new release on it. Both flows (push and tagging) are triggered by GitHub. Would it be the same flow that you described? I saw your actions repository and it's really strange that merges don't triggered other actions. Is there some documentation about it?

merlinnot:
This might give you some more information: https://github.community/t5/GitHub-Actions/Allow-triggering-actions-by-other-actions/td-p/34793

docs/recipes/github-actions.md
```diff
@@ -0,0 +1,66 @@
...
+      - name: Generate release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: npx semantic-release
+```
+
+## Pushing `package.json` changes to `master` branch
+
+If you want to keep your `package.json` updated in your code versioning with your released version you could use [`@semantic-release/git`](https://github.com/semantic-release/git) plugin. To use it you'll need to generate a [`GITHUB_TOKEN`](https://help.github.com/en/articles/creating-a-personal-access-token-for-the-command-line) with [permission to push changes to `master` branch](https://help.github.com/en/articles/enabling-branch-restrictions).
```
merlinnot: I think we can still improve this message. It suggest that users have to create this token manually in this case, but the default one already has these permissions. Do you think we can rephrase it somehow, so it's more clear?

alissonperez: Hum, good point.. I'll think about it..

alissonperez:
> This might give you some more information: https://github.community/t5/GitHub-Actions/Allow-triggering-actions-by-other-actions/td-p/34793

hum, makes sense to work for me, I'm using a specific `GH_TOKEN` for it, if I understood well how to trigger workflows from other workflows.

docs/recipes/github-actions.md
gr2m: The problem with the provided `GITHUB_TOKEN` is that it cannot push to master if you have branch protection enabled. For that you need to create a personal access token.

You can rephrase it and say 

> To use it you'll need to [generate a personal access token](https://help.github.com/en/articles/creating-a-personal-access-token-for-the-command-line) with [permission to push changes to `master` branch](https://help.github.com/en/articles/enabling-branch-restrictions). Store the personal access token as a secret, then set the `GITHUB_TOKEN` environment variable from it (instead of from the provided `GITHUB_TOKEN` secret)

```yml
        env:
          GITHUB_TOKEN: ${{ secrets.MY_PERSONAL_ACCESS_TOKEN }}
```

rarkins:
I think it’s becoming very common to have master branch protection enabled. What functionality exactly requires “pushing to master”?

alissonperez:
@rarkins Yes, it's very common. I faced this problem when I need to push my `package.json` changes (with new version) to master branch using https://github.com/semantic-release/git . What I did is to enable only the user from GH_TOKEN to push changes to `master`.

docs/recipes/github-actions.md
alissonperez: Thanks @gr2m , I changed with your changes. Please, take a look what do you think guys. @merlinnot

merlinnot:
@alissonperez I opened a PR which should hopefully shed some light on what's broken with `GITHUB_TOKEN`: https://github.com/alissonperez/semantic-release/pull/1. Could you take a look and provide me with some feedback?

alissonperez:
Yes, awesome!! Indeed! thanks @merlinnot !

PR #1328, opened 2019-10-21
feat: add ability to filter commits by path

JoA-MoS:
* adds an optional parameter commitPaths that can be defined as an array of strings. This will be used to filter commit messages similar to git log -- \<paths\>

### DIFF
--- a/lib/get-commits.js
+++ b/lib/get-commits.js
@@ -9,7 +9,7 @@ const debug = require('debug')('semantic-release:get-commits');
  *
  * @return {Promise<Array<Object>>} The list of commits on the branch `branch` since the last release.
  */
-module.exports = async ({cwd, env, lastRelease: {gitHead}, logger}) => {
+module.exports = async ({cwd, env, lastRelease: {gitHead}, logger, commitPaths}) => {
   if (gitHead) {
     debug('Use gitHead: %s', gitHead);
   } else {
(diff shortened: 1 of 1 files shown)


JoA-MoS:
Could someone provide some guidance on how I should add tests for this?

lib/get-commits.js
```diff
@@ -18,7 +18,7 @@ module.exports = async ({cwd, env, lastRelease: {gitHead}, logger}) => {
 
   Object.assign(gitLogParser.fields, {hash: 'H', message: 'B', gitTags: 'd', committerDate: {key: 'ci', type: Date}});
   const commits = (await getStream.array(
-    gitLogParser.parse({_: `${gitHead ? gitHead + '..' : ''}HEAD`}, {cwd, env: {...process.env, ...env}})
+    gitLogParser.parse({ _: [`${gitHead ? gitHead + '..' : ''}HEAD`,  ...(commitPaths ? ['--', ...commitPaths] : '')]}, {cwd, env: {...process.env, ...env}})
```
felixfbecker:
```suggestion
    gitLogParser.parse({ _: [`${gitHead ? gitHead + '..' : ''}HEAD`,  ...(commitPaths ? ['--', ...commitPaths] : [])]}, {cwd, env: {...process.env, ...env}})
```

JoA-MoS: Yes, I will make the recommend change and look at working on the documentation updates tonight/tomorrow.

JoA-MoS:
Realized there are a bunch more changes that need to be made, to be able to pull from the release config

Issue #1279, opened 2019-09-03
Add Path Filter for Analyzed Commits


DanielHabenicht:
## New feature motivation
We have a folder with an app in our repo which should be released semantically but as all commits written in the repo follow commitizen guidelines other commits do also trigger a release even if nothing changed in the app folder. 

Also, this might be the first step to semantic release supporting monorepos.

## New feature description
Add a configuration option to define a path that should be included (or narrowed down) for commit analyzation. 
Example: 
```jsonc
// releaserc.json
{
  "scopeToPath": "./path"
}
```

## New feature implementation
Just pass the option to the git-log-parser and add an option: 
```javascript
//get-commits.js
 gitLogParser.parse({_: [`${gitHead ? gitHead + '..' : ''}HEAD`, '--', './path/to/get/analyzed']}, {cwd, env: {...process.env, ...env}})
```
There are certain edge cases: 
1. If implemented the default value can't be `./`, as in conjunction with `cwd` the analyzed commits in a subfolder would be different. That means **no path should be set** by default. 

Please write back what you think about it and how the option should be named. I implement it afterwards. 🚀

pvdlg:
I don't think we would implement that.

You mention that you have commits that trigger a release even though they don't modify your app. 
That seems contrary to any commit convention. By default in semantic-release (the Angular convention is the default) only `fix`, `feat` and `perf` commit triggers a release. If a commit doesn't modify your app it shouldn't be a `fix`, a `feat` or a `perf`.

DanielHabenicht:
As I said I think that's the first step for making semantic release usable for monorepos. 

Of course, only `fix`, `feat` and `perf` should be used for changes that will trigger a release and vice versa for changes that should not trigger a release. But I am not committing against the convention: 

I have two apps in one repo which are living right beside one another: 
```
📦 MyRepo
 ┣ 📂app1
 ┣ 📂app2
```
Each app is released individually but with the current `semantic-release` version a commit which changes files in the folder of `app1` will also trigger a release for `app2` because `semantic-release` does not distinguish between the folder it is executed in (**only for commit analyzation**, I do know that it has a `cwd` setting for execution). 
So what I end up with are two apps which do get released if either of one gets a commit. 
Maybe you want to reconsider?

DanielHabenicht:
@pvdlg I think we talked past another. I want to make semantic-release filter the commits for a specific folder, as they do not modify the version of `app1` but the version of `app2`.

Issue #1212, opened 2019-06-20
Add the availability to filter commits on a sub path of the repository


jBouyoud:
## New feature motivation

With particular repository structure like this : 
```
/    | src  (that must be taken in account when releasing)
     | other stuff (that must NOT taken in account when releasing)
``` 
All commits are take in account but I'd like to take only commits that impact `src` folder.

Or with a monorepo structure : 
```
/    | pkg1
     | pkg2
```
When I run `cd pkg1 && semantic-release --tagPrefix=pkg1` or independently the same command on pkg2, all commits are take in account. 
I'd like to filter those commit to only one subpackage (at a time)

## New feature description

Add an option to be able to run semantic-release to a sub path (or a list of sub path) in the released git repository.

## New feature implementation

Add a configuration option in main semantic release configuration (or an environment variable), to be able to configure a list of subpath where semantic-release will works on.
This could be a cli option like `--gitSubPaths` or global configuration file key `gitSubPaths: []`
or a specific environment variable.

When retrieving commits to inject into the release pipeline (https://github.com/semantic-release/semantic-release/blob/master/lib/get-commits.js#L21), handle this new option by forwarding it to the git command (https://stackoverflow.com/questions/16343659/how-to-show-git-log-history-for-a-sub-directory-of-a-git-repo) 

Maybe with something like : 
```
const subPathOption = gitSubPaths.length == 0  ? '' : ` -- ${gitSubPaths.join(' ')}`;
const commits = (await getStream.array(
    gitLogParser.parse({_: `${gitHead ? gitHead + '..' : ''}HEAD${subPathOption}`}, {cwd, env: {...process.env, ...env}})
  ))
...
```

pvdlg:
Monorepo are currently not supported. If we support we want to support every cases that can occur in a monorepo. Simply filtering commits that impact a sub directory won't cover every case.
See #193 for more details.

Regarding your first example (with `src`), if you follow a commit convention, which is a semantic-release requirement, all the commits that warrant a release should impact your sources.
By default `fix`, `feat`, `perf` and `BREAKING CHANGE` trigger a release. Those type of commits should impact your sources. Other commits that doesn't impact should be a `chore` or a `build`. Those will not trigger a release.

PR #1332, opened 2019-10-22
feat: add ability to filter commits by path

JoA-MoS:
* adds an optional parameter commitPaths that can be defined as an array of strings. This will be used to filter commit messages similar to git log -- \<paths\>

RELATES TO: #1279, #1212, #193


lib/get-commits.js
```diff
@@ -9,16 +9,28 @@ const debug = require('debug')('semantic-release:get-commits');
...
+  Object.assign(gitLogParser.fields, {
+    hash: 'H',
+    message: 'B',
+    gitTags: 'd',
+    committerDate: {key: 'ci', type: Date},
+  });
   const commits = (await getStream.array(
-    gitLogParser.parse({_: `${gitHead ? gitHead + '..' : ''}HEAD`}, {cwd, env: {...process.env, ...env}})
+    gitLogParser.parse(
+      {_: [`${gitHead ? gitHead + '..' : ''}HEAD`, '--', ...commitPaths]},
```
DanielHabenicht: If `commitPaths` is omitted it should not be included here. Otherwise, the default behaviour would change, as only commits in the current directory (set by `cwd`) will be analyzed instead of previously analyzing the whole repository regardless of the path. 
```suggestion
      {_: [`${gitHead ? gitHead + '..' : ''}HEAD`, ...(commitPaths ? ['--', ...commitPaths] : [])]},
```

pvdlg:
We won't consider such PR until other issues related to monorepo are figured out. See https://github.com/semantic-release/semantic-release/issues/193#issuecomment-545118972

If there is a solution for the issue I describe there then please point me to it in #193 as I might have missed it.

lib/get-commits.js
JoA-MoS: This PR is essentially dead but to answer your question `commitPaths` defaults to an empty array and an empty array is still truthy. The code you suggested will only result in the true side of the ternary so is essentially the same as what was commited in the PR.

Maybe I could write the code in a different way to be clearer and easier to understand.

DanielHabenicht: Mhm, yep but with the `'--'` in front and no input afterwards git is defaulting to `./` so we have to check for `length != 0` then.

JoA-MoS: Is that any different from how `git log` without the trailing `--` works?

JoA-MoS: Just tested running `git log --` in the root of my project has the same results as running `git log --` in my test folder (which has much different set of commits

DanielHabenicht: Oh, ok. Then I misread my own comment on the other issue. So its even easier to implement :)

Issue #653, opened 2018-02-10
Documentation improvement - Help wanted!


pvdlg:
Most of the documentation has been rewritten recently. The content minimum content we need is there but there is a lot that can be done to improve it.

If you'd like to contribute to the semantic-release project, helping with the following would be much appreciated:
- Fix typo, misspelling and gramatical mistakes
- Improve language: make short and simpler sentences, improve clarity
- Harmonize the style and format, especially regarding `code quote`, markdown table, capitalization of brands, links

See [CONTRIBUTING.md#documentation](https://github.com/semantic-release/semantic-release/blob/caribou/CONTRIBUTING.md#documentation) for all the details.

All the repository in the [semantic-release organization](https://github.com/semantic-release) need some improvement regarding the documentation!

Thanks!

ts-npm:
I could help with this.

gr2m:
Pull requests welcome :) Thanks for your help!

whosford:
I have a question about the dryRun configuration documentation. It states the following for the default value.

`Default: false if running in a CI environment, false otherwise`

This makes it sound as though the default is always false, is this correct?

gr2m:
yes, I think that should be 

> Default: `false` if running in a CI environment, `true` otherwise

But maybe double check the code to make sure :)

EfstathiadisDimitris:
Do you guys still need help. I can jump on this in the weekend..

pvdlg:
@EfstathiadisDimitris yes, I think the documentation can always be improved. I wrote most of the doc and I'm not a native English speaker not a good documentation writer in general.
So help would be greatly appreciated. A lot of doc has been added recently on the `beta` branch, so you might want to look at that as it's probably what needs the most improvements.

ray1981jacob:
Is anything else know if I need to do

abdelrahmanahmed:
I going to write an article on how to use semantic-release, is this gonna help you with your documentation?

JoA-MoS:
I saw commitizen was removed in #1255 the contributing guide still references this as well as an npm script that doesn't exist. What guidelines do you want to provide contributors? Should it just say use the Angular Commit conventions and link to that page?

pvdlg:
The paragraph https://github.com/semantic-release/semantic-release/blob/master/CONTRIBUTING.md#commits should be just removed. The rest of the doc details the convention already.

seeliang:
do we have a document about how to make `beta` branch generate `beta` build?

[...]

PR #1335, opened 2019-10-24
docs(contributing): correct lint fix command

JoA-MoS:
The lint fix command did not work because it was missing the extra `--`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -224,7 +224,7 @@ All the [semantic-release](https://github.com/semantic-release) repositories use
 Before pushing your code changes make sure there are no linting errors with `npm run lint`.
 
 **Tips**:
-- Most linting errors can be automatically fixed with `npm run lint --fix`.
+- Most linting errors can be automatically fixed with `npm run lint -- --fix`.
 - Install the [XO plugin](https://github.com/sindresorhus/xo#editor-plugins) for your editor to see linting errors directly in your editor and automatically fix them on save.
 
 ### Tests


CONTRIBUTING.md
```diff
@@ -3,6 +3,7 @@
 ✨ Thanks for contributing to **semantic-release**! ✨
 
 As a contributor, here are the guidelines we would like you to follow:
+
```
pvdlg: Please remove the extra empty lines that were added in this PR

pvdlg: Please remove the extra empty lines that were added in this PR

CONTRIBUTING.md
```diff
@@ -145,14 +156,15 @@ The **footer** can contain a [closing reference to an issue](https://help.github
 
 #### Revert
 
+<!-- prettier-ignore -->
```
pvdlg: Please remove this comment

CONTRIBUTING.md
```diff
@@ -145,14 +156,15 @@ The **footer** can contain a [closing reference to an issue](https://help.github
...
 If the commit reverts a previous commit, it should begin with `revert: `, followed by the header of the reverted commit. In the body it should say: `This reverts commit <hash>.`, where the hash is the SHA of the commit being reverted.
 
 #### Type
 
 The type must be one of the following:
 
 | Type         | Description                                                                                                 |
-|--------------|-------------------------------------------------------------------------------------------------------------|
```
pvdlg: Please remove the change in table formatting

CONTRIBUTING.md
JoA-MoS: Ok, I will remove them 

Unless I missed something it doesn't seem like XO formats markdown files? 

I still had prettier extension enabled witch formats the markdown in a consistent manner and Based on #653 is something you are trying to do. 

prettier uses [remark-parse](https://github.com/wooorm/remark/tree/master/packages/remark-parse) to format the markdown

pvdlg: We don't format markdown with prettier because there is no way to enforce the format on the CI (with something like `xo` or `eslint`)

JoA-MoS: Alright, all the changes are made. 

as for using prettier in you CI we do this on our ci using https://prettier.io/docs/en/cli.html#check 

we also use husky to setup git hooks to make sure staged files are formatted before pushing to remote and that the commit message follows our standards. So basically the only time someone pushes something that doesn't match our requirements is when they actively try to bypass the git hooks.

pvdlg: At the time I configured the linting the `check` option didn't exists. I might revisit that at some point but not before we merge the `beta` branch to `master` as I don't want to add formatting change to the merge.

We don't use commit hook on purpose as we want to be able to push commits that might make violations while a PR is a work in progress. We only want to lint on the CI before merging a PR.
In addition we don't want to run the formatter on commit as it is confusing and error prone as it makes changes to the code without the developer control.

PR #1337, opened 2019-10-26
fix: require Node.js >=8.16

pvdlg:


tuckbick:
Seems like this should have been a breaking release 😢

pvdlg:
Yes it should have but we've decided to make an exception and avoid a breaking change just for changing the min requirement from 8.3 to 8.16...

chrskrchr:
> Yes it should have but we've decided to make an exception and avoid a breaking change just for changing the min requirement from 8.3 to 8.16

The irony of `semantic-release` not following semantic versioning... 😄

gr2m:
Only the Sith deal in absolutes 😉

PR #1361, opened 2019-11-21
fix: update beta deps for peerDependency patches

hershmire:


pvdlg:
I think it can be updated to the following if you don't mind:
```
"@semantic-release/commit-analyzer": "7.0.0-beta",
"@semantic-release/github": "^5.6.0-beta",
"@semantic-release/npm": "^6.0.0-beta",
```

This way we'll always get the latest beta of the plugins without having to make a new release every time.

hershmire:
Done

hershmire:
@pvdlg looks like there's an errant beta release for `@semantic-release/commit-analyzer@7.0.0-beta-.2` that was published previously causing what we just did to use that. I can open a new PR to point directly to `7.0.0-beta.6` unless you want to delete that package on NPM?

```
yarn info @semantic-release/commit-analyzer versions
yarn info v1.19.1
[ ...
  '6.3.3',
  '7.0.0-beta.1',
  '7.0.0-beta.2',
  '7.0.0-beta.3',
  '7.0.0-beta.4',
  '7.0.0-beta.5',
  '7.0.0-beta.6',
  '7.0.0-beta-.2' ]
```

pvdlg:
Oh yeah...that was a mistake release. Unfortunately it cannot be deleted from npm...
We can set the `@semantic-release/commit-analyzer` dependency to `^7.0.0-beta.6` so  the future ones will still work.

hershmire:
I think `^7.0.0-beta.6` will still resolve to `7.0.0-beta-.2`.

pvdlg:
Indeed...I can't find a range that works. So it's gonna have to be `7.0.0-beta.6`.
The prerelease ranges works in a very weird way...

pvdlg:
The only way to remove a version from npm is to contact the support. So I'm going to do that but that might take a while to be effective...

hershmire:
Yes they do. PR for this update here: https://github.com/semantic-release/semantic-release/pull/1362

PR #1368, opened 2019-11-28
feat: use Git notes to store the channels on which a version has been released

### LINKED ISSUE #1101: Pre-release versions without channels
## New feature motivation

We've historically versioned our packages using pre-releases (e.g. `1.0.0-beta.1`) and have always used the prerelease type as the release channel. It appears this is also the default behavior in the beta release (see https://github.com/semantic-release/semantic-release/blob/beta/lib/branches/normalize.js#L23)

This however also adds the channel to git tags (e.g. `1.0.0-beta.1@beta`). This is not only redundant in the case where the prerelease type is always the channel, but also incongruent with our historical versions. As a result, existing pre-release versions/tags aren't found because they're parsed as `channel===undefined` when searching for previous tags:

`context` from the plugin pipeline:

```js
  branches:
   [ { channel: undefined,
       tags: [Array],
       type: 'release',
       name: 'master',
       range: '>=1.0.0',
       accept: [Array] },
     { channel: undefined,
       type: 'prerelease',
       name: 'beta',
       prerelease: 'beta',
       tags: [Array] } ],
  branch:
   { channel: 'beta',
     type: 'prerelease',
     name: 'beta',
     prerelease: 'beta',
     tags: [ [Object], [Object] ] }
```

So a pre-existing tag `1.0.0-beta.1` isn't considered because it's actually looking for `1.0.0-beta.1@beta` because the normalization process is, by default, setting the channel to the branch name if it `isNil`. 

## New feature description

It would be convenient if the `isNil` check was removed; thus, if channels were intended to be used, that they would need to be explicitly specified in the config.

## New feature implementation

In _very_ limited local testing, the `isNil` check could be removed without ill effect, other than potentially breaking build/release chains that rely on the implicit use of the branch name as the channel.

### LINKED ISSUE #1191: Tag suffix @branch is unexpected
## Current behavior

It seems like for prerelease support, semantic-release tags the released version with a git tag `version@branch`, e.g. `1.1.0-beta.1@beta`. 

The problem I met with this is that if a manual release was done somewhere in between with `npm version prerelease --preid beta`, which creates a tag without the branch suffix like `2.0.0-beta.1`, that release will not be recognized as the last release, and instead the previous automatic release will be used.

It doesn't make sense to me to use this prefix, because without the suffix the commit is already unambiguous. In npm, when you set a dist tag `beta` to a version, and then also set `latest` to that version, they will refer to the exact same package contents, it's impossible that they would be different. But the tag scheme suggest it's possible for them to be different.

It also doesn't fit into the git model - branches are mutable commit pointers, tags are immutable commit pointers. A version, once created, will always be that specific commit and never change. A branch like `beta` or `latest` however moves forward all the time and changes. And when you promote a commit tagged with a prerelease version from `beta` to `latest` (through `git merge beta`, `git merge beta --no-ff`, `git push origin beta:latest`, `git reset --hard beta`, a pull request with merge etc) the tag will still be the same version, just the other branch will now include it. So what's the point of including the suffix?

Other tools may also expect that the git tag matches the npm version name without a suffix.

It also seems like this is not possible to configure with `tagFormat`: https://semantic-release.gitbook.io/semantic-release/v/beta/usage/configuration#tagformat

## Expected behavior

I would expect it to just tag it with `1.0.0-beta.1`, and know the branch/dist-tag from the currently checked out branch name.

## Environment

- **semantic-release** version: 16.0.0-beta.18
- CI environment: Travis
- Plugins used: none
- **semantic-release** configuration: https://github.com/sourcegraph/lsp-client
- CI logs: https://travis-ci.org/sourcegraph/lsp-client/jobs/538334800

pvdlg:
### BREAKING CHANGE

this feature change the way semantic-release keep track of the channels on which a version has been released.

It now use a JSON object stored in a [Git note](https://git-scm.com/docs/git-notes) instead of Git tags formatted as v{version}@{channel}.

The tags formatted as v{version}@{channel} will now be ignored. If you have made releases with v16.0.0 on branches other than the default one you will have to update your repository.

The changes to make consist in:
- Finding all the versions that have been released on a branch other than the default one by searching for all tags formatted v{version}@{channel}
- For each of those version:
  - Create a tag without the `{@channel}` if none doesn't already exists
  - Add a Git note to the tag without the `{@channel}` containing the channels on which the version was released formatted as `{"channels":["channel1","channel2"]}` and using `null` for the default channel (for example.`{"channels":[null,"channel1","channel2"]}`)
  - Push the tags and notes
  - Update the GitHub releases that refer to a tag formatted as v{version}@{channel} to use the tag without it
  - Delete the tags formatted as v{version}@{channel}

Fix #1191, Fix #1101

See this script to help with the migration: https://gist.github.com/pvdlg/6b19e529ee5c1a20645675a44e5b3239

### DIFF
--- a/index.js
+++ b/index.js
@@ -17,7 +17,7 @@ const {extractErrors, makeTag} = require('./lib/utils');
 const getGitAuthUrl = require('./lib/get-git-auth-url');
 const getBranches = require('./lib/branches');
 const getLogger = require('./lib/get-logger');
-const {verifyAuth, isBranchUpToDate, getGitHead, tag, push, getTagHead} = require('./lib/git');
+const {verifyAuth, isBranchUpToDate, getGitHead, tag, push, pushNotes, getTagHead, addNote} = require('./lib/git');
 const getError = require('./lib/get-error');
 const {COMMIT_NAME, COMMIT_EMAIL} = require('./lib/definitions/constants');
 
@@ -109,9 +109,10 @@ async function run(context, plugins) {
       if (options.dryRun) {
         logger.warn(`Skip ${nextRelease.gitTag} tag creation in dry-run mode`);
       } else {
-        await tag(nextRelease.gitTag, nextRelease.gitHead, {cwd, env});
+        await addNote({channels: [...currentRelease.channels, nextRelease.channel]}, nextRelease.gitHead, {cwd, env});
         await push(options.repositoryUrl, {cwd, env});
-        logger.success(`Created tag ${nextRelease.gitTag}`);
+        await pushNotes(options.repositoryUrl, {cwd, env});
+        logger.success(`Add channel ${nextRelease.channel} to tag ${nextRelease.gitTag}`);
       }
 
       context.branch.tags.push({
@@ -148,7 +149,7 @@ async function run(context, plugins) {
 
   const nextRelease = {
     type: await plugins.analyzeCommits(context),
-    channel: context.branch.channel,
+    channel: context.branch.channel || null,
     gitHead: await getGitHead({cwd, env}),
   };
   if (!nextRelease.type) {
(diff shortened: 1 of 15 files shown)


gr2m:
I've created a test repository here: https://github.com/gr2m/semantic-release-test-1368

I tested the happy paths without any rebasing of `beta` or `next` so far. It all worked as expected: https://github.com/gr2m/semantic-release-test-1368/releases 👍👍👍

pvdlg:
A good test would be to run what's described in https://github.com/semantic-release/semantic-release/tree/beta/docs/recipes#release-workflow

Issue #1370, opened 2019-11-28
💡 end-to-end tests


gr2m:
I've discussed with @pvdlg how we could send up end-to-end tests for semantic-release, that would tests the full life cycle. As semantic-release will become significantly more powerful with v16, having proper end-to-end tests will be even more valuable. 

Here is my plan to implement end-to-end tests using GitHub Actions

- [x] Create a new GitHub organization in which we can create test repositories: https://github.com/semantic-release-test
- [x] Create new NPM organization: https://www.npmjs.com/org/semantic-release-test
- [ ] Create a GitHub Action that gets triggered by pull requests and has `GITHUB_TOKEN` and `NPM_TOKEN` set to github/npm tokens by @semantic-release-bot 
      - create and change into a new temporary directory
      - Create a new repository with the name including the current commit hash for isolation
      - create package.json with `"name"` set to the same name as the repository
      - create a README.md and push it to the repository
      - npm install semantic-release from the current commit
      - commit new files and add commit `feat: initial release`
      - run `npx semantic-release`
      - run assertions
        - git tag was created
        - github release was created with correct notes
        - npm package was created
      - If all assertions succeeded, exit with 0. Otherwise log failed assertions and exit with 1

Once we have that setup working, we can implement more actions, e.g. one for each release workflow as described here: [docs/recipes](https://github.com/semantic-release/semantic-release/tree/2caafbaa2be54330b5b3e6dd71dda0270b566663/docs/recipes#release-workflow)

PR #1372, opened 2019-11-29
end-to-end tests

gr2m:
First stab to address #1370


pvdlg:
I have a few question regarding the implementation of those end-to-end tests.

### PRs from third parties
Would tests run on a PR opened by someone outside of the semantic-release organization, who would have to fork the repo? Last time I tested the secrets were not available for those PRs. That mean those tests will always fail when someone other than you or me opens a PR.

If those secrets were available that would be even worst as someone could open a PR that changes the tests to do something destructive. we might limit the permission on the `semantic-release` organization but such person could still mess up with `semantic-release-test` on both GitHub and npm.

### Debugging
When I open a PR and the tests fails, what do I do? How do I debug? Contrary to the tests in [integration.test.js](https://github.com/semantic-release/semantic-release/blob/master/test/integration.test.js) it doesn't seems there is a way to run those tests locally. The ability to debug seems also  limited as the only way I can think of is to add `console.log` in the source code and run push to the PR branch, wait for the whole test suite to run and read the logs. That seems to be a very tedious process especially if we add more tests that will make the run time longer.

### GitHub actions/yaml vs JavaScript
Why the choice of using GitHub actions? It forces use to write the tests in YAML which is not the most convenient and also limit the type of assertion we can do. We might be able to do every assertion we need for this simple but we might encounter limitations with more complex tests, especially if we want to test with other plugins (check the content of `@semantic-release/changelog`, check the files committed by `@semantic-release/git`, check an apm release made with `@semantic-release/apm`, check a command executed with `@semantic-release/exec`, etc...). With tests written in JavaScript we have more flexibility.
for info, we currently have similar end-to-end tests in [integration.test.js](https://github.com/semantic-release/semantic-release/blob/master/test/integration.test.js). The difference is that we use docker images to simulate the npm registry and the GitHub repo instead of using real ones.

### Safety
I understand that the problem that deleted the semantic-release repo has been solved (per #1373) however I still think creating, deleting repo and making calls to the GitHub API presents some risks as mistake can happen easily.

### Possible improvements
Here is a few that thing I think would improve end to end tests and allow to provide more value than the existing  tests in [integration.test.js](https://github.com/semantic-release/semantic-release/blob/master/test/integration.test.js):
- Run the end-to-end tests outside of the semantic-release organization on released versions (on `@next` and `@beta` and `@latest`) rather than on PRs
  - Would test the actual final product we publish (the semantic-release package on npm and its dependency) and would allow to find bugs that local tests based on the source code cannot find
  - Would prevent any accidents
  - Would work around the permission problem for PRs coming from forked repo (by mot running on PRs anymore)
- Allow to write tests more easily:
  - Provide functions doing the repetitive tasks such as, setting up node, creating a repo with `package.json` and other files, set up environment variables etc...
  - Maybe find a way to write with something more convenient than YAML?
- Allow us (the maintainer) to run the tests locally if necessary for debugging purpose

I don't know how this could be achieved though...

Unless I'm mistaken, the current implementation prevent outside contributions, as test for PRs from forked repo would always fail due to missing credentials to `semantic-release-test`. I don't think that a worthwhile trade-off.

My impression is the current implementation comes with serious drawbacks compared to what we already have in [integration.test.js](https://github.com/semantic-release/semantic-release/blob/master/test/integration.test.js) with the sole advantage of using a real GitHub repo and npm organization. Using a real GitHub repo and npm registry has some value, but not at the expanse of outside contributions. 

I didn't put that much though in a possible solution, but I think the only one would be to run those tests outside of the semantic-release repo, and trigger them on a semantic-release release not on a PR being opened.

gr2m:
> Would tests run on a PR opened by someone outside of the semantic-release organization, who would have to fork the repo? Last time I tested the secrets were not available for those PRs. That mean those tests will always fail when someone other than you or me opens a PR.

That is correct. GitHub is aware of the shortcoming so they might provide a better path in future, such as a manual trigger of the CI by one of the maintainers on PRs coming from forks.

For the time being I could implement a way to do that ourselves. It would never allow non-contributors to trigger a build.

> When I open a PR and the tests fails, what do I do? How do I debug?

I agree, debugging Actions is very annoying. 

What I would suggest we do is to move the logic for the end-to-end test into a JavaScript file that just gets called by the Action. That way we could run it locally by setting some environment variables.

> I understand that the problem that deleted the semantic-release repo has been solved (per #1373) however I still think creating, deleting repo and making calls to the GitHub API presents some risks as mistake can happen easily.

I don't see the risk any longer of accidentally deleting one of the `semantic-release/*` any longer. Also rewriting the end-to-end test into a JavaScript code would make it easier to build in more security measurements to prevent bad things from happening again.

> Run the end-to-end tests outside of the semantic-release organization on released versions (on `@next` and `@beta` and `@latest`) rather than on PRs

I think that's a good idea! We can definitely start that way, it will cause less friction. We can still add a way to run the tests on our pull requests later, too.

> Allow to write tests more easily

I think that would be addressed by rewriting the test in JavaScript and just calling it from the GitHub Action

pvdlg:
> I don't see the risk any longer of accidentally deleting one of the semantic-release/* any longer.

See https://github.com/semantic-release/semantic-release/pull/1372/checks?check_run_id=325716677#step:19:88

Now the permission are limited which is great. But still it show mistake happen really easily.
In that case if fails because it uses `SEMANTIC_RELEASE_BOT_GITHUB_TOKEN` but a regular token is also provided by default for the action. I'm guessing is the line https://github.com/semantic-release/semantic-release/pull/1372/files#diff-a675fe805bc357e64d82ceb9f9227071R114 was missing it might have deleted the repo again.

> I think that's a good idea! We can definitely start that way, it will cause less friction. We can still add a way to run the tests on our pull requests later, too.

I don't think end-to-end test should be meant to run on PRs at all as unit and integration tests already do that.
The end-to-end should focus on the things we can't test in a PR (real repo/registry, real semantic-release package, etc...). I think we should avoid redundancy between unit, integration and end-to-end tests.

- Unit tests
  - Goal: test a single function or module
  - Avoid mocking (by developing with pure functions) but provide intermediary data/input/state to tested functions/modules
  - Run locally and on PRs
- Integration tests
  - Goal: test the whole semantic-release by invoking the CLI
  - Use dockers images to mock for third-party services, doesn't mock anything else
  - Run locally and on PRs
- End-to-end tests:
  - Test what cannot be tested by integration tests: real repo/registry, real published npm package for semantic-release and plugins, possible edge cases that can't be simulated by docker images (can't think of any though)
  - Use real third-party services
  - Run outside of semantic-release repo, triggered on a semantic-release or a plugin release or manually
  - Used mainly to test a `@beta` or `@next` release before releasing it to `@latest`

gr2m:
> I'm guessing if the line https://github.com/semantic-release/semantic-release/pull/1372/files#diff-a675fe805bc357e64d82ceb9f9227071R114 was missing it might have deleted the repo again.

No, the `GITHUB_TOKEN` needs to be set explicitly, it's not provided to each step automatically. Also the GITHUB_TOKEN does not have admin privileges, so it cannot delete the repository.

> I don't think end-to-end test should be meant to run on PRs at all as unit and integration tests already do that.

End to end tests have a different role. I usually try to have end-to-end tests for the most common usage usages of semantic-release.

---

I suggest I move out the end-to-end tests from semantic-release and move them instead into a repository in @semantic-release-test. I can use [dispatch events](https://developer.github.com/v3/repos/#create-a-repository-dispatch-event) to trigger builds and use a GitHub app to create check runs on commits back in the @semantic-release repositories. GitHub Apps have more granular permissions, the app would only have permission to create check runs, it would not have access to the code at all.

Does that sound good to you?

pvdlg:
> I suggest I move out the end-to-end tests from semantic-release and move them instead into a repository in @semantic-release-test. I can use dispatch events to trigger builds and use a GitHub app to create check runs on commits back in the @semantic-release repositories. GitHub Apps have more granular permissions, the app would only have permission to create check runs, it would not have access to the code at all.

I'm not following...Do you intend to run the tests on PRs? Using the source code from the PR from the `semantic-release` repo?
If you do that you'll create a security breach, as anyone would be able to fork the repo, open a PR that logs the credentials and obtain them this way.
I don't think there is a way to run tests on a PR code securely when tests require any kind of credentials.
That's the reason GitHub actions and other CI don't provide secrets in PRs from forks. If we bypass that we'll leak the credentials.

That's why I proposed to run the end-to-end tests on a semantic-release (or a plugin) release. The test would not clone the semantic-release to get the code, it would install the semantic-release npm module.
This way  the end-to-end test would not run the PR code, only the release module.

The general release process would be:
1/ Open a PR (as we currently do)
2/ That trigger Travis to run unit and integration test (as we currently do)
3/ Review PR and merge to `next` branch
4/ That trigger a Travis build that will release on `@next`
5/ That trigger end-to-end test, testing using the npm package from `@next` (alternatively end-to-end test can be triggered manually)
6/ If the end-to-end tests passes, then we merge `next` into `master` and release on `@latest` (we don't have to merge right away though, we can wait as long as we want to let `@next` user test)

gr2m:
> Do you intend to run the tests on PRs

No, I'd run tests when releases where created, as you suggested

pvdlg:
An interesting thing to do in end-to-end test would be to test different CIs.
See https://github.com/semantic-release/semantic-release/pull/1376

Each CI has a different way to clone the repo and that can have an impact in what we do.
In the case case above a tag was not seen as part of a branch history even though it was on the remote, even after unshallowing the repo...

gr2m:
I've updated the action to only [create a dispatch event](https://developer.github.com/v3/repos/#create-a-repository-dispatch-event) in https://github.com/semantic-release-test/dispatcher. I'll do the rest from there, entirely separated from the @semantic-release org. I'll also create a separate user without access to the @semantic-release org to as another safety measure to prevent accidents.

PR #1391, opened 2019-12-15
fix: allow plugins to set environment variables to be used by other plugins

pvdlg:


dominykas:
This will be very useful! Thanks!

Will test it out on Monday latest.

dominykas:
👍 

- https://github.com/dominykas/test-things/blob/master/lib/poc-npm-set-otp.js
- https://travis-ci.org/dominykas/test-things/builds/628716345#L254

🎉

dominykas:
Is there anything I can do to move this forward?

gr2m:
sorry Dominykas for not following up, thanks for reviewing and testing, much appreciated!

pvdlg:
This was provided as a proposed solution for https://github.com/semantic-release/npm/pull/234
We didn't get a response from the author of https://github.com/semantic-release/npm/pull/234, so we don't know if that PR would be useful.

Also, I'm not sure yet if that's the best solution.

So it shouldn't have been merged at that point. I should have mentioned that in the description though...

gr2m:
Yeah, it would be good if you add a rejected review to make that clear, it looked like we just forgot about merging it to me

pvdlg:
yes, my bad.

I'll wait to see if we get a resolution in semantic-release/npm#234 and I'll revert if necessary.
I'd like us to support `continuous-auth`. We basically have two solutions:
- Integrate https://github.com/continuousauth in `@semantic-release/npm`
- Create a plugin for https://github.com/continuousauth and keep this PR, which has the inconvenient of allowing any plugin to mess with the environment variables used by other plugins...

I'm still not sure about which solution to go with. If you have an opinion on the question feel free to share.

pvdlg:
https://github.com/semantic-release/npm/pull/234#issuecomment-583644120 is also an argument to go for a proper integration in `@semantic-release/npm`.

gr2m:
+1 on integrating with https://github.com/continuousauth, it looks great!

gr2m:
Maybe start out by creating an optional plugin, before baking it into the standard npm plugin itself, to see how it's being adapted and if there are any concerns?

pvdlg:
> Maybe start out by creating an optional plugin, before baking it into the standard npm plugin itself, to see how it's being adapted and if there are any concerns?

That plugin would copy/paste most of the current `@semantic-release/npm` though.
There is already a fork existing https://github.com/electron/semantic-release-npm-cfa for a while. The goal is to avoid maintaining a a fork.
We can release it on `@next` and keep it there for a while (easy with semantic-release v16)

PR #1415, opened 2020-01-15
Filter out empty branch results

jozefcipa:
This PR adds `try-catch` block to `git.js`'s `getBranches` function. When you run the script without having any branches on remote origin (an empty repo) it throws an unclear error
```
An error occurred while running semantic-release: TypeError: Cannot read property '1' of null`
```
This PR fixes it so now the message looks like this:

```
ERELEASEBRANCHES The release branches are invalid in the `branches` configuration.
A minimum of 1 and a maximum of 3 release branches are required in the branches configuration.

This may occur if your repository does not have a release branch, such as master.

Your configuration for the problematic branches is [].

AggregateError:
    SemanticReleaseError: The release branches are invalid in the `branches` configuration.
```

### DIFF
--- a/lib/git.js
+++ b/lib/git.js
@@ -65,8 +65,8 @@ async function getCommits(from, to, execaOpts) {
 async function getBranches(repositoryUrl, execaOpts) {
   return (await execa('git', ['ls-remote', '--heads', repositoryUrl], execaOpts)).stdout
     .split('\n')
-    .map(branch => branch.match(/^.+refs\/heads\/(?<branch>.+)$/)[1])
-    .filter(Boolean);
+    .filter(Boolean)
+    .map(branch => branch.match(/^.+refs\/heads\/(?<branch>.+)$/)[1]);
 }
 
 /**
--- a/test/git.test.js
+++ b/test/git.test.js
@@ -33,6 +33,7 @@ import {
   gitAddNote,
   gitGetNote,
   gitFetch,
+  initGit,
 } from './helpers/git-utils';
 
 test('Get the last commit sha', async t => {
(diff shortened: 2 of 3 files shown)


pvdlg:
When I run `git ls-remote --heads <repoUrl>` for a repo without any branches it doesn't errors out.
So the error you experience is most likely related to something else.

Please provide information about the actual error thrown by `git ls-remote`.

jozefcipa:
When I run it on newly created empty repo it returns obviously nothing. There is no problem with `git ls-remote` itself. But when the code runs this command and then it tries to split it by new line
```
(await execa('git', ['ls-remote', '--heads', repositoryUrl], execaOpts)).stdout
        .split('\n')
```
it will return 
```
['']
```
and if we iterate over it via `map`, `branch.match` returns `null`, so result of the following code 
```
(await execa('git', ['ls-remote', '--heads', repositoryUrl], execaOpts)).stdout
        .split('\n')
        .map(branch => branch.match(/^.+refs\/heads\/(?<branch>.+)$/))
```
would be
```
[ null ]
```
But the current code always tries to fetch `[1]` of this value (currently null, though)

pvdlg:
So the solution is to move `.filter(Boolean)` before the `map`

jozefcipa:
Ah, good point 👍 Didn't realize that.

pvdlg:
Please format your code. See https://github.com/semantic-release/semantic-release/blob/master/CONTRIBUTING.md#lint

Also you would need a unit test for `getBranches` to verify the case where there is no remote branches (mostly so we won't make the problem happen again inadvertently in a future change).

jozefcipa:
Yeah, I've already noticed wrong formatting, sorry about that 🤦‍♂️I'll add tests and get back to you later. Thanks for your comments.

jozefcipa:
@pvdlg take a look now, please

pvdlg:
For information I squashed your commits together. It's better to keep related changes together in the same commit. this way when looking at the Git history or using `git blame` we can figure out easily what changes belongs together. Similarly if we want to rollback a change we avoid giving ourself the chance or rolling back only the code change and not the test or vice-versa.

PR #1416, opened 2020-01-16
Replacing firefox plugin in the community list

dbasilio:
Our company built an in-house extension and wanted to be able to release it to Chrome and Firefox both for distribution. The Chrome plugin worked without any issues, but we could not get the Firefox plugin to work. The package hasn't had a release for almost a year and half, and we think the Mozilla Add On store API changed since their last release. 

We built our own version using a different method to communicate with the Add On store that we've been using for almost 6 months now without any issues. We decided to open source the package, and will be maintaining it for the future. You can find our package [here](https://github.com/tophat/semantic-release-firefox-add-on).

We have detailed documentation about how to configure the plugin and unit testing. We're not on version 1.0 yet but will be soon, we just wanted wider adoption to make sure we didn't miss any features before bumping the major version.


gr2m:
Also, if you ask me, don't overthink a major version. It's not marketing, it's for machines. If find the need for another breaking change, just release another major version :)

PR #1417, opened 2020-01-16
Allow custom arguments for finding git commits

sagarjauhari:
Use case: If we have a release branch `release-1` which has been merged into
master and on the master branch we want semantic-release to consider only the
commits which are exclusive to the `master` branch since 'A', then the output
should only include the commits B, C, merged, E and F.
Git history:

```
-----------------------------
A-B-C-merged-E-F....(master*)
 \   /
  P-Q...........(release-1.0)
-----------------------------
```

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -24,6 +24,7 @@ Usage:
     .option('t', {alias: 'tag-format', describe: 'Git tag format', type: 'string', group: 'Options'})
     .option('p', {alias: 'plugins', describe: 'Plugins', ...stringList, group: 'Options'})
     .option('e', {alias: 'extends', describe: 'Shareable configurations', ...stringList, group: 'Options'})
+    .option('gitlogarg', {describe: 'Override git log parsing argument', type: 'string', group: 'Options'})
     .option('ci', {describe: 'Toggle CI verifications', type: 'boolean', group: 'Options'})
     .option('verify-conditions', {...stringList, group: 'Plugins'})
     .option('analyze-commits', {type: 'string', group: 'Plugins'})
--- a/index.js
+++ b/index.js
@@ -86,7 +86,7 @@ async function run(context, plugins) {
   await fetch(options.repositoryUrl, {cwd, env});
 
   context.lastRelease = await getLastRelease(context);
-  context.commits = await getCommits(context);
+  context.commits = await getCommits(context, options.gitlogarg);
 
   const nextRelease = {type: await plugins.analyzeCommits(context), gitHead: await getGitHead({cwd, env})};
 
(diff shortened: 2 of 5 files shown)


pvdlg:
Can you clarify what you try to achieve?
What is this `release-1.0` branch? Is it where you run semantic-release to make releases? Do you also make releases from `master`?
What's the point of ignoring some commits to make the release decision?

sagarjauhari:
Great questions @pvdlg . The use case here is to
1. Run semantic-release on all branches
1. Maintain several 'release' branches which are regularly up-merged all the way up to `master`
1. When a new release is cut (say `release-2.0` from master) semantic-release should only analyze commits which were made exclusively on the master branch (A, B, C, D, E) and ignore commits which were up-merged from other release branches (P, Q).

image

pvdlg:
> When a new release is cut (say release-2.0 from master) semantic-release should only analyze commits which were made exclusively on the master branch (A, B, C, D, E) and ignore commits which were up-merged from other release branches (P, Q).

But why? What problem are you trying to work around?
I don't understand what you want to ignore commits that are in the history of a branch. The code resulting of those commits will be part of the release, ignore them doesn't seems to make any sense.

Can you provide a step by step scenario that illustrate what you try to achieve and explain what problem you are facing?

sagarjauhari:
One of the reasons why I need to do this is:

I use semantic-release to  generate release notes and separate release notes are generated for each release branch. If I don't ignore the commits up-merged from other branches, then the release notes of a newly cut release-branch uses commits of *every* release branch that has commits merged to master. For the release notes of a new release, I want to use the commits which were exclusive to the master branch (because the other changes which were up-merged would have been included in the release notes of their respective release branches)

So, although the default behaviour makes sense (use all the commits in your git history), my use case is to generate a set of release notes for each release branch that I'm maintaining. Which means: I need a way to filter out the commits.

pvdlg:
If you make a release on a branch, a tag is created. when you merge those commit to another branch the tag should follow, so when semantic-release runs it will consider only the commits since the last tag, therefore no commit will be included the release note of two different releases.

So I don't understand how you can end up in the situation you describe.

sagarjauhari:
Well, in the case below, when I create the new release (release-2.0) I need to be able to only use the commits A-E and not P,Q. The issue is that because the release branches are regularly merged to master the commits from release-branches are *interleaved* with the commits on master and I need a way to filter out only the commits that were made exclusively on master. 

image

pvdlg:
Please provide a step by step way to reproduce what you are talking about.
Something like:
- `git checkout some-branch`
- `git commit -m something`
- run semantic-release on branch some-branch
- `git merge ....`
- etc...

And explain what yu

pvdlg:
Closing as no response provided

PR #1421, opened 2020-01-20
docs(travis): build on all branches by default

prokopsimek:
If you want to have more prereleases than default so you have to enable deploy on all branches. ref.: https://docs.travis-ci.com/user/deployment#examples-of-conditional-deployment

### DIFF
--- a/docs/recipes/travis.md
+++ b/docs/recipes/travis.md
@@ -92,4 +92,6 @@ jobs:
           # Use nvm to install and use the Node LTS version (nvm is installed on all Travis images)
           - nvm install lts/*
           - npx semantic-release
+        on:
+          all_branches: true
 ```


gr2m:
Thanks Prokop!

pvdlg:
I don't think we want to recommend to run semantic-release on every branches as it's unnecessary and slow down the build.
Instead we should mention in a note to filter out the branches with a glob (which the same globs used to configure semantic-release `branches`).

gr2m:
I thought about that, but I think that setting will work for all cases. If we recommend a glob there and people change the branches setting in semantic-release, it will stop working if the glob no longer matches.

Maybe we can add a glob with a warning note that it must be kept in sync with the semantic-release branches setting?

prokopsimek:
@pvdlg @gr2m The problem is that Travis ignores the deploy script at all and will not throw the warning note. It would be better to change the Travis example to a common script instead of using the Travis' deploy provider.

pvdlg:
I don't we need to change anything. With the config in the doc it should already run on every branches.

Adding the `on:branch` in the `deploy` just give the opportunity to have a different config in `branches` which would result in the test not being run on a branch while semantic-release still run on it.

At least more information should have been ask what @prokopsimek problem was and why semantic-release doesn't run on a certain branch (I'm guessing that branch is excluded from the `branches` option at the top of the Travis file which is the real problem).

pvdlg:
> @pvdlg @gr2m The problem is that Travis ignores the deploy script at all and will not throw the warning note. It would be better to change the Travis example to a common script instead of using the Travis' deploy provider.

Can you clarify that? What is you full configuration? When is it excluded?

prokopsimek:
@pvdlg That's not true. If I set the "branches" to 
```
[
    "master",	    "master",
    "next",	    "next",
    { "name": "beta", "prerelease": "beta" },	    { "name": "beta", "prerelease": "beta" },
    { "name": "alpha", "prerelease": "alpha" },	    { "name": "alpha", "prerelease": "alpha" },
    { "name": "feature/*", "prerelease": "${name.replace(/^feature|^feat\//g, 'feature')}" },	    { "name": "feature/*", prerelease: '${name.replace(/^feature|^feat\//, "feature")}' },
    { "name": "hotfix/*", "prerelease": "${name.replace(/^hotfix|^fix\//g, 'hotfix')}" },	    { "name": "hotfix/*", prerelease: '${name.replace(/^hotfix|^fix\//, "hotfix")}' },
    { "name": "debug/*", "prerelease": "${name.replace(/^debug\//g, '')}" }	    { "name": "debug/*", prerelease: true }
  ]	  ]
```
so the Travis skipped the build with 
```
Skipping a deployment with the script provider because this branch is not permitted: debug/semantic-beta-releases
```

Job link: https://travis-ci.org/DXHeroes/dx-scanner/jobs/639502673
Its commit: https://github.com/DXHeroes/dx-scanner/commit/1c1835d4564cfb74a0f4c31f245f23c412de2856

pvdlg:
Ok so after some research the `deploy` script is a particular case in Travis as by default it only deploys to specifically `master` (https://docs.travis-ci.com/user/deployment-v2/conditional#branch).

I was confused because I was wondering why we don't encounter this issue in our repos. The reason is that we don't use `deploy` (https://github.com/semantic-release/semantic-release/blob/master/.travis/semantic-release.yml#L10).

Not sure why we recommend to use `deploy` in the recipe while we don't use it ourselves. Maybe it's because the doc was written when Travis Build stage were in beta.

I think we should change the recipe to:
- add a note saying that if the Travis `branches` is set then it must match the semantic-release `branches`
- Use a regular Build Stage job instead of `deploy`

prokopsimek:
> Ok so after some research the `deploy` script is a particular case in Travis as by default it only deploys to specifically `master` (https://docs.travis-ci.com/user/deployment-v2/conditional#branch).
> 
> I was confused because I was wondering why we don't encounter this issue in our repos. The reason is that we don't use `deploy` (https://github.com/semantic-release/semantic-release/blob/master/.travis/semantic-release.yml#L10).
> 
> Not sure why we recommend to use `deploy` in the recipe while we don't use it ourselves. Maybe it's because the doc was written when Travis Build stage were in beta.
> 
> I think we should change the recipe to:
> 
> * add a note saying that if the Travis `branches` is set then it must match the semantic-release `branches`
> * Use a regular Build Stage job instead of `deploy`

I'd prefer to change the recipe to a regular Build Stage job instead of `deploy`. It will be the most common case to go to the recipe and just copy&paste a Travis configuration.

travi:
> Use a regular Build Stage job instead of deploy

i still use `deploy` in most of my projects because i find it simpler than using stages. i have the luxury of normally not needing to run my tests against several node versions, so that might not be my approach if that was different. 

however, i also have not had any success using the new importable travis configs on a project that uses stages, so i think there are downsides to only recommending stages.

pvdlg:
I think we had this debate somewhere but can find it anymore. I think we went with `deploy` because Build Stages were in beta at the time, but I'm not sure. @travi maybe you remember?
In any case it would make sense to recommend the configure that we use (or use the config that we recommend).

travi:
> Build Stages were in beta at the time, but I'm not sure.

that timing sounds about right, but i dont remember the specifics.

i've found the [importable configs](https://docs.travis-ci.com/user/build-config-imports/) so helpful recently, that i would almost suggest pointing to that instead of either option. however, it is currently in beta, so that kind of leaves us in the same situation. 

regardless, it further highlights that its hard to suggest a single approach that works well for all using travis-ci. i think i still lean toward the suggesting the simplest approach in the hopes that those needing more complex approaches are more familiar with how to configure travis-ci. i certainly hear the argument to match what is used by this org, though. tough call.

in case it is helpful, [this is my common config](https://github.com/form8ion/.travis-ci/blob/master/semantic-release.yml) that i'm using across most of my projects now, but i actually import it through another config that includes encrypted secrets, like [this one](https://github.com/form8ion/.travis-ci/blob/master/authenticated-semantic-release.yml). the wrapper is limited to projects within the same org, but the one without secrets is not.

in addition to docs, we could potentially provide a config that projects could import directly if they trust us to not inject any malicious scripts.

travi:
also, for completeness at least, here is a [thread about using imports in jobs with stages](https://travis-ci.community/t/import-different-files-into-different-jobs/6292), but i haven't seen much helpful information there so far.

PR #1432, opened 2020-01-27
docs: Add Gitea integration to plugins list

saitho:
I've written an integration for Gitea releases ~~and a plugin which enables back-merging a release into a develop branch (Git Flow)~~.


pvdlg:
For information semantic-release does support GitFlow. The goal of GitFlow is to manually manage your releases by "storing" commits in a release branch and then later on trigger that release manually.
the goal of semantic-release is to do the exact opposite by release any release commit right away and by automating the decision of making a release.

As a result, it doesn't make much sense to use semantic-release with GitFlow. If you want to manage your release manually than keep using GitFlow, if you want to automate the release decision then use semantic-release with the GitHub flow.

saitho:
Right. I removed the latter plugin, leaving only the Gitea integration. :)

PR #1444, opened 2020-02-03
docs(Troubleshooting): release not found in prereleases branch (e.g. `beta`) after rebase on `master`)

### LINKED ISSUE #1442: semantic-release does not find latest tag on beta branch
## Current behavior

`semantic-release` does not find last tag `v3.0.0-beta.1` when run on the beta branch. Instead it finds `v2.2.0` which is the last tag on the master branch.

I did rebase `beta` on master, but I deleted and re-created the `v3.0.0-beta.1` tag. So I assumed this should just work?

## Expected behavior

It should publish `v3.0.0-beta.2`

## Environment

- **semantic-release** version: 17.0.0
- CI environment: GitHub Actions
- Plugins used: _default_
- **semantic-release** configuration: _defualt_
- CI logs: https://github.com/octokit/plugin-rest-endpoint-methods.js/runs/420238746?check_suite_focus=true#step:7:67

gr2m:
fixes #1442

### DIFF
--- a/docs/support/troubleshooting.md
+++ b/docs/support/troubleshooting.md
@@ -54,3 +54,14 @@ $ git branch --contains <tag name>
 $ git tag -d <tag name>
 $ git push origin :refs/tags/<tag name>
 ```
+
+## release not found in prereleases branch (e.g. `beta`) after rebase on `master`
+
+**semantic-release** is using [git notes](https://git-scm.com/docs/git-notes) to store information about which releases happened in which branch. Git notes are attached to commits. After git history rewrite due to a rebase, the git notes created by **semantic-release** are lost.
+
+To recover from that situation, do the following:
+
+1. Delete the tag(s) for the release(s) that have been lost from the git history. You can delete each tag from remote using `git push origin :[TAG NAME]`, e.g. `git push origin :v2.0.0-beta.1`. You can delete tags locally using `git tag -d [TAG NAME]`, e.g. `git tag -d v2.0.0-beta.1`.
+2. Re-create the tags locally: `git tag [TAG NAME] [COMMIT HASH]`, where `[COMMIT HASH]` is the new commit that created the release for the lost tag. E.g. `git tag v2.0.0-beta.1 abcdef0`
+3. Re-create the git notes for each release tag, e.g. `git notes --ref semantic-release add -f -m '{"channels":["beta"]}' v3.0.0-beta.1`
+4. Push the local notes: `git push --force origin refs/notes/semantic-release`. The `--force` is needed after the rebase. Be careful.
--- a/docs/support/troubleshooting.md
+++ b/docs/support/troubleshooting.md
@@ -55,13 +55,13 @@ $ git tag -d <tag name>
 $ git push origin :refs/tags/<tag name>
 ```
 
-## release not found in prereleases branch (e.g. `beta`) after rebase on `master`
+## release not found release branch after `git push --force`
 
-**semantic-release** is using [git notes](https://git-scm.com/docs/git-notes) to store information about which releases happened in which branch. Git notes are attached to commits. After git history rewrite due to a rebase, the git notes created by **semantic-release** are lost.
+**semantic-release** is using [git notes](https://git-scm.com/docs/git-notes) to store information about which releases happened in which branch. Git notes are attached to commits. After git history rewrite due to a `git push --force`, the git notes created by **semantic-release** are lost.
 
 To recover from that situation, do the following:
 
 1. Delete the tag(s) for the release(s) that have been lost from the git history. You can delete each tag from remote using `git push origin :[TAG NAME]`, e.g. `git push origin :v2.0.0-beta.1`. You can delete tags locally using `git tag -d [TAG NAME]`, e.g. `git tag -d v2.0.0-beta.1`.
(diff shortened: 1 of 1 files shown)


docs/support/troubleshooting.md
```diff
@@ -54,3 +54,14 @@ $ git branch --contains <tag name>
 $ git tag -d <tag name>
 $ git push origin :refs/tags/<tag name>
 ```
+
+## release not found in prereleases branch (e.g. `beta`) after rebase on `master`
```
pvdlg: This is not specifically related to prereleases branches. This principle is true for any branch.

gr2m: semantic-release is using git notes for `master`, too?

docs/support/troubleshooting.md
```diff
@@ -54,3 +54,14 @@ $ git branch --contains <tag name>
...
+
+**semantic-release** is using [git notes](https://git-scm.com/docs/git-notes) to store information about which releases happened in which branch. Git notes are attached to commits. After git history rewrite due to a rebase, the git notes created by **semantic-release** are lost.
+
+To recover from that situation, do the following:
+
+1. Delete the tag(s) for the release(s) that have been lost from the git history. You can delete each tag from remote using `git push origin :[TAG NAME]`, e.g. `git push origin :v2.0.0-beta.1`. You can delete tags locally using `git tag -d [TAG NAME]`, e.g. `git tag -d v2.0.0-beta.1`.
+2. Re-create the tags locally: `git tag [TAG NAME] [COMMIT HASH]`, where `[COMMIT HASH]` is the new commit that created the release for the lost tag. E.g. `git tag v2.0.0-beta.1 abcdef0`
+3. Re-create the git notes for each release tag, e.g. `git notes --ref semantic-release add -f -m '{"channels":["beta"]}' v3.0.0-beta.1`
```
pvdlg: Here we should specify that the `channel` prop should contain the list of channel on which that version was added. If it's the default channel (the first release channel, usually `master`) then it should be set as `null`.

docs/support/troubleshooting.md
pvdlg: When a new tag is created yes. But if there is a tag without a note then we consider the version was published on the channel associated with `master`.
So if your version has been published on the channel associated with `master` (by default not specified so each plugin would use it's specific default such as `latest` for npm) and `next` then you will have a note with `{"channels:[null,"next"]"}`.

docs/support/troubleshooting.md
```diff
@@ -54,3 +54,14 @@ $ git branch --contains <tag name>
 $ git tag -d <tag name>
 $ git push origin :refs/tags/<tag name>
 ```
+
+## release not found release branch after `git push --force`
+
+**semantic-release** is using [git notes](https://git-scm.com/docs/git-notes) to store information about which releases happened in which branch. Git notes are attached to commits. After git history rewrite due to a `git push --force`, the git notes created by **semantic-release** are lost.
```
pvdlg: After git history rewrite => After **a** git history rewrite

pvdlg: For more clarity I would suggest:
> **semantic-release** is using both [git tags](https://git-scm.com/docs/git-tag) and [git notes](https://git-scm.com/docs/git-notes) to store information about which releases happened in which branch.
> After a git history rewrite due to a `git push --force`, the git tags and notes associated with commit that were rewritten are lost.

Just to make clear that everything won't be lost, only the things associated with rewritten commits.

gr2m: You can use 

````
```suggestion
Your alternative line here
```
````

to give me a one-click way of applying your suggestion

File: docs/support/troubleshooting.md [hunk 4]
```diff
@@ -54,3 +54,14 @@ $ git branch --contains <tag name>
...
+
+To recover from that situation, do the following:
+
+1. Delete the tag(s) for the release(s) that have been lost from the git history. You can delete each tag from remote using `git push origin :[TAG NAME]`, e.g. `git push origin :v2.0.0-beta.1`. You can delete tags locally using `git tag -d [TAG NAME]`, e.g. `git tag -d v2.0.0-beta.1`.
+2. Re-create the tags locally: `git tag [TAG NAME] [COMMIT HASH]`, where `[COMMIT HASH]` is the new commit that created the release for the lost tag. E.g. `git tag v2.0.0-beta.1 abcdef0`
+3. Re-create the git notes for each release tag, e.g. `git notes --ref semantic-release add -f -m '{"channels":["beta"]}' v3.0.0-beta.1`. Add all channels where the version was added. If it's only the default channel (usually `master`), then set it to an empty array: `'{"channels":[]}'`
```
pvdlg: > If it's only the default channel (usually `master`), then set it to an empty array: `'{"channels":[]}'`

That's incorrect. And empty array would creates issues. It should be:
> If this version was published only in the default channel (the first [release branch](https://github.com/semantic-release/semantic-release/blob/master/docs/usage/workflow-configuration.md#release-branches), usually `master`), then the Git note is unecessary.
> If this version was published both on the default channel and another one, then a Git note is necessary. The `channels` property must contains all the channel on which the version was published and `null` as to be used to represent the default channel, e.g. `{"channels":[null,"beta"]}`

gr2m: Ah, I misunderstood your previous feedback, sorry, will fix

Issue #1208, opened 2019-06-17
The local branch master is behind the remote one, therefore a new version won't be published


thopaw:
## Current behavior
I have a pipeline which is running for some time.
It runss unit tests, it deploys the current commit into a test stage, it runs e2e against it and then waits for a manual approval.

When it is approved, semanic-release is started to build tag the version which should go into production.

When in the meantime there are new commits on the master branch, the manual approved version can not be deployed anymore as semantic release is throwing the following error:

> The local branch master is behind the remote one, therefore a new version won't be published

This comes from index.js: 68
```
if (!(await isBranchUpToDate(options.branch, {cwd, env}))) {
        logger.log(
          `The local branch ${options.branch} is behind the remote one, therefore a new version won't be published.`
        );
        return false;
      }

```

An every build which is not on the latest version of a branch semantic-release exits with the error explained above.

## Expected behavior

In longer running pipelines or when a certain version is approved manually the execution of a deicated commit should not run in a failure but build this dedicated version.

## Environment

- **semantic-release** version:
    ```
"@semantic-release/changelog": "^3.0.4",
    "@semantic-release/gitlab": "^3.1.5",
    "semantic-release": "^15.13.16",
    "semantic-release-slack-bot": "^1.0.4",
```
- CI environment: gitlab
- Plugins used: gitlab
- **semantic-release** configuration: 

```
branch: master
plugins:
  - '@semantic-release/commit-analyzer'
  - '@semantic-release/release-notes-generator'
  - '@semantic-release/gitlab'
  - - semantic-release-slack-bot
    - notifyOnSuccess: true
      notifyOnFail: true

```
- CI logs:
```
[14:04:55] [semantic-release] › ℹ  Running semantic-release version 15.13.16
[14:04:55] [semantic-release] › ✔  Loaded plugin "verifyConditions" from "@semantic-release/gitlab"
[14:04:55] [semantic-release] › ✔  Loaded plugin "analyzeCommits" from "@semantic-release/commit-analyzer"
[14:04:55] [semantic-release] › ✔  Loaded plugin "generateNotes" from "@semantic-release/release-notes-generator"
[14:04:55] [semantic-release] › ✔  Loaded plugin "publish" from "@semantic-release/gitlab"
[14:04:55] [semantic-release] › ✔  Run automated release from branch master
[14:04:56] [semantic-release] › ℹ  The local branch master is behind the remote one, therefore a new version won't be published.

```

pvdlg:
[...]

PR #1463, opened 2020-02-17
fix: make error more verbose "local branch * is behind the remote *"

nodkz:
More details see in https://github.com/semantic-release/semantic-release/issues/1208#issuecomment-586888289


pvdlg:
See #1467. I added the `repositoryUrl` in a the logs in a different place so it would be logged every time and be useful in more cases.

nodkz:
Agreed!

PR #1471, opened 2020-02-21
Add alternative trigger options

abatilo:
These options work the same as the curl command, but use a GitHub App so there's no need to provision and manage a Personal Access Token in order to trigger the Semantic Release action.

Full disclosure, I'm personally involved with Actions Panel. Not involved at all with Action Button.


pvdlg:
The goal of semantic-release is to help releasing any code change as soon as possible, which is why we recommend to run it on each push.
The solution that you mention might be useful for certain semantic-release users but I don't know if we want to recommend those as it goes against what we recommend everywhere else.
@gr2m any thoughts on that?

abatilo:
If it's of any value, I have work projects where we don't release every time we merge because we have specific deploy days.

It feels like you lose some of the magic of using semantic-release when every release only has one note in it, versus having many stock piled.

abatilo:
@pvdlg @gr2m Following up on this thread. Any other thoughts?

PR #1497, opened 2020-03-19
docs: add tool for semantic commits

JPeer264:


pvdlg:
The goal here is to give a few example of the most popular tools that can be used, not to be exhaustive.
We want to keep the README file as short and concise as possible, so we won't list every variation of commit formatter out there. Sorry.

That would be more appropriate in something like an [awesome list](https://github.com/sindresorhus/awesome). I wanted to create one in the GitHub organization, but didn't find time yet.

PR #1505, opened 2020-03-26
chore(package): update xo to version 0.28.1

gr2m:
Closes #1488

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -41,18 +41,18 @@ Usage:
     .exitProcess(false);
 
   try {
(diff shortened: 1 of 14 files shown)


gr2m:
@pvdlg do you have an idea what the cause of this error might be? If not, I can dig into it, but wanted to ask you first

https://travis-ci.org/github/semantic-release/semantic-release/jobs/667336160#L577

pvdlg:
The rule `unicorn/string-content` create code that doesn't work as a Lodash template.
I disabled it. I think it's rules that bring close to no value and create issues.

PR #1521, opened 2020-04-11
fix: use correct ci branch context

iamogbz:
Tried testing in circleci and got this message
```txt
This test run was triggered on the branch undefined, while semantic-release is configured to only publish from publish-e2e, therefore a new version won’t be published.
```
**Note the `undefined` in the message above**

In the `env-ci` package there are distinct properties that need to be used to the get the correct ci branch context
https://github.com/pvdlg/env-ci/blob/73299d7f729971a2725cff8b582367f108cfb5ea/services/circleci.js#L21-L24


poteirard:
Any chance we can merge this? I'm having the same issue in CircleCI. Thanks for the contribution ❤️  /cc @gr2m

PR #1527, opened 2020-04-16
docs: update docs regarding angular conventions

henrist:


henrist:
I haven't checked exactly what is different between the two documents, but this was changed in `conventional-changelog-angular` 2 years ago.

https://github.com/conventional-changelog/conventional-changelog/commit/ca24a93964a6dee4d959b2a9f05c56a347fc1e11

And I guess this is relevant and shows some of the changes:

https://github.com/conventional-changelog/conventional-changelog/pull/296

PR #1530, opened 2020-04-16
fix: adapt for semver to version 7.3.2 (part II)

gr2m:
follow up for https://github.com/semantic-release/semantic-release/commit/0363790b8a5f91a8c95fc6905e3e20305db7c539\#r38547135


gr2m:
I went ahead merging it because #1528 introduced a bug

PR #1565, opened 2020-05-25
Set x-access-token as user when using Github Apps

psalaberria002:
Related to https://github.com/semantic-release/github/pull/272


gr2m:
Similar to the pull request at https://github.com/semantic-release/github/pull/272, I think introducing a new check for the `GITHUB_APP` environment variable is problematic. `GITHUB_ACTION` is a convention from GitHub itself, `GITHUB_APP` is not. The blog post you linked (https://www.jenkins.io/blog/2020/04/16/github-app-authentication/) shows that the `GITHUB_APP` variable is configurable

> ```js
> usernameVariable: 'GITHUB_APP',
>```

So even that would not be reliable. 

Unlike the workaround at https://github.com/semantic-release/github/pull/272, we cannot send REST API requests, because the `semantic-release` module should only interact with standard git commands. What we could do is to try to verify auth and in case it fails for a GitHub environment, try again using the `x-access-token` prefix.

Rada8732 (approved):

PR #1570, opened 2020-05-31
Optimize npm package size

havenchyk:
It doesn't make much sense to deliver docs to users via package

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,6 @@
   },
   "files": [
     "bin",
-    "docs",
     "lib",
     "index.js",
     "cli.js"


gr2m:
I wouldn't generalize your own opinion about distributing docs together with the code. How much of a file size difference does it make?

havenchyk:
Hey, thanks for an answer.

I’m sorry, I didn’t mean to make a general statement, instead I shared my own opinion that you already pointed out.

As for change itself, of course docs are relatively small, but CI that installs this package thousands times a day will save a couple of megabytes per day.

My idea was to create a proposal not to deliver docs (like not to deliver tests, eslintrc etc), but it’s up to you to consider or not this change as valuable

gr2m:
I'm fine either way. Any thoughts @pvdlg?

Issue #1507, opened 2020-03-27
Allow major version to be zero


jcornaz:
## New feature motivation

Semantic-versioning allow to communicate that a software is in its initial development phase by having the major version of 0. Example: `0.3.1`

This is very useful in order to start shipping very early a new software/library and gather as much feedback as possible as early as possible. Yet we don't want to miss-inform users. They must know from the version number that the software is in its initial development phase and that anything can change at any time.

See:  https://semver.org/#spec-item-4

## New feature description

When configuring the branch, asides the `prerealease: Boolean` option, there could be a `initialRelease: Boolean` (name can be improved)

In case you wonder, no `prerelease` and `initialrelease` are not mutualy exclusive:

| prerelease | initialrelease | version        |
|------------|----------------|----------------|
| true       | true           | `0.1.0-beta.1` |
| false      | true           | `0.1.0`        |
| true       | false          | `1.0.0-beta.1` |
| false      | false          | `1.0.0`        |

gr2m:
version numbers are for machines, not people. They are a contract between a library and its dependents. I'd strongly recommend to start at version 1.0.0, just like npm does when you use `npm init` by default.

You can create pre-releases as you suggest (`1.0.0-beta.1`, etc). And if you want users to know that your software is not yet complete, I would recommend to state so in the projects README

I don't see a need for yet another flag.

jcornaz:
>I'd strongly recommend to start at version 1.0.0

Thanks for your recommendation. I however would like to follow [semantic-versioning](https://semver.org) recommendation.

This tool is named "semantic-release" wouldn't it be nice if it supported semantic versioning?

At least maybe it could be documented that semantic versioning is not fully supported and why. That would have saved me hours of looking in the doc for the feature.

gr2m:
The meaning of `0.x` version is not clearly defined/understood. 

I agree could document that fact somewhere, if we haven't yet. I'm sorry you spent so much time because of this :(

travi:
there is an FAQ about this in the documentation: https://semantic-release.gitbook.io/semantic-release/support/faq#can-i-set-the-initial-release-version-of-my-package-to-0-0-1

jcornaz:
Thanks @travi for pointing the FAQ!

travi:
[...]

PR #1574, opened 2020-06-13
feat: Support staying on v0.x.y with --unstable

dcodeIO:
As a follow-up to https://github.com/semantic-release/semantic-release/issues/1507#issuecomment-643552928 this is an initial draft implementation to opt-in to keep versions in the `v0.x.y` range.

Context is that we are building what's essentially a [TS to Wasm compiler](https://github.com/AssemblyScript/assemblyscript), with the underlying tech being pretty much still in flux, so we'd prefer to keep the version below a `v1` while certain critical language features are not yet available and do a `v1` once the project is ready to be perceived as something reasonably finished. Yet we'd like to switch to using semantic-release.

Idea is to specify the `--unstable` option as long as a `v1` is not wanted, and remove it again once a `v1` is ready. As such there are no functional changes for existing users, and it's fully opt-in.

~~Coincidentally, I also fixed linting and testing on Windows by making Git checkout with `\n` instead of `\r\n` line endings.~~ addressed in https://github.com/semantic-release/semantic-release/pull/1575

Is this something you'd be willing to support officially? And yeah, from the existing issues I learned that this is a somewhat sensitive topic :)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf
--- a/cli.js
+++ b/cli.js
@@ -35,6 +35,7 @@ Usage:
     .option('fail', {...stringList, group: 'Plugins'})
     .option('debug', {describe: 'Output debugging information', type: 'boolean', group: 'Options'})
     .option('d', {alias: 'dry-run', describe: 'Skip publishing', type: 'boolean', group: 'Options'})
+    .option('unstable', {describe: 'Enforce staying on v0.x.y.', type: 'boolean', group: 'Options'})
     .option('h', {alias: 'help', group: 'Options'})
     .option('v', {alias: 'version', group: 'Options'})
     .strict(false)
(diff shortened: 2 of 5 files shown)


gr2m:
Thanks Daniel! I'll have to discuss this with Pierre, who is not available right now, so please give us a moment.

> Coincidentally, I also fixed linting and testing on Windows by making Git checkout with `\n` instead of `\r\n` line endings.

Could you create a separate PR for this, so that we can discuss it separately?

dcodeIO:
Synched again on top of master. Have been using the feature in production for a while now without any issues. Did you guys have a chance to discuss yet? :)

gr2m:
not yet, sorry.  Thank you for keeping the pull request up-to-date, it does look pretty good to me. Could you update docs as well?

gr2m:
We talked once more about the topic, but our conclusion remains that this is much more complicated than it looks. Adding this feature will likely create lots of more issues, and we are having a hard time supporting semantic-release as is today. 

Sorry :( 

We will not be implementing it in semantic-release. And we have an alternative solution that is supported today: using pre-release channels, see https://semantic-release.gitbook.io/semantic-release/support/faq#can-i-set-the-initial-release-version-of-my-package-to-0-0-1

dcodeIO:
Ok. But thanks for looking into it!

PR #1578, opened 2020-06-19
feat(bitbucket-basic-auth): support for bitbucket server basic auth

Sly321:
Bitbucket Servers currently only support basic auth, so it is not possible to authenticate with `x-auth-token:{token}`, you need to put your username before the token or use bearer headers.


gr2m:
Danke Sven, great PR 👍

Issue #1582, opened 2020-06-22
Add configuration support for final version string


techninja:
## New feature motivation

As standardized as semantic version strings are, I imagine not everyone supports every part of the standard. The ability to customize what `version` becomes would be incredibly handy. Specific to my needs is that on Drupal.org, beta versions are not allowed to have a period separating pre-release revisions. EG: Does not accept: `2.0.0-beta.2`, will accept `2.0.0-beta2`

## New feature description

Offer a `releaserc` configuration option to provide a template for all version components for both full releases and prereleases. EG:

```javascript
// .releaserc.js
module.exports = {
  versionFormat: '${major}.${minor}.${patch}',
  prereleaseVersionFormat: '${major}.${minor}.${patch}-${branch.prerelease}${prereleaseVersion}',
  tagFormat: 'v${version}',
  plugins: [...],
};
```

## New feature implementation

Near as I can tell [this is hard-coded](https://github.com/semantic-release/semantic-release/blob/beta/lib/get-next-version.js#L22) and the reasons behind that may very well deter this from simply being implemented as it's likely that reverse parsing and keeping to a specific format are incredibly important to knowing what's going on, and likely someone could configure a version string that's entirely non-standard and breaks everyone's expectations. 

Maybe what's actually needed is some kind of configuration for disabling beta/alpha periods 😆

travi:
i actually ran into a similar complication recently when considering creating a plugin for managing ruby gems. the reason i was looking into it has gone away, but i did look deep enough to learn that the [convention for pre-releases](https://guides.rubygems.org/patterns/#prerelease-gems) in that community does not follow the spec either (i'm not usually a ruby developer, so i hadn't run across this detail before). in this case, the issue was that `.` is used instead of a `-` to separate the pre-release version.

techninja:
Interesting! Well, I figure there's a need here there's definitely some opinions not shared across communities. I'd hate to fork just for this, and I'm sure there's lots of known use cases here.

gr2m:
Thank you James & Matt for sharing your use cases. I'm not sure what the implications would be in terms of added code complexity. If you consider to implement it anyway, I'd appreciate a pull request from your fork. It would be a great base of discussion. I cannot make promises whether we can merge it or not, semantic-release is hard to maintain as is, so we carefully weight the added complexity with the added maintenance overhead.

techninja:
[...]

PR #1611, opened 2020-07-31
Configurable pre-release version format

techninja:
Aimed at fixing my original issue #1582, Don't worry about dealing with this quickly, I'll be filling this PR out a bit more tomorrow.


gr2m:
This looks like it adds a lot of complexity, I won't be able to review that myself any time soon. Do you see any chance of reducing the impact of this change?

techninja:
@gr2m Sorry for the late reply, it's been a busy month. I think that this is pretty simple when it comes down to simply being able to customize what would eventually only be 1 or two characters difference in the output. Maybe it would be better to allow some kind of plugin to rewrite the final string as they see fit? Or perhaps I could cut this change down to the two use cases identified in the ticket with only change here being a configuration to customize only the separator for prereleases? I'm open to whatever you think will be simplest.

ccjjmartin:
As a note, this is currently only altering the prerelease version format for the first tag generated after enabling this patch, it will need to be expanded to search existing tags and increment version numbers from existing tags.  This is a great start though.

PR #1622, opened 2020-08-26
docs: improve the docs navigation

pcandido:
Links to the previous and next articles were added in all articles to improve the linear reading. Currently, the reader must back to the index to select the next article. After this change, the reader will jump to the next article by clicking in a link at the end of the page.

### DIFF
--- a/docs/developer-guide/js-api.md
+++ b/docs/developer-guide/js-api.md
@@ -282,3 +282,6 @@ Example:
    }
  ]
 ```
+
+---
+[<- Publishing pre-releases](../recipes/pre-releases.md) | [Plugins ->](plugin.md)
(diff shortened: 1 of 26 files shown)


gr2m:
Don't we have the navigation buttons at the bottom of every page already, e.g. at the bottom of https://semantic-release.gitbook.io/semantic-release/developer-guide/js-api?

image

pcandido:
Hum, It seems to be some website feature, those buttons are not available
while reading on GitHub or even on a checked-out repository. I will keep my
suggestion but feel free to disagree.

gr2m:
I fear accepting your PR would make double the links on the website. I'll close the PR, but thanks for offering the help, improvement to the documentation are the best PRs :)

Issue #1625, opened 2020-08-30
Docs on GitHub Actions recipe can be improved on @semantic-release/git plugin


JoepKockelkorn:
The recipe for GitHub Actions gives us a nice push in the right direction, but misses some key info. When using the `@semantic-release/git` plugin in combination with branch protection rules, it will fail to push tags/commits. The generated `GITHUB_TOKEN` interferes with the `GITHUB_TOKEN` env variable you need to expose to the shell (from Secrets). So in the workflow this is needed:

```yaml
- name: Checkout
     uses: actions/checkout@v2
       with:
         persist-credentials: false  # <--- this
         fetch-depth: 0
```

## Security

Even though [it is described as a security risk](https://github.com/semantic-release/semantic-release/blob/master/docs/recipes/github-actions.md#pushing-packagejson-changes-to-a-master-branch), you're gonna have to expose a custom Personal Access Token with push permission as the generated `GITHUB_TOKEN` only has minimum read-only permissions. I don't really agree with the Note in the docs, because you can perfectly restrict the repo to only let collaborators edit (read: run actions from any branches). [The GitHub secrets you entered are not available for any forked repo's that submit a PR to your repo](https://docs.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#using-encrypted-secrets-in-a-workflow).

Related issues:

- [semantic-release/git#196 (this is where I found the 'fix')](https://github.com/semantic-release/git/issues/196)
- https://github.com/semantic-release/github/issues/175

JoepKockelkorn:
Created a PR: https://github.com/semantic-release/semantic-release/pull/1626

PR #1626, opened 2020-08-30
docs: improve github actions recipe on git plugin

JoepKockelkorn:
This PR should add to the docs about the issue that is linked: fixes https://github.com/semantic-release/semantic-release/issues/1625.


docs/recipes/github-actions.md
```diff
@@ -50,6 +50,16 @@ To keep `package.json` updated in the `master` branch, [`@semantic-release/git`]
...
+If the risk is acceptible, some extra configuration is needed. The [actions/checkout `persist-credentials`](https://github.com/marketplace/actions/checkout#usage) option needs to be `false`, otherwise the generated `GITHUB_TOKEN` will interfere with the custom one. Example:
+
+```yaml
+- name: Checkout
+    uses: actions/checkout@v2
+    with:
+      fetch-depth: 0
+      persist-credentials: false # <--- this
+```
+
```
gr2m, line 62: I've set the `GITHUB_TOKEN` env variable to a personal access token for a git push in the past and I cannot recall ever running into problems.

The pull request is missing some context. Did you open an issue first? If not, could you point to a build that failed without this setting, and worked after you made the change?

JoepKockelkorn: Did you have branch protection enabled?

[This is the build where it fails](https://github.com/JoepKockelkorn/git-branch-picker/runs/1048217247?check_suite_focus=true), [and this is where it succeeds](https://github.com/JoepKockelkorn/git-branch-picker/runs/1048226894?check_suite_focus=true). I force-pushed the branch so it might not be clear what is the difference. Should I create a separate example where I showcase this?

JoepKockelkorn: I've [pushed a feat commit](https://github.com/JoepKockelkorn/git-branch-picker/runs/1049963033?check_suite_focus=true) to my repo which showcases that without the setting the publish step fails.

JoepKockelkorn:
@gr2m I've shown in my repo that is fails without the `persist-credentials` setting. Could you take another look?

gr2m (approved):
Thanks for the help, I appreciate you taking the time sharing the reference links.

PR #1639, opened 2020-09-17
fix: Add logging for when ssh falls back to http

Alphy11:
When setting up semantic release in CI, the debug logging does not provide enough context on why it is trying to push via https, and is hard to debug without deeply reading the code.  Adding some debug logging to indicate that will make it easier to know why you're calling git via https.


gr2m (approved):
thanks for digging through the code and making it easier for future users to identify the same problem!

PR #1644, opened 2020-09-30
feat!: support multiple git tags and publish major and minor version tag

### LINKED ISSUE #1515: Add ability to update a major tag
## New feature motivation

In order to support the [recomendation from GitHub Actions versioning guideline](https://github.com/actions/toolkit/blob/master/docs/action-versioning.md#recommendations):

> Make the new release available to those binding to the major version tag: Move the major version tag (v1, v2, etc.) to point to the ref of the current release. This will act as the stable release for that major version. You should keep this tag updated to the most recent stable minor/patch release.

But despite the main motivation being GitHub Actions, this can also be useful in others scenarios. 

## New feature description

Add the ability, probably a new configuration section, to define wheter a major tag should be updated or not. Such as:

``` json
"release": {
  "updateMajorTag": true
}
```

**Release branches** solves the issue partially, because it will still lack a git reference for the latest major version, since there won't be a release branch until the next major version comes. Also, in small projects, which might be the case for most of the GitHub Actions, creating release branches manually might not be necessary at all, so then a major git tag would be enough.

Currently, some projects such as the [cycjimmy/semantic-release-action](https://github.com/cycjimmy/semantic-release-action) itself, uses 
a [workaround](https://github.com/cycjimmy/semantic-release-action/blob/master/.github/workflows/release.yml#L45-L49): 

> ``` yaml
> - name: Push updates to branch for major version
>         if: steps.semantic.outputs.new_release_published == 'true'
>         run: "git push https://x-access-> 
> token:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git HEAD:refs/heads/v${{steps.semantic.outputs.new_release_major_version}}"
>         env:
>           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
> ```

Which updates the release branch automatically.

felipecrs:
This adds support for multiple git tags to be specified as input. The user
can now use three new variables: `major`, `minor`, and `patch`.

The default tags aren't only `v${version}` anymore, but `v${major}.${minor}`
and `v${major}`.

This new pattern helps to spread the good practice of creating a git tag
with the major version number so the users can quickly identify what's the
revision of the latest major (or minor) release.

BREAKING CHANGE: the CLI parameter `--tag-format` was renamed to
`--tag-formats` and the `gitTag` variable in the `release`s objects from the
JavaScript API was renamed to `gitTags`, with the `Array` type rather than
`String`. For those relying on the `gitTag` variable, you should now use
`gitTags[0]`.

Fixes #1515

### DIFF
--- a/docs/developer-guide/js-api.md
+++ b/docs/developer-guide/js-api.md
@@ -220,7 +220,7 @@ Information related to the newly published release:
 | type    | `String` | The [semver](https://semver.org) type of the release (`patch`, `minor` or `major`).                                           |
 | version | `String` | The version of the new release.                                                                                               |
 | gitHead | `String` | The sha of the last commit being part of the new release.                                                                     |
-| gitTag  | `String` | The [Git tag](https://git-scm.com/book/en/v2/Git-Basics-Tagging) associated with the new release.                             |
+| gitTags | `Array`  | The [Git tag](https://git-scm.com/book/en/v2/Git-Basics-Tagging)s associated with the new release.                             |
 | notes   | `String` | The release notes for the new release.                                                                                        |
 | channel | `String` | The distribution channel on which the next release will be made available (`undefined` for the default distribution channel). |
 
@@ -230,7 +230,7 @@ Example:
   type: 'minor',
   gitHead: '68eb2c4d778050b0701136ca129f837d7ed494d2',
   version: '1.1.0',
-  gitTag: 'v1.1.0',
+  gitTags: ['v1.1.0', 'v1.0', 'v1'],
   notes: 'Release notes for version 1.1.0...',
   channel : 'next'
 }
--- a/docs/usage/configuration.md
+++ b/docs/usage/configuration.md
@@ -88,15 +88,15 @@ The git repository URL.
 
 Any valid git url format is supported (See [Git protocols](https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols)).
 
-### tagFormat
+### tagFormats
 
-Type: `String`<br>
-Default: `v${version}`<br>
-CLI arguments: `-t`, `--tag-format`
+Type: `Array`<br>
+Default: `['v${version}', 'v${major}', 'v${major}.${minor}']`<br>
+CLI arguments: `-t`, `--tag-formats`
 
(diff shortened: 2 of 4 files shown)


felipecrs:
Hi maintainers, do you approve the design on the table (read the commit message)? Can I move ahead with the implementation? Or should we take a less intrusive approach, such as adding a new variable in the release object such as `additionalGitTags` in order to prevent the breaking change?

gr2m:
I feel like this shouldn't be part of `semantic-release` core, but rather live in a separate plugin. This adds a lot of complexity to the `semantic-release` core, I think it outweights the value

felipecrs:
> I feel like this shouldn't be part of semantic-release core, but rather live in a separate plugin.

I can buy that, however, my goal is to make it the default behavior. That would mean a new official plugin + including this new official plugin in the list of the default plugins. This sounds too complex, and since the `core` already deals with `git tag`s, so I wonder why not?

> This adds a lot of complexity to the semantic-release core

Not that much, we're just talking about basically changing one property from String type to Array type. Right?

gr2m:
You have been the first to ask for this feature as far as I can tell, I would definitely not add it to `semantic-release` core. 

> we're just talking about basically changing one property from String type to Array type

I fear this will have big implications across the plugin ecosystem

felipecrs:
> You have been the first to ask for this feature as far as I can tell, I would definitely not add it to semantic-release core.

Because others implemented their own workarounds, as stated in the Issue description.

> I fear this will have big implications across the plugin ecosystem

You're totally right. That's why I asked for the design review first. How about the proposed alternative:

> Or should we take a less intrusive approach, such as adding a new variable in the release object such as additionalGitTags in order to prevent the breaking change?

gr2m:
The less intrusive approach I prefer would be to create an external plugin, see if/how people are using it, and then come back to the discussion whether to make this a core feature or not

felipecrs:
You're right @gr2m. What do you think about a new optional option in `@semantic-release/git`?

gr2m:
yeah you can explore that if you prefer it over creating a separate plugin. Even if we decide to create a separate plugin in the end, it should be simple to transfer the code

PR #1648, opened 2020-10-02
docs(CONTRIBUTING): fix commit message examples

MicaelJarniac:
It had the backticks (`) around the messages, but I think they were unintentional, probably a leftover from copy-pasting.


gr2m (approved):
Thanks 👍🏼

PR #1667, opened 2020-10-23
docs: add npm-deprecate-old-versions in plugins list

ghusse:
Adds a new plugin in the list of additional plugins for `semantic-release`.

This plugin allows to automatically deprecate old versions as part of the release process.


gr2m (approved):
very cool!

Issue #1668, opened 2020-10-28
When multiple git credentials are available in environment, the first one is used regardless of the others


arthurchaloin:
## Current behavior

If both `GITHUB_TOKEN` and `GITLAB_TOKEN` are defined in the environment, semantic-release will use `GITHUB_TOKEN` to authenticate, even if the repository is hosted on Gitlab.

This is due to `get-git-auth-url` that stops on the first available token (using the order defined in [`lib/get-git-auth-url.js:19`](https://github.com/semantic-release/semantic-release/blob/d74ffef788df841a4dcb65dda4d857edd30b13c3/lib/get-git-auth-url.js#L19)).

This is most probably what is causing #1484.
I propose a fix in #1669.

## Expected behavior

Semantic release should look for all tokens and use the one that actually works.

## Environment

- **semantic-release** version: 17.2.1, `master`
- CI environment: any
- Plugins used: not relevant
- **semantic-release** configuration:
```json
{
  "plugins": [
    "@semantic-release/commit-analyzer",
    "@semantic-release/release-notes-generator",
    [
      "@semantic-release/gitlab",
      { "gitlabUrl": "https://gitlab.example.com" }
    ]
  ]
}
```
- CI logs:
```
$ echo GITHUB_TOKEN
[MASKED]
$ echo GITLAB_TOKEN
[MASKED]
$ git push --dry-run "https://gitlab-ci-token:$GITLAB_TOKEN@gitlab.example.com/group/repo.git" HEAD:master
Everything up-to-date
$ semantic-release --ci --debug
[12:23:39 PM] [semantic-release] › ℹ  Running semantic-release version 17.2.1
2020-10-27T12:23:39.657Z semantic-release:config load config from: /builds/group/repo/.releaserc.json
2020-10-27T12:23:39.675Z semantic-release:config options values: { branches:
   [ '+([0-9])?(.{+([0-9]),x}).x',
     'master',
     'next',
     'next-major',
     { name: 'beta', prerelease: true },
     { name: 'alpha', prerelease: true } ],
  repositoryUrl:
   'https://gitlab-ci-token:[secure]@gitlab.example.com/group/repo.git',
  tagFormat: 'v${version}',
  plugins:
   [ '@semantic-release/commit-analyzer',
     '@semantic-release/release-notes-generator',
     [ '@semantic-release/gitlab', [Object] ] ],
  _: [],
  ci: true,
  debug: true,
  '$0': '/usr/local/bin/semantic-release' }
[...]

PR #1669, opened 2020-10-28
Use valid git credentials when multiple are provided

arthurchaloin:
See #1668 for details about the issue. Here I will discuss about the fix implementation.

If the given URL works out of the box or if a single token is available in the environment, the behaviour remains unchanged.
However if there is multiple tokens available, it will perform a `git.verifyAuth` call for each of them, and return the first of them that is working.
If none of them are working, it will return the URL as is.

This behaviour is tested in `integration.test.js`. It isn't tested in `get-git-auth-url.test.js` because I need gitbox functionalities to test it.

### DIFF
--- a/test/integration.test.js
+++ b/test/integration.test.js
@@ -678,7 +678,7 @@ test('Use the valid git credentials when multiple are provided', async (t) => {
   );
 });
 
-test('Use the reposirotry URL as is if none of the given git credentials are valid', async (t) => {
+test('Use the repository URL as is if none of the given git credentials are valid', async (t) => {
   const {cwd} = await gitbox.createRepo('test-invalid-auth');
   const dummyUrl = 'http://toto@localhost:2080/git/test-auth.git';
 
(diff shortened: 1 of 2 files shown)


test/integration.test.js
```diff
@@ -656,3 +657,43 @@ test('Hide sensitive environment variable values from the logs', async (t) => {
...
+        GIT_TERMINAL_PROMPT: 0,
+      },
+      branch: {name: 'master'},
+      options: {repositoryUrl: 'http://toto@localhost:2080/git/test-auth.git'},
+    }),
+    authUrl
+  );
+});
+
+test('Use the reposirotry URL as is if none of the given git credentials are valid', async (t) => {
```
gr2m, line 681:
```suggestion
test('Use the repository URL as is if none of the given git credentials are valid', async (t) => {
```

gr2m (commented):
Great work! Just a small typo. I would have fixed it myself but don't have write access to your fork

arthurchaloin:
Fixed the typo and the conflicts, ready to merge!

gr2m (approved):
Thanks 👍🏼

PR #1693, opened 2020-11-18
build(deps): update @semantic-release/github to v7.2.0

wingyplus:


travi:
@wingyplus could you help us understand why this change would be helpful? we normally don't tighten the range for in-range dependencies unless there is a new feature added that we specifically need to use.

if you are simply wanting to get the latest version in a project where you have a lockfile, try uninstalling semantic-release and reinstalling in order to update the dependency sub-tree under semantic-release (or completely remove the lockfile and regenerate, if you prefer going that far)

wingyplus:
Sorry, I didn't explain it in details. `github@v0.7.2` introduced set target branch when you publishing a release. The previous version doesn't set it when create a release which's fallback to default branch that you set in repository settings.

wingyplus:
> if you are simply wanting to get the latest version in a project where you have a lockfile, try uninstalling semantic-release and reinstalling in order to update the dependency sub-tree under semantic-release (or completely remove the lockfile and regenerate, if you prefer going that far)

Oh, let me try. If it's work, then this PR will not necessarily accept. :)

wingyplus:
@travi It seems that's work out of the box. So I'll close this. Thanks for your help

travi:
glad that worked out for you. thank you for following up to let us know

PR #1707, opened 2020-12-02
docs: add semantic-release-pypi plugin

abichinger:
documentation for semantic-release-pypi plugin


gr2m (approved):
Awesome, thank you!

If you like, you can also submit a tweet so we can probot your plugin from https://mobile.twitter.com/SemanticRelease:
https://github.com/semantic-release/twitter-together/

PR #1721, opened 2020-12-28
docs(plugins): document `addChannel` step

cherryblossom000:
Document the `addChannel` step defined in https://github.com/semantic-release/semantic-release/blob/e35e5bb73f3628906553709ac0f5beb8f565c61a/lib/definitions/plugins.js#L83-L84

This step is used in [@semantic-release/npm](https://github.com/semantic-release/npm).


gr2m (approved):
Thanks!

PR #1731, opened 2021-01-12
docs: clarify node vs npm in FAQ

vegardok:
> If you wish to only update the `package.json` and push via Git you can set the project to `"private": true,` within your `package.json` to prevent publishing to [npm](https://www.npmjs.com). This can be useful for using **semantic-release** with a non-node project.

This is still describing a node-project.

### DIFF
--- a/docs/support/FAQ.md
+++ b/docs/support/FAQ.md
@@ -8,7 +8,7 @@ By default, only the published package will contain the version, which is the on
 
 However, the [`@semantic-release/git`](https://github.com/semantic-release/git) plugin can be used to push the updated `package.json` as well as other files to the Git repository.
 
-If you wish to only update the `package.json` and push via Git you can set the project to `"private": true,` within your `package.json` to prevent publishing to [npm](https://www.npmjs.com). This can be useful for using **semantic-release** with a non-npm project.
+If you wish to only update the `package.json` and push via Git you can set the project to `"private": true,` within your `package.json` to prevent publishing to [the npm registry](https://www.npmjs.com).
 
 ## How can I use a npm build script that requires the `package.json`’s version ?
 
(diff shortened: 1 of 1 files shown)


docs/support/FAQ.md
```diff
@@ -8,7 +8,7 @@ By default, only the published package will contain the version, which is the on
 
 However, the [`@semantic-release/git`](https://github.com/semantic-release/git) plugin can be used to push the updated `package.json` as well as other files to the Git repository.
 
-If you wish to only update the `package.json` and push via Git you can set the project to `"private": true,` within your `package.json` to prevent publishing to [npm](https://www.npmjs.com). This can be useful for using **semantic-release** with a non-node project.
+If you wish to only update the `package.json` and push via Git you can set the project to `"private": true,` within your `package.json` to prevent publishing to [npm](https://www.npmjs.com). This can be useful for using **semantic-release** with a non-npm project.
```
gr2m, line 11: I think `non-node` is correct here. There are alternatives to the `npm` CLI such as `yarn`, but both are Node projects. The `package.json` file is a node convention, not an npm convention

gr2m: we could rephrase 

> within your `package.json` to prevent publishing to [npm](https://www.npmjs.com)

to 

> within your `package.json` to prevent publishing to [the npm registry](https://www.npmjs.com)

to make it a bit more clear?

vegardok: > The package.json file is a node convention, not an npm convention

Yes, exactly. The paragraph describes using package.json with `{ ..., "private": true, ...}` so it's still definitely a node project, not a non-node project :).

My diffs intention was to differentiate between published to npm/npmjs.com vs unpublished node projects (not cli-1 vs cli-2).



> >    within your package.json to prevent publishing to the npm registry

> to make it a bit more clear?

Only if you remove the last sentence completely. I think the paragraph will be fine without it and using sematic-versioning in actual non-node projects is documented below.

gr2m (approved):
great change, thank you!

PR #1736, opened 2021-01-15
fix: add logging for top-level errors

### LINKED ISSUE #1734: Exceptions in bin/semantic-release are swallowed silently
We just ran into a failing CircleCI pipeline, which was in the end caused by https://github.com/semantic-release/semantic-release/issues/1694 as it turned out.
Really tough one to figure out, as we did not change anything in the semantic-release set up, but we apparently added a new CircleCI job context, which contained an env variable value containing a `(` character, which triggered https://github.com/semantic-release/semantic-release/issues/1694.

However, it took way longer than necessary to find this root cause, as semantic-release was failing with exit code 1, but not printing **any** output to the console (not even with `--debug`).

In our case, it would probably have helped to see the exception that is silently swallowed here:

https://github.com/semantic-release/semantic-release/blob/21c151f167960ffa9f43069bc109900ebf57aed3/bin/semantic-release.js#L43

tzimmermann:
Log errors in the main binary to the console to help with debugging.
I wasn't sure how to add tests for this, as I couldn't find any existing tests for that file.

closes #1734

### DIFF
--- a/bin/semantic-release.js
+++ b/bin/semantic-release.js
@@ -40,6 +40,7 @@ require('../cli')()
   .then((exitCode) => {
     process.exitCode = exitCode;
   })
-  .catch(() => {
+  .catch((error) => {
+    console.error(error);
     process.exitCode = 1;
   });


gr2m (approved):
>  I wasn't sure how to add tests for this, as I couldn't find any existing tests for that file.

that's okay. Logging the error is better than not logging the error at all. We can still figure out tests later

Queens666:
https://github.com/semantic-release/semantic-release/pull/2211#issue-1037705681

PR #1739, opened 2021-01-19
fix(deps): update to yargs 16.2.0

rolego:
fixes vulnerability in y18n which is a dependency of yargs version 15:
https://nvd.nist.gov/vuln/detail/CVE-2020-7774


gr2m (approved):
Thanks!

PR #1740, opened 2021-01-20
fix: pin initial branch to master

jsoref:
> Do not throw error when unshallow a complete repository

will fail if one uses a new enough git and has set `init.defaultBranch` (mine is set to `main`).


gr2m (approved):
Thanks!

PR #1763, opened 2021-01-26
build(engine-strict): strictly enforced node version compatibility for dependencies

travi:
to avoid updating to a dependency version that does not support our supported range

i know we've been attempting to avoid config files in the root of the projects where possible, but i'm not sure if there is a good way to accomplish this same goal in another way. i'm open to thoughts around whether we are ok proceeding in this way or if we should consider an alternative.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,6 @@ jobs:
       matrix:
         node-version:
           - 10.18.0
-          - 10
           - 12
           - 14
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,6 +17,7 @@ jobs:
     strategy:
       matrix:
         node-version:
+          - 10.18.0
           - 10
           - 12
           - 14
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,6 +31,8 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: ${{ matrix.node-version }}
+      - name: Ensure dependencies are compatible with the version of node
+        run: echo 'engine-strict=true' >> .npmrc
       - uses: bahmutov/npm-install@v1
       - run: npm run test:ci
 
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,0 @@
-engine-strict=true
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+engine-strict=true


travi:
if we decide to continue in this direction, this may be a good excuse for me to learn to write an octoherd script for applying this same change to the other projects in the org

.github/workflows/test.yml
```diff
@@ -17,6 +17,7 @@ jobs:
     strategy:
       matrix:
         node-version:
+          - 10.18.0
           - 10
```
gr2m, line 21: do we need both? Maybe testing with the lowest version we officially support will suffice?

travi: I could be convinced either way. It's possible for a change to be introduced in a later version that could cause a problem, but probably a bug in nice if it's breaking.

gr2m: I'd just add `10.18.0` and trust the node folks that they don't break anything in maintenance version. It's very unlikely. It will speed up our tests, as GitHub Actions limit the amount of parallel builds

travi: i pulled `10` out of the list

gr2m: I still see it?

https://github.com/semantic-release/semantic-release/blob/engine-strict/.github/workflows/test.yml#L21

travi: somehow managed to not push the change 🤦🏼 should be there now

.npmrc
```diff
@@ -0,0 +1 @@
+engine-strict=true
```
gr2m, line 1: I wonder if we could add a flag when running npm install as part of the CI instead of adding a dotfile? I'm just trying to avoid dotfiles if we can help it. New contributors will wonder what it is for, it adds clutter. The benefit might outweight the cost here, but there still is a cost, and I wonder if there is an alternative way to achieve the same thing?

travi: yeah, this one isnt as clear cut since the install command is hidden by the action. i'm open to ideas if you have thoughts.

i think the best alternative i can come up with would be to write to the file from the ci steps. would something along the lines of `echo 'engine-strict=true' >> .npmrc` before the install step be a reasonable middle ground?

gr2m: > i think the best alternative i can come up with would be to write to the file from the ci steps. would something along the lines of `echo 'engine-strict=true' >> .npmrc` before the install step be a reasonable middle ground?

let's do that, I like it

travi: updated to limit the check to the ci workflow

gr2m (approved):

PR #1808, opened 2021-02-17
docs: add main as push branch in sample workflow

marcprux:


gr2m:
we willl change that to `main` once we switch our repository's default branch. Until then, the change does not have any effect

Issue #1814, opened 2021-02-22
Support *.cjs configs


opiation:
## New feature motivation

When attempting to use `semantic-release` with a ESM-written package (`"type": "module"`) and node with ES modules enabled, a plain `release.config.js` config fails understandably given the tool is written with `require()`s.  Renaming the config to `release.config.cjs` does not work as the version of `cosmicconfig` used by `semantic-release` does not search for `.cjs` files.  Thus, only the `json` config options are available.

Updating the `cosmicconfig` dependency to v7.0.0 would allow one to use `semantic-release` developing a package implicitly in ES modules (`--experimental-modules`).

## New feature description

Supports `*.cjs` file extensions like `release.config.cjs` for configuring `semantic-release`.

## New feature implementation

Update `cosmicconfig` dependency to v7.0.0 which [adds support for `*.cjs` file extensions](https://github.com/davidtheclark/cosmiconfig/blob/main/CHANGELOG.md#700).

travi:
please double check the version of `semantic-release` that you are using. the current version of `cosmiconfic` that is depended on by `semantic-release` is already `v7.0.0`:

https://github.com/semantic-release/semantic-release/blob/6013a5633ecb71aac80f7b68b8e7250c5c58f7c0/package.json#L29

opiation:
My apologies, @travi.  I could have sworn that I looked at `master` but must have still been looking at the v17.3.0 tag.  You're correct that `cosmicconfig` was indeed updated recently ([v17.3.6](https://github.com/semantic-release/semantic-release/releases/tag/v17.3.6)).  I've updated my local version to use v17.3.9 however, `.cjs` still is not read.

My config:

```js
// release.config.cjs

const releaseRules = [
  { type: 'docs', release: 'patch' }
]

const gitOptions = {
  assets: ['CHANGELOG.md', 'dist/**/*', 'package.json']
}

module.exports = {
  plugins: [
    ['@semantic-release/commit-analyzer', { releaseRules }],
    '@semantic-release/release-notes-generator',
    '@semantic-release/changelog',
    '@semantic-release/npm',
    ['@semantic-release/git', gitOptions]
  ]
}
```


```sh
$ yarn run semantic-release --debug --dry-run

[11:22:25 AM] [semantic-release] › ℹ  Running semantic-release version 17.3.9                                                                                                                  
  semantic-release:config load config from: undefined +0ms 
```

[...]

PR #1815, opened 2021-02-22
feat(config): Use cosmiconfig defaults to support .cjs config files

opiation:
This change adds support for `*.cjs` config files by removing the explicit use of `searchPlaces` options and relying and the default search places generated by `cosmicconfig`.  As per the docs for [`cosmicconfig`][cc], the defaults include all the extensions/formats previously supported by semantic-release in addition to the new .cjs variants.

Resolves [#1814][issue].

[issue]: https://github.com/semantic-release/semantic-release/issues/1814
[cc]: https://github.com/davidtheclark/cosmiconfig#searchplaces

### DIFF
--- a/lib/get-config.js
+++ b/lib/get-config.js
@@ -9,19 +9,10 @@ const plugins = require('./plugins');
 const {validatePlugin, parseConfig} = require('./plugins/utils');
 
 const CONFIG_NAME = 'release';
-const CONFIG_FILES = [
-  'package.json',
-  `.${CONFIG_NAME}rc`,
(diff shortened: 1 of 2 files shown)


travi (approved):
this looks good to me 👍🏼

travi:
thank you for investigating and tracking down these details. i did dig a little bit to try to understand why they were defined explicitly before, but even the version that resulted in the addition appears to have supported similar defaults. from what i can tell this change aligns with the original intent fro before that change too.

PR #1836, opened 2021-03-08
feat: use `conventionalcommits` as the default preset

### LINKED ISSUE #1652: Switch default preset from angular to conventionalcommits
I believe Conventional Commits has grown as a more general pattern enough to now become the default preset for semantic-release plugins, over the angular one.

### LINKED ISSUE #3406: Make Conventional Commits the default preset
### New feature motivation

Hello,

First of all, I would like to thank you for such an amazing tool.

I was curious whether it would make sense to make the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) the default preset.

The Angular preset does not allow for marking breaking changes in a commit message title. It's a major drawback. Also I think the community is more aware of the conventional commit style rather than Angular commit style.

There has been some discussion in [this PR](https://github.com/semantic-release/commit-analyzer/issues/231#issuecomment-2127394159) to make the Conventional Commits the default, so I thought it would be good to bring it up in a separate issue.

Thank you for considering this.

### New feature description

Change the default preset to conventionalcommits.

### New feature implementation

_No response_

felipecrs:
Nowadays the `conventionalcommits` is the closest thing we have that looks like a standard and neutral.

Also, a new cli option was added: `--preset`.

BREAKING CHANGE: Previously, the default preset was `angular`. Despite the similarities, the new one has some differences, such as accepting exclamation marks (e.g. `feat!:`) for identifying breaking changes.  To restore the previous behavior, use `preset: 'angular'` or the new cli option, `--preset angular`.

Closes #1652
Closes #3406

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "@semantic-release/npm": "^13.1.1",
     "@semantic-release/release-notes-generator": "^14.1.0",
     "aggregate-error": "^5.0.0",
+    "conventional-changelog-conventionalcommits": "^9.0.0",
     "cosmiconfig": "^9.0.0",
     "debug": "^4.0.0",
     "env-ci": "^11.0.0",
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
     <img alt="OpenSSF Scorecard" src="https://api.securityscorecards.dev/projects/github.com/semantic-release/semantic-release/badge">
   </a>
   <a href="#badge">
-    <img alt="semantic-release: angular" src="https://img.shields.io/badge/semantic--release-angular-e10079?logo=semantic-release">
+    <img alt="semantic-release: conventionalcommits" src="https://img.shields.io/badge/semantic--release-conventionalcommits-e10079?logo=semantic-release">
   </a>
 </p>
 <p align="center">
@@ -50,7 +50,7 @@ This removes the immediate connection between human emotions and version numbers
 **semantic-release** uses the commit messages to determine the consumer impact of changes in the codebase.
 Following formalized conventions for commit messages, **semantic-release** automatically determines the next [semantic version](https://semver.org) number, generates a changelog and publishes the release.
 
-By default, **semantic-release** uses [Angular Commit Message Conventions](https://github.com/angular/angular/blob/main/contributing-docs/commit-message-guidelines.md).
+By default, **semantic-release** uses the [Conventional Commits convention](https://www.conventionalcommits.org/).
 The commit message format can be changed with the [`preset` or `config` options](https://semantic-release.org/usage/configuration/#options) of the [@semantic-release/commit-analyzer](https://github.com/semantic-release/commit-analyzer#options) and [@semantic-release/release-notes-generator](https://github.com/semantic-release/release-notes-generator#options) plugins.
 
 Tools such as [commitizen](https://github.com/commitizen/cz-cli) or [commitlint](https://github.com/conventional-changelog/commitlint) can be used to help contributors and enforce valid commit messages.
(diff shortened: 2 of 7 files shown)


travi:
I'm supportive of this change, since I do think it makes more sense as the default now that things have evolved over time. 

@semantic-release/maintainers, if you agree, what thoughts do you have around timing of introducing a breaking change like this? do we think the differences are enough to attempt more communication than would be included in the release notes of the major version bump?

gr2m:
I'm in favor, too. I don't know if we have other breaking changes lined up, I'd suggest we do an audit, and assign issues we want to ship with the upcoming breaking version to a milestone.

But when in doubt, I'd rather ship it than waiting too long, it's an overdue change.

atanunq:
Hello! I hope you don't mind me jumping in but are there plans to merge these changes? I am looking forward to setting up a semantic-release pipeline using only command line flags in a Makefile and this would come in very handy.

Your work is greatly appreciated!

gr2m:
I think so, eventually. We have other priorities right now. Sorry it takes so long

felipecrs:
I think this one can be included in 18.0 :)

gr2m:
> I think this one can be included in 18.0 :)

probably not in 18 because we want to get it out of the door quickly and are focusing on maintainability, but we can prioritize it for 19, which we can release soon after, together with some other priority changes. Stay tuned

felipecrs:
I rebased the PR, just in case.

travi:
> Can this be included in v19?

this is still on our radar and is a change that i personally want to see move forward. however, v19 is targeting resolution of a few vulnerabilities and we don't want to delay it with verifying other changes or making acceptance of the vulnerability fixes more risky for consumers by including more changes than absolutely necessary.

felipecrs:
Well, that was said for v18 as well. 😞

travi:
> Perhaps it won't cause any noticeable delay if you give it a chance to review?

more importantly, we don't want to add the surprise of switching conventions to consumers that are upgrading to get the vulnerability fix.

travi:
> I really think no changes will be required at all

also, we want to be sure that we are careful with this change so that release notes and other parts of the release process align properly without unexpected surprises. this convention has been the default from the beginning, so it would not be surprising if there are assumptions somewhere that we need to find and adjust.

felipecrs:
Ok, no problem. Please let me know if there is anything I can do to help.

BTW, for most people migrating, there won't be any required changes to their workflow. They will just have to have in mind the new things that conventionalcommits supports (AFAIK it does not drop any feature from Angular, but adds on it). Like the `!` for denoting breaking changes.

felipecrs:
I've refreshed this PR, and I think it's in good shape.

Note I've kept conventional-changelog-conventionalcommits at v9 because of https://github.com/conventional-changelog/conventional-changelog/issues/1495.

package.json
```diff
@@ -33,6 +33,7 @@
     "@semantic-release/npm": "^13.1.1",
     "@semantic-release/release-notes-generator": "^14.1.0",
     "aggregate-error": "^5.0.0",
+    "conventional-changelog-conventionalcommits": "^9.0.0",
```
armano2, line 36: q: should't we remove `conventional-changelog-angular` from dependencies?

felipecrs: `conventional-changelog-angular` is not part of the dependencies. It comes transitively via release-notes-generator and commit-analyser.

babblebey (commented):
👋🏾 @felipecrs 

Manually setting a preset directly in the `semantic-release` package isn't the way we are approaching this. We have never done this and wouldn't want to go that route now.

We are instead subscribed (sticking) to setting the default across board which means multiple official semantic-release plugin has to be coordinated in this effort starting from the `commit-analyzer` and that step as begun with the PR - https://github.com/semantic-release/commit-analyzer/pull/858

We sure recognize the `conventional-changelog` issue - https://github.com/conventional-changelog/conventional-changelog/issues/1495 and as much as we are working on this, we are also open to efforts that can help us move forward via this front.

felipecrs:
Fair enough.

In that case, what remains from this PR is the introduction of the `--preset` CLI flag as well as updating the badges in documentation.

I'll close this PR for now, and me/someone else can propose the improvements above if desired.

Thanks for the feedback!

Issue #1849, opened 2021-03-15
Allow semantic-release to run if local branch is behind remote


agaudreault-jive:
## New feature motivation

When semantic release is added to run automatically on CI for a project with a lot of contributions, the remote branch is quite often outdated and semantic-release will fail to create releases with `The local branch master is behind the remote one, therefore a new version won't be published.`. 

A master build for a small micro-service may take about 10 minutes if you add up the time to build, test, run static code analysis, etc. If at least one PR is merged during that timeframe, the release executed at the end of the build will fail.

The larger the repository, the longer the build may take, and the more contribution it will have, making the situation worse.

Having the build fail during the release step is not convenient for continuous delivery workflow where the master branch is deployed in production on each successful release. Having semantic-release failures because the branch is behind the remote cause the next release to contain more changes and be more impactful.

By merging the changes it can complicate versioning for downstream users as well as make reverting things unnecessarily more complex at the commit level rather than tag level.

## New feature description

Semantic-release should let the user decide if it is problematic to publish a release when the local branch is behind the remote. 
In that case, it is the user responsibility to make sure of the following:

- Plugins that push changes such as `@semantic-release/git` will fail if they are used.
- Concurrent executions are not possible to avoid conflict during releases.

The following changes should be implemented:

- [x] Add a parameter to allow semantic-release to run even if the local branch is behind remote

## New feature implementation

- [x] Validate that the current branch head is not a local commit
- [x] Validate that user has git push permission even if branch is behind remote
- [ ] Add documentation and warning for concurrent/outdated executions
- [ ] Add documentation and warning for plugins that push new commits

Related to https://github.com/semantic-release/semantic-release/issues/1208

travi:
The referenced thread very thoroughly explains why this approach is currently not supported. Adding more configuration options not only makes it easier for our users to end up in unexpected situations, but also adds maintenance and support burden to our team.   

We don't typically recommend applying semver to applications anyway, but worst case, your dependent will succeed once your builds catch up.

agaudreault-jive:
> We don't typically recommend applying semver to applications anyway, but worst case, your dependent will succeed once your builds catch up.

I think the referenced issue already explain that this is not convenient in a continuous delivery workflow where you want your releases to be as small as possible so you do not impact production too much.

`Semantic-release` has awesome features:

- It enforces semantic versioning
- It updates each PR and indicates which version they are part of
- It uses conventional commits to compute the next version
- It creates github tags and release notes

[...]

PR #1850, opened 2021-03-15
feat: add option to run when branch is behind remote

agaudreault-jive:
There is probably a better name that can be used instead of `isRemoteHead` and `--allow-outdated-branch`.

- [x] Validate that the current branch head is not a local commit. **Seems like this could be a breaking change, but in my opinion, it makes no sense that it was allowed.**
- [x] Validate that user has git push permission even if branch is behind remote.
- [x] Validate that the remote commits not on the local branch hasn't been tagged by other semantic-release executions (`git tag --no-merged <branch>` == `git tag --no-merged origin/<branch>`)
- [ ] Add documentation and warning for concurrent/outdated executions
- [ ] Add documentation and warning for plugins that push new commits

Closes https://github.com/semantic-release/semantic-release/issues/1849
Related to https://github.com/semantic-release/semantic-release/issues/1208


jeffrichie:
@agaudreault-jive I am stoked about this feature!

agaudreault-jive:
@jeffrichie we are running this fork on our CI for quite some times, it works the way it is intended to! But the maintainers does not seem thrilled to the idea even though many people need it. You are welcome to try it and let me know if something doesn't work!

jeffrichie:
@agaudreault-jive question for you - we've forked your branch and are using it locally. Do we run it like `npx @ourOrg/semantic-release --allow-outdated-branch true` ?

agaudreault-jive:
> @agaudreault-jive question for you - we've forked your branch and are using it locally. Do we run it like `npx @ourOrg/semantic-release --allow-outdated-branch true` ?

@jeffrichie It's a CLI flag so I would assume you only need to add `--allow-outdated-branch` without the boolean. We have it installed in a docker container and that is how we execute it. I don't know about npx much. If you have semantic release configs, you can add `"allowOutdatedBranch": true`

tmeijn:
Hey @agaudreault-jive do you still actively use this fork? I am using it happily, wondering if you could update it with the latest changes from `master`.

agaudreault-jive:
> Hey @agaudreault-jive do you still actively use this fork? I am using it happily, wondering if you could update it with the latest changes from `master`.

@tmeijn it is still used and we haven't had any problems with it in a long time. I might not have the time soon to work on the upgrade and test it again, but if you open a PR against `agaudreault-jive:allow-outdated-head` i can definitely take the time to review and merge it. Similar to https://github.com/agaudreault-jive/semantic-release/pull/1

tmeijn:
> @tmeijn it is still used and we haven't had any problems with it in a long time. 

Jep, same for over 1,5 years now using your fork, such a shame this will not merge into the mainline, unless... @pvdlg, @travi? 🙏🏾 

> I might not have the time soon to work on the upgrade and test it again, but if you open a PR against agaudreault-jive:allow-outdated-head i can definitely take the time to review and merge it. Similar to https://github.com/agaudreault-jive/semantic-release/pull/1

I'll try to do that if I find some time! Tried to do it already, but has a lot of conflicts: https://github.com/agaudreault-jive/semantic-release/pull/2

tmeijn:
Hey @agaudreault-jive would you be able to rebase this branch onto the latest changes in `master`? 🙏🏾  AFAIK it should be a simple press of the rebase/merge button.

In general: I use your fork for production workflows and I would like it for it be fairly up to date with `master`. Do you mind if I nudge you every once in a while (say every two months) to press the rebase/merge button? If you any other suggestions please LMK!

travi:
wanted to provide an update here since there has continued to be activity around the potential of this behavior being a part of semantic-release. this type of feedback really is helpful to help us determine where we direct the attention we have available to invest into the project. 

however, in this case, our answer remains that this is working as designed and we do not have plans to change our core implementation to support this capability. most of the reasoning for this stance already exists in the linked issues, but to summarize here:
* the release is not blocked, it simply happens on the next execution instead of the current one
* we need to account for plugins, official or not, that create commits as part of the release process, so it is not safe to assume that being behind the remote does not miss details of the current release
* while this flag might account for the use cases of some that avoid some of the complexities we are concerned about, we would need to account for those other complexities too before truly supporting this behavior officially. we believe that complexity goes beyond the current contribution and could be a significant effort to get right. even after the initial investment, that also burdens us with maintenance of complexity long term for use cases that we dont consider to be in scope of our encouraged use cases
* there are some use cases where this situation becomes even more complex when attempting to use semantic-release in the context of a monorepo, which is not officially supported. if the release of each package in the monorepo uses a plugin that creates commits as part of the release, this situation arrises for the next package in that monorepo. quite honestly, this is in the category of _why_ semantic-release does not officially support monorepos. semantic-release makes certain assumptions based on the idea that it is executing in the context of a single, isolated project. we dont want to add complexity to our core for situations like this that dont exist for the most part in intended use cases.

i'm going to close this PR based on this. we do appreciate the contribution of the suggestion, but in this case we are not going to move forward with it at this time. i would also ask that those using this branch as a fork for their workflows find a different place to maintain your fork. we're open to linking from our docs to such a fork published as a package with its own name. for now, i will leave the branch open to enable creating a more separate fork, but cannot promise that it will remain forever when eventually we clean up stale branches.

agaudreault:
I will keep my fork for anyone that wants it. However, i am now using https://github.com/intuit/auto and it seems to allow me to have the same behavior this fork is for. It is the desired behavior of the tool, and is more suitable to use for non js projects.

For my use cases, it has the same or similar capabilities that I needed, while being easier to integrate into a CI.

PR #1855, opened 2021-03-23
Update resources.md

ragoob:
A new article to explain semantic and  how to use it on gitLab pipeline


ragoob:
docs(recipes): semantic explanation and how to use it in GitLab

PR #1871, opened 2021-04-06
feat: add optional annotated or signed tags

DiogoAbu:
This adds the option to annotate or sign the tag.
This is opt-in, the default values will result in the current implementation.

To make an unsigned, annotated tag:
```json
{
  "tagAnnotate": true
}
```

To make an signed tag:
```json
{
  "tagSign": true
}
```
*Or by setting the git config `tag.gpgsign` to true.*

To customize the tag message:
```json
{
  "tagMessage": "release ${nextRelease.version}"
}
```

The message will be applied to the tag only if one of the following is `true`: `tagAnnotate`, `tagSign` or git config `tag.gpgsign`. Otherwise the tag is considered lightweight, which is the current implementation.

Also, `tagSign` takes precedence over `tagAnnotate`. The git config `tag.gpgsign` takes precedence over both.


DiogoAbu:
#1589 #1192 #214

t1m0thyj:
Hi, may I ask what's the status of this PR? We are also missing this feature, and adding it will be highly appreciated 🙏

gr2m:
sorry for not getting back to it. Both @travi and I are too occupied right now, but we are discussing how to make more time for @semantic-release. Please give us some more time

stuart-warren:
tools like miniver currently require annotated tags

https://github.com/jbweston/miniver

DanySK:
I need annotated tags as well

DanySK:
Meanwhile, I concocted a *terrible* workaround to be able to get annotated tags:

```yaml
tagFormat: "${version}"
plugins:
  - "@semantic-release/commit-analyzer"
  - "@semantic-release/release-notes-generator"
  - "@semantic-release/changelog"
  -
    - "@semantic-release/exec"
    - publishCmd: |
        git tag -a -f ${nextRelease.version} ${nextRelease.version} -F CHANGELOG.md
        git push --force origin ${nextRelease.version}
        # commands that need an annotated tag to work
  - "@semantic-release/github"
  - "@semantic-release/git"
```

DiogoAbu:
As a alternative to @DanySK...
I use patch-package with this file passing `-s` to make it signed and `-m` to add the required message:

`semantic-release+18.0.0.patch`
````diff
diff --git a/node_modules/semantic-release/lib/git.js b/node_modules/semantic-release/lib/git.js
index 8aaf884..53a610f 100644
--- a/node_modules/semantic-release/lib/git.js
+++ b/node_modules/semantic-release/lib/git.js
@@ -221,7 +221,7 @@ async function verifyAuth(repositoryUrl, branch, execaOptions) {
  * @throws {Error} if the tag creation failed.
  */
 async function tag(tagName, ref, execaOptions) {
-  await execa('git', ['tag', tagName, ref], execaOptions);
+  await execa('git', ['tag', tagName, ref, '-s', '-m', `release ${tagName}`], execaOptions);
 }
 
 /**
````

**Comment by DanySK** [comment]
Still, it would be nice to see your PR merged

**Comment by eladchen** [comment]
@DanySK - While you wait, I recommend you check out https://github.com/abstracter-io/atomic-release

You can create a custom command that creates an annotated tag instead of a lightweight

**Comment by lsagetlethias** [comment]
Hello, what is the status of this PR?

**Comment by DanySK** [comment]
I'm also interested to understand what is holding this back, it's a major hassle for me at the moment

**Comment by Langstra** [comment]
We ran into an issue that looked like this, but in the end turned out to be something slightly different. Semantic-release creates notes and Gitlab treats those the same as commits. Thus when requiring signed commits, they also require you to sign notes, which is not possible.
https://gitlab.com/gitlab-org/gitlab/-/issues/209901

**Comment by maaheeb** [comment]
Would like to see this as well!

**Comment by sharkymcdongles** [comment]
+1 for asking for a merge pretty please <3

@travi @gr2m maybe pretty please

**Comment by SchoolGuy** [comment]
@travi Is there anything one can do to help with the progress of this PR? I would like to enable signed commits and tags for my repository but cannot do that due to the fact that this feature is missing with this project. My GPG key for the project would be available as a secret in GitHub Actions and Git would obviously be configured to use the mentioned key.

**Comment by fastman** [comment]
Looking forward to having this merged.

File: lib/git.js [hunk 1]
```diff
@@ -218,12 +218,36 @@ export async function verifyAuth(repositoryUrl, branch, execaOptions) {
...
+ * @param {Object} tagOptions Tag config options.
  * @param {Object} [execaOpts] Options to pass to `execa`.
  *
  * @throws {Error} if the tag creation failed.
  */
-export async function tag(tagName, ref, execaOptions) {
-  await execa("git", ["tag", tagName, ref], execaOptions);
+async function tag(tagName, ref, {tagAnnotate, tagSign, tagMessage}, execaOptions) {
+  const args = ["tag", tagName, ref];
+  const gitTagSign = await shouldSignTag(execaOptions);
```
sheerlox, line 228: What is the reasoning behind accounting for & giving precedence to the git configuration?

I see a few problems with that:
- if I set `tagSign` to `false` in my config, it might still sign the tag based on my system configuration
- it makes the changes breaking, as it will start unsolicitedly signing tags

Because of that, I don't think the git configuration should be accounted for. If overriding the release configuration behavior is needed in a specific setup, overriding it with the CLI flag is enough.

sheerlox, line 228: What is the reasoning behind accounting for & giving precedence to the git configuration?

I see a few problems with that:
- if I set `tagSign` to `false` in my config, it might still sign the tag based on my system configuration
- it makes the changes breaking, as it will start unsolicitedly signing tags

Because of that, I don't think the git configuration should be accounted for. If overriding the release configuration behavior is needed in a specific setup, overriding it with the CLI flag is enough.

test/get-config.test.js
```diff
@@ -296,7 +320,9 @@ test.serial('Read configuration from file path in "extends"', async (t) => {
     branches: ["test_branch"],
     repositoryUrl: "https://host.null/owner/module.git",
     tagFormat: `v\${version}`,
-    plugins: ["plugin-1", ["plugin-2", { plugin2Opt: "value" }]],
```
sheerlox, line 299: This should probably not be removed.

docs/usage/configuration.md
```diff
@@ -104,6 +104,36 @@ The [Git tag](https://git-scm.com/book/en/v2/Git-Basics-Tagging) format used by
...
+
+**Note**: `tagSign` takes precedence over `tagAnnotate`. The git config `tag.gpgsign` takes precedence over both.
+
+### tagMessage
+
+Type: `String`<br>
+Default: `release ${nextRelease.version}`<br>
+CLI arguments: `-tm`, `--tag-message`
+
+The message for the tag. The tag message is generated with [Lodash template](https://lodash.com/docs#template) and will be compiled with the `nextRelease` variable.
```
sheerlox, line 133: `nextRelease` being an internal variable, there should be a link to [its documentation](https://semantic-release.gitbook.io/semantic-release/developer-guide/js-api#nextrelease).

cli.js
```diff
@@ -23,6 +23,14 @@ Usage:
     .option("b", { alias: "branches", describe: "Git branches to release from", ...stringList, group: "Options" })
     .option("r", { alias: "repository-url", describe: "Git repository URL", type: "string", group: "Options" })
     .option("t", { alias: "tag-format", describe: "Git tag format", type: "string", group: "Options" })
+    .option("ta", {
+      alias: "tag-annotate",
+      describe: "Whether to make an unsigned, annotated tag",
```
sheerlox, line 28: Streamline to "Toggle tag annotation".

cli.js
```diff
@@ -23,6 +23,14 @@ Usage:
...
+      type: "boolean",
+      group: "Options",
+    })
+    .option("ts", {alias: "tag-sign", describe: "Whether to make GPG-signed tag", type: "boolean", group: "Options"})
```
sheerlox, line 32: Streamline to "Toggle tag signing".

index.js
```diff
@@ -202,7 +202,16 @@ async function run(context, plugins) {
...
   } else {
     // Create the tag before calling the publish plugins as some require the tag to exists
-    await tag(nextRelease.gitTag, nextRelease.gitHead, { cwd, env });
+    await tag(
+      nextRelease.gitTag,
+      nextRelease.gitHead,
+      {
+        tagAnnotate: options.tagAnnotate,
+        tagSign: options.tagSign,
+        tagMessage: template(options.tagMessage)({ nextRelease }),
```
sheerlox, line 211: I would make the `tagMessage` optional, and default to the `nextRelease.gitTag`.

Example use case: I want my release tags to be signed, but I don't care about the annotation message.

sheerlox (commented):
I've added a few comments. The main doubt I have with this change is the git config precedence over the `semantic-release` one.

I also notice you haven't added test cases covering your changes.

lib/git.js
DiogoAbu: If I remember correctly...
I check for the git config first because signed tags **must** have a message, otherwise it fails. Currently if I set `tag.gpgsign` to `true` it will fail when creating a tag.
So if you set `tagSign` to `false` and `tag.gpgsign` is `true` it will still sign a tag to prevent the release from failing.
That said I don't think it is unsolicitedly if I previously set `tag.gpgsign`.
Does this make sense?

AliSajid:
Looking forward to this getting merged.

Korben11:
In the meantime, my workaround doesn't require a force push for using Gitlab API.

```sh
#!/usr/bin/env bash
# release.sh

# TODO: remove when merged, this is workaround for annotated tags
# https://github.com/semantic-release/semantic-release/pull/1871

if [ -z "$1" ]; then
	echo "Please provide a tag."
	echo "Usage: ./release.sh v[X.Y.Z]"
	exit
fi

GIT_TAG=$1

JSON_STRING=$( jq -n \
                  --arg tag_name "${GIT_TAG}" \
                  --arg ref "main" \
                  '{tag_name: $tag_name, ref: $ref}' )

if [ -n "$2" ]; then
  NOTES=$2
  JSON_STRING=$(echo $JSON_STRING | jq -r --arg message "${NOTES}" '.message = $message')
fi

echo "Recreating annotated tag ${GIT_TAG}..."

URL="https://gitlab.com/api/v4/projects/${CI_PROJECT_ID}/repository/tags"

curl -X DELETE "${URL}/${GIT_TAG}" -H "PRIVATE-TOKEN: ${GITLAB_TOKEN}"
curl -X POST "${URL}" -H "Content-Type: application/json" -H "PRIVATE-TOKEN: ${GITLAB_TOKEN}" -d "${JSON_STRING}"

```

```json
{
  ...
  "plugins": [
    "@semantic-release/release-notes-generator",
    "@semantic-release/changelog",
    "@semantic-release/git",
    [
      "@semantic-release/exec",
      {
        "publishCmd": "./release.sh ${nextRelease.gitTag} \"${nextRelease.notes}\""
      }
    ],
    "@semantic-release/gitlab"
  ],
  ...
}
```

babblebey (commented):
Hi @DiogoAbu, 

Thank you for sticking to this, I have seen the eagerness within the community to get this one merged 😃. I hope we will be able to make this happen in the coming days/weeks now.

I have these few observation I'd like us to address when you have the time 😉 

- I noticed that in the test cases, we only have coverage for cases where the config have the below particular setup testing the new properties; It'd be nice to have coverage for other cases...

    ```js
    {
      tagAnnotate: false,
      tagSign: false,
      tagMessage: `release \${nextRelease.version}`,
    }
    ```
- Let's resolve the current conflict and pull master into your branch
- Let's resolve some of @sheerlox review

eladchen:
How about trying out [atomic-release](https://github.com/abstracter-io/atomic-release)?

Its an SDK, which means you can create a release flow you fully control...

thasmo:
Would love to see this merged. ❤️

Xe:
If I donate to the project with the sole requirement that this PR get merged, how much would it cost?

tburny:
Please merge this!

PR #1874, opened 2021-04-07
Minor fixes to documentation

XC-:
As I was reading the documentation and starting my own plugin project, I noticed while trying to print the context object content in different situations, that the documentation has some errors:

- analyzeCommits missing from the lifecycle list
- "verify" lifecycle should be "verifyConditions"

These are addressed by this pull request.


gr2m (approved):
thanks a lot for taking the time 💐

PR #1892, opened 2021-04-16
docs(workflow-configurator): fix single typo

jpsullivan:
Just a small typo fix where incorrect usage of "there" was found.

Thanks for building and maintaining such a great tool!


gr2m (approved):
thanks!

Issue #1793, opened 2021-02-12
Improved handling of rebases on pre-release branches


dominykas:
## New feature motivation

When a pre-release branch is rebased, the pre-release tags may no longer exist on it, and so `semantic-release` may no longer be able to detect a correct version to release next. This is mostly a correct behavior and resolving it manually is explained in [troubleshooting docs](https://github.com/semantic-release/semantic-release/blob/master/docs/support/troubleshooting.md#release-not-found-release-branch-after-git-push---force).

That said, the manual steps are cumbersome and the UX is not that great.

The situation is especially painful if one tries to implement pre-releases for PRs (e.g. https://github.com/semantic-release/semantic-release/issues/1433), although I have run into this in the past with `beta` branch approaches as well.

## New feature description

`semantic-release` should check the list of existing tags _for the specific pre-release_ across _all_ commits (not just the ones in the current branch) to determine the next increment. Example:

- Next release, as it would get detected today: `v5.1.0-new-feature.1`
- There are dangling tags: `v5.1.0-new-feature.1`, `v5.1.0-new-feature.2`
- Desired next release: `v5.1.0-new-feature.3`

## New feature implementation

When calculating the `nextRelease` (continuing with the example above):

- perform a `git tag --list 'v5.1.0-new-feature.*'`
- `semver.rsort` the result
- `semver.parse` the first item, if available
    - It will include `{ prerelease: [ 'new-feature', 2 ] }`
- increment the count to `3`, instead of using `1`

I figure that folks can only get into this situation via rebasing or via tag deletion, and while I'm not sure if there's undesirable implications for this, I figure that getting a release at a new version is more desirable than the existing failure to do anything?

Happy to PR, if this is an acceptable approach.

bigbite95110:
I Nkow I Try every way Now we can connect for don't Have Figth  Wait .....
🤫

ในวันที่ ศ. 12 ก.พ. 2021 20:02 Dominykas Blyžė <notifications@github.com>
เขียนว่า:

gr2m:
I agree that the current process is fragile, I'd love to see it improved

dominykas:
OK so, I opened up a WIP PR: https://github.com/semantic-release/semantic-release/pull/1820

Is there a better way? Also curious if there's any objections to making `getNextVersion` becoming `async`.

[...]

PR #1894, opened 2021-04-16
feat: Add new option for appending a build suffix to prereleases

adambullmer:
Fixes: #1793


eLarocque:
We would love to see this approved, merged and released since we are also having issues with pre-releases in combination with squashes and rebases.

Thanks!

shih-js:
Hello, what is the status of this pull request?

I'm running into this exact issue /w rebasing pre-release branches.

shih-js (approved):

mamachanko:
The suggested approach appears to limit optional addition of build metadata to prereleases. However, I would find it helpful if they wouldn't be limited to pre-releases. After all, `1.2.3+stuff` is a valid and reasonable semver as per spec.

chrisEff:
I need exactly this feature. I'd gladly review it myself if I were familiar with the project's code/structure enough to do so, but unfortunately I'm not. So I would be VERY grateful if someone from the main contributors could take a look at this...

ThommyH:
we also need this, is there a plan to merge that in?

suhaspotluri (approved):

JTeeuwissen:
Bump, this would be very helpful.

PR #1901, opened 2021-04-20
Enable option to add custom git note

nozelrosario:
Currently the git push is rejected if there are server side git hooks to validate git notes. So need option to add custom git note.


nozelrosario:
need work

Issue #1168, opened 2019-04-30
Dry run still tries to connect to git


atdiff:
## Current behavior
When running `semantic-release --dry-run --ci false`, git errors occur.

## Expected behavior
On a dry-run, git shouldn't be needed, right? And particularly with ci turned off.

## Environment

- **semantic-release** version: 15.13.3
- CI environment: Jenkins
- Plugins used: @semantic-release/npm, semantic-release-verify-deps
- **semantic-release** configuration: 
```
{
  "name": "example",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "repository": {
    "type": "git",
    "url": "http://gitexample.com/gitexample/example.git"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "dependencies": {
    "@example/example": "0.0.1"
  },
  "publishConfig": {
    "registry": "http://nexusexample.com/repository/npm-internal/",
    "tag": "latest"
  },
  "release": {
    "branch": "origin/master",
    "plugins": [
      "@semantic-release/npm",
      "@semantic-release/git"
    ],
    "verifyConditions": [
      {
        "path": "semantic-release-verify-deps",
        "dependencies": true,
        "devDependencies": false,
        "regExps": [
          "\\D$"
        ]
      }
    ]
  },
  "devDependencies": {
    "@semantic-release/npm": "^5.1.4",
    "@semantic-release/git": "^7.0.8",
    "semantic-release-verify-deps": "^1.0.2"
  }
}

```
- CI logs:
```
[1:36:52 AM] [semantic-release] › ✔  Loaded plugin "verifyConditions" from "semantic-release-verify-deps"
2019-04-30T01:36:52.912Z semantic-release:plugins options for @semantic-release/commit-analyzer/analyzeCommits: {}
[1:36:52 AM] [semantic-release] › ✔  Loaded plugin "analyzeCommits" from "@semantic-release/commit-analyzer"
2019-04-30T01:36:52.913Z semantic-release:plugins options for @semantic-release/npm/prepare: {}
[1:36:52 AM] [semantic-release] › ✔  Loaded plugin "prepare" from "@semantic-release/npm"
2019-04-30T01:36:52.913Z semantic-release:plugins options for @semantic-release/git/prepare: {}
[1:36:52 AM] [semantic-release] › ✔  Loaded plugin "prepare" from "@semantic-release/git"
[...]

PR #1924, opened 2021-05-08
feat: added disable-push option to disable pushing to git

doteric:
I firstly thought about skipping the whole `verifyPushAuth`, but then just went with a bit of a more safe approach.

`isBranchUpToDate` is done pretty badly so it also returns false when local is up to date, but just there is some additional commit. Also the user might have or not permissions to read the repo therefore an error will throw if he can't read the repo for some reason. I didn't want to put an additional (third) try-catch, because I think it's a trivial case.
  
This will be pretty useful when using the JS API to grab the newRelease version and note.

### DIFF
--- a/docs/usage/configuration.md
+++ b/docs/usage/configuration.md
@@ -3,7 +3,7 @@
 **semantic-release** configuration consists of:
 - Git repository ([URL](#repositoryurl) and options [release branches](#branches) and [tag format](#tagformat))
 - Plugins [declaration](#plugins) and options
-- Run mode ([debug](#debug), [dry run](#dryrun) and [local (no CI)](#ci))
+- Run mode ([debug](#debug), [dry run](#dryrun), [local (no CI)](#ci) and [disable push](#disablepush))
 
 All of these options can be configured through config file, CLI arguments or by extending a [shareable configuration](shareable-configurations.md).
 
@@ -118,7 +118,15 @@ CLI arguments: `-d`, `--dry-run`
 
 The objective of the dry-run mode is to get a preview of the pending release. Dry-run mode skips the following steps: prepare, publish, success and fail. In addition to this it prints the next version and release notes to the console.
 
-**Note**: The Dry-run mode verifies the repository push permission, even though nothing will be pushed. The verification is done to help user to figure out potential configuration issues.
+**Note**: The Dry-run mode verifies the repository push permission, even though nothing will be pushed. The verification is done to help user to figure out potential configuration issues. If you do not want to verify push permissions nor push any changes use the [disable push](#disablepush) flag.
+
+### disablePush
+
+Type: `Boolean`<br>
+Default: `false`<br>
+CLI arguments: `--disable-push`
+
+Set to `true` to disable verifying push permissions and pushing anything. This can be useful when wanting to handle the release by yourself.
 
 ### ci
 
--- a/index.js
+++ b/index.js
@@ -17,7 +17,7 @@ const {extractErrors, makeTag} = require('./lib/utils');
 const getGitAuthUrl = require('./lib/get-git-auth-url');
 const getBranches = require('./lib/branches');
 const getLogger = require('./lib/get-logger');
-const {verifyAuth, isBranchUpToDate, getGitHead, tag, push, pushNotes, getTagHead, addNote} = require('./lib/git');
+const {verifyPushAuth, isBranchUpToDate, getGitHead, tag, push, pushNotes, getTagHead, addNote} = require('./lib/git');
 const getError = require('./lib/get-error');
 const {COMMIT_NAME, COMMIT_EMAIL} = require('./lib/definitions/constants');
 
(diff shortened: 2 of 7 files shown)


index.js
```diff
@@ -111,6 +114,8 @@ async function run(context, plugins) {
 
       if (options.dryRun) {
         logger.warn(`Skip ${nextRelease.gitTag} tag creation in dry-run mode`);
+      } else if (options.disablePush) {
+        logger.warn(`Skip ${nextRelease.gitTag} tag creation in disable-push mode`);
```
travi, line 118: the version tag is very core to the operation of semantic-release. it is _very_ unlikely for us to accept a change that would enable disabling this functionality

docs/usage/configuration.md
```diff
@@ -118,7 +118,15 @@ CLI arguments: `-d`, `--dry-run`
...
-**Note**: The Dry-run mode verifies the repository push permission, even though nothing will be pushed. The verification is done to help user to figure out potential configuration issues.
+**Note**: The Dry-run mode verifies the repository push permission, even though nothing will be pushed. The verification is done to help user to figure out potential configuration issues. If you do not want to verify push permissions nor push any changes use the [disable push](#disablepush) flag.
+
+### disablePush
+
+Type: `Boolean`<br>
+Default: `false`<br>
+CLI arguments: `--disable-push`
+
+Set to `true` to disable verifying push permissions and pushing anything. This can be useful when wanting to handle the release by yourself.
```
travi, line 129: > This can be useful when wanting to handle the release by yourself

what are you hoping to accomplish by "handling the release by yourself"? semantic-release is not a general purpose workflow tool. it is an opinionated release tool that is meant to manage the steps of releasing a version for you. if there is a desire to "handle the release yourself", semantic-release is likely not the appropriate tool for your goal

travi:
what is your goal with this change? you havent provided us context around what you are trying to accomplish or gaps that you see with the current implementation of semantic-release. is there a conversation somewhere that provides more context around your goals here? if not, starting with a discussion in the future before putting effort into a PR could save you some effort before gaining agreement on goals and direction from the project maintainers.

as general feedback, we are trying to limit the addition of new options without _very_ strong justification. they introduce additional codepaths that have the potential for significant maintenance effort over time as well as new situations that can result in user error or confusion and support issues for maintainers to spend time helping to resolve.

doteric:
Hey @travi ,
sorry for not explaining the reasoning in details. So let me give you some more details.
What I want to do basically is use the semantic-release JS API to get the new release version and notes without the requirement of having push permissions to the repository. Having those I can create a GitLab release with a CI_JOB_TOKEN and no push permissions will be required. GitLab token scope is pretty limited so using a project access token does not suit us, because later that token does not have access to other repositories.
I was searching for some projects that would get the new version tag and notes, but I couldn't find anything that would suit my needs.

Answering your question about additional paths - yes it maybe does create additional paths, but I tried to do it the most safest way there could be I think. This means that if you are not running the --disable-push mode then the code looks exactly the same like previously. It's actually very similar to --dry-run, but without verifying for push permissions and throwing an error if they are not present.

docs/usage/configuration.md
doteric: Answered in: https://github.com/semantic-release/semantic-release/pull/1924#issuecomment-835769990

index.js
doteric: Answered in: https://github.com/semantic-release/semantic-release/pull/1924#issuecomment-835769990

paulosoaresvieira:
I'd like to share that this feature will solve my problem... 
Currently, I have a CI pipeline that has multiple slaves in parallel generating the build on different platforms mac/linux/windows...
So using the current implementation when i call `npx semantic-release` it'll be called on all slaves so it try to create a tag on each node... so the first finishing will create the tag but the others will fail... Then my idea is to not create the tag on the slaves and run the `npx semantic-release` only to genereate the version and after all slaves finished there create the version tag...
Is it possible to go on with this PR?

doteric:
> I'd like to share that this feature will solve my problem...
> Currently, I have a CI pipeline that has multiple slaves in parallel generating the build on different platforms mac/linux/windows...
> So using the current implementation when i call `npx semantic-release` it'll be called on all slaves so it try to create a tag on each node... so the first finishing will create the tag but the others will fail... Then my idea is to not create the tag on the slaves and run the `npx semantic-release` only to genereate the version and after all slaves finished there create the version tag...
> Is it possible to go on with this PR?

Hey, if the Maintainers are willing to accept such features then I can for sure polish my PR wherever needed 😃   
It's a pretty basic draft, probably if we get the appropriate support for this and more scenarios this is useful then we might be able to proceed.  
The PR itself has no breaking nor potentially misleading or adding additional complexity changes imo.

paulosoaresvieira:
> > I'd like to share that this feature will solve my problem...
> > Currently, I have a CI pipeline that has multiple slaves in parallel generating the build on different platforms mac/linux/windows...
> > So using the current implementation when i call `npx semantic-release` it'll be called on all slaves so it try to create a tag on each node... so the first finishing will create the tag but the others will fail... Then my idea is to not create the tag on the slaves and run the `npx semantic-release` only to genereate the version and after all slaves finished there create the version tag...
> > Is it possible to go on with this PR?
> 
> Hey, if the Maintainers are willing to accept such features then I can for sure polish my PR wherever needed 😃
> It's a pretty basic draft, probably if we get the appropriate support for this and more scenarios this is useful then we might be able to proceed.
> The PR itself has no breaking nor potentially misleading or adding additional complexity changes imo.

Thanks! Currently I'm using it and it's working fine... Let's wait for it...

giladgd:
Is there any update on this?
I would greatly appreciate it if this PR could be merged.

@travi You can see the number of upvotes on this comment to get an idea of the demand for the ability to skip the git push check, which this PR covers: https://github.com/semantic-release/semantic-release/issues/1168#issuecomment-552883808

PR #1925, opened 2021-05-10
docs(distribution-channels): fix branch name

mhagnumdw:
If I understand correctly, I believe that this correction is valid, otherwise please disregard this PR.


travi (approved):
thanks!

PR #1961, opened 2021-05-31
docs(plugin developer guide): correction, additional details / advanced

XC-:
A correction to the earlier documentation: the "nextRelease" object is NOT similar to the
lastRelease and there are some differences. Because of this, the object is now documented as a
separate entry. Added also information on how the error handling should be done so that
semantic-release would be able to handle errors from plugins properly and call "fail" lifecycle
plugins. Included some more advanced information on how the analyzeCommits-lifecycle plugins work
and how the result of the initial analyzer can be overridden. This is an edge case, but could happen
and is good to understand.


gr2m (approved):
This is great, thank you!

PR #1962, opened 2021-06-01
docs(README): improve sentence

### LINKED ISSUE #1960: The phrase "an example of the release type" in README is misleading
## Current behavior
In [Commit Message Format](https://github.com/semantic-release/semantic-release#commit-message-format) section of the README, it states `Here is an example of the release type that will be done based on a commit messages:` right before the table and this is misleading since these three patterns listed in the table are the only ones that trigger the release and nothing else. I had some discussion whether if chore or perf would trigger the release by default.

This is another issue stating something similar
https://github.com/semantic-release/semantic-release/issues/1949

## Expected behavior
I understand that the types that trigger release can be configured by config/preset settings, but having a reference to which types of commit actually triggers patch/minor/breaking change release by default (which I am assuming this https://github.com/semantic-release/semantic-release/issues/1720) would resolve this confusion. 

If this is correct, we can replace "the example of the release type" with "the list of the release type" along with a sentence that states this is set by default.

## Environment

- **semantic-release** version: 
- CI environment: 
- Plugins used: 
- **semantic-release** configuration: 
- CI logs:

HonkingGoose:
fixes #1960

## Changes:

- Improve sentence

## Context:

The previous sentence was OK, but I think it can be improved a bit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ By default **semantic-release** uses [Angular Commit Message Conventions](https:
 
 Tools such as [commitizen](https://github.com/commitizen/cz-cli) or [commitlint](https://github.com/conventional-changelog/commitlint) can be used to help contributors and enforce valid commit messages.
 
-The table below shows which commit message gets you which release type when `semantic-release` runs:
+The table below shows which commit message gets you which release type when `semantic-release` runs (using the default configuration):
 
 | Commit message                                                                                                                                                                                   | Release type               |
 | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | -------------------------- |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -52,7 +52,7 @@ By default **semantic-release** uses [Angular Commit Message Conventions](https:
 
 Tools such as [commitizen](https://github.com/commitizen/cz-cli) or [commitlint](https://github.com/conventional-changelog/commitlint) can be used to help contributors and enforce valid commit messages.
 
-Here is an example of the release type that will be done based on a commit messages:
+The table below shows which commit message gets you which release type when `semantic-release` runs:
```
gr2m, line 55: maybe we should add "using the default configuration", as commits can be configured using [`@semantic-release/commit-analyzer`](https://github.com/semantic-release/commit-analyzer) options?

HonkingGoose:
```suggestion
The table below shows which commit message gets you which release type when `semantic-release` runs (using the default configuration):
```

gr2m (approved):
Thanks a lot!

HonkingGoose:
@gr2m Does the new text help with issue #1960, or maybe even resolve that issue?

gr2m:
> @gr2m Does the new text help with issue #1960, or maybe even resolve that issue?

good point, I think the issue is resolved thanks to your change 👍🏼

PR #1989, opened 2021-06-25
docs(plugins): clarify order of execution

nandorojo:
See this discussion: https://github.com/semantic-release/semantic-release/discussions/1987#discussioncomment-917795

I'm new to semantic-release and was originally confused by the order, so hopefully this clears it up a little for newcomers.

cc @travi

### DIFF
--- a/docs/usage/plugins.md
+++ b/docs/usage/plugins.md
@@ -16,6 +16,8 @@ A plugin is a npm module that can implement one or more of the following steps:
 | `success`          | No       | Responsible for notifying of a new release.                                                                                                                                                                          |
 | `fail`             | No       | Responsible for notifying of a failed release.                                                                                                                                                                       |
 
+Release steps will run in that order. At each step, **semantic-release** will run every plugin in the [`plugins` array](#plugins-declaration-and-execution-order), as long as the plugin implements the step.
+
 **Note:** If no plugin with a `analyzeCommits` step is defined `@semantic-release/commit-analyzer` will be used.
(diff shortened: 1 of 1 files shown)


gr2m (approved):
I like it 👍🏼

travi (approved):

travi:
thanks @nandorojo. this is a great clarification and we really appreciate the contribution!

nandorojo:
Happy to help. I'm blown away by the precise thought and clear mental model put into semantic-release (and its docs).

Once I figure out how to use it in a monorepo, and how to do multiple major version bumps in a prerelease branch, I'll be using it forever.

travi:
> how to do multiple major version bumps in a prerelease branch

please see my response in https://github.com/semantic-release/semantic-release/discussions/1988#discussioncomment-935933

> Once I figure out how to use it in a monorepo

semantic-release does not officially support monorepos. there are some thirdparty tools that add compatibility on top of semantic-release, but only support limited subsets of the various approaches to managing monorepos. i have personally used [semantic-release-monorepo](https://www.npmjs.com/package/semantic-release-monorepo) in situations where i've needed to work in a monorepo situation. in order to support officially, we would want to close far more of the gaps left by those tools, which would add significant complexity to what we already manage. 

I understand that both of these answers are probably disappointing, but please understand that semantic-release is an opinionated tool with specific targeted use cases. we have to draw the line somewhere with the flexibility of use cases that we support.

PR #2018, opened 2021-07-07
docs(circleci): Update CircleCI recipes config

KyleTryon:
Updates the included CircleCI config file example.

* Now uses 2.1 style config
* Uses Node Orb
* Use pre-defined node/test job
* Uses Matrix jobs to test multiple node versions.
* Updated Node versions
* Using new CircleCI docker images (from node executor)
* Updated language/terminology in description

### DIFF
--- a/docs/recipes/circleci-workflows.md
+++ b/docs/recipes/circleci-workflows.md
@@ -10,46 +10,39 @@ Alternatively, the default `NPM_TOKEN` and `GH_TOKEN` can be easily [setup with 
 
 ### `.circleci/config.yml` configuration for multiple Node jobs
 
-This example is a minimal configuration for **semantic-release** with a build running Node 6 and 8. See [CircleCI documentation](https://circleci.com/docs/2.0) for additional configuration options.
+This example is a minimal configuration for **semantic-release** with tests running against Node 16 and 14. See [CircleCI documentation](https://circleci.com/docs/2.0) for additional configuration options.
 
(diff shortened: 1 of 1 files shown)


KyleTryon:
If desired I could instead make a PR that does not use Orbs. The outcome would look much more similar to the existing config but only with updated images and such. The suggestion in this PR is more optimized though perhaps maybe a bit opaque for new users.

gr2m (approved):
I'd say you are the expert and know what's best 😄 thanks for the updates!

docs/recipes/circleci-workflows.md
```diff
@@ -10,46 +10,39 @@ Alternatively, the default `NPM_TOKEN` and `GH_TOKEN` can be easily [setup with
...
-    docker:
-      - image: circleci/node:8
+    executor: node/default
     steps:
       - checkout
-      - run: npm install
+      - node/install-packages # Install and automatically cache packages
       # Run optional required steps before releasing
       # - run: npm run build-script
       - run: npx semantic-release
```
travi, line 31: sorry that i missed asking this before the merge, but does this step still belong here after defining the separate release job? it seems like this would run unconditionally w/o considering the required test job, but i could be misunderstanding something

KyleTryon: Hey @travi,

This is assuming the user needs to run `npm run build` or something similar prior to running the semantic-release CLI. In that case, this is a more efficient way of installing packages.

Further down in the workflow you'll see this job requires the `node/test` job(s) to have completed first.

```
- release:
          requires:
            - node/test
```

Let me know if that clears it up, I could be misunderstanding something too

travi: ok, i think the detail that i was overlooking was the `release` name on line 24. without noticing that, i wasnt sure how the requirement was tied back to those steps. this is just my lack of familiarity with the CircleCI specifics, but it makes sense now. thanks for clarifying!

PR #2064, opened 2021-07-30
🚧 ESM

gr2m:
I'm using this pull request to asses if moving to ESM is doable as part of v18. Still working through it


gr2m:
Actually I won't be able to work on this before Friday, August 6th. If anyone would like to work through the errors when running `npm test`, please go ahead. Any help is much appreciated!

travi:
for the `proxyquire` errors, we may need to swap for [`testdouble`](https://github.com/testdouble/testdouble.js/#api). it is the only one that i'm aware of that is ready to be used as a module loader to actually replace native es modules.

travi:
closing in favor of https://github.com/semantic-release/semantic-release/pull/2607

PR #2077, opened 2021-08-13
fix: prevent concurrency issue whene multiple channel are released

KillianHmyd:
This would reduce the risk of conflicts error when two different channels are released at the same time.

Because at each release for every channel a new note is pushed on `refs/notes/semantic-release`, sometimes we have the following error : 
```
To https://github.com/PayFit/subscriptions.git
 ! [rejected]        refs/notes/semantic-release -> refs/notes/semantic-release (fetch first)
error: failed to push some refs to 'https://github.com/PayFit/subscriptions.git'
hint: Updates were rejected because the remote contains work that you do
hint: not have locally. This is usually caused by another repository pushing
hint: to the same ref. You may want to first integrate the remote changes
hint: (e.g., 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
```

A fetch before adding a new note would reduce this risk.

### DIFF
--- a/index.js
+++ b/index.js
@@ -17,7 +17,17 @@ const {extractErrors, makeTag} = require('./lib/utils');
 const getGitAuthUrl = require('./lib/get-git-auth-url');
 const getBranches = require('./lib/branches');
 const getLogger = require('./lib/get-logger');
-const {verifyAuth, isBranchUpToDate, getGitHead, tag, push, pushNotes, getTagHead, addNote} = require('./lib/git');
+const {
+  verifyAuth,
+  isBranchUpToDate,
+  getGitHead,
+  tag,
+  push,
+  pushNotes,
+  getTagHead,
+  addNote,
+  fetchNotes,
+} = require('./lib/git');
 const getError = require('./lib/get-error');
 const {COMMIT_NAME, COMMIT_EMAIL} = require('./lib/definitions/constants');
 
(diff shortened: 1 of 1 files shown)


gr2m:
Could you add a test please?

KillianHmyd:
@gr2m I fixed the code, and so the tests, but not sure which tests should I add. As I only do a fetch note before pushing them, the only solution to test this concurrent issue would be to add mocks over the `git` helpers which doesn't seems to be a thing in this project 🤔

index.js
```diff
@@ -112,6 +122,7 @@ async function run(context, plugins) {
       if (options.dryRun) {
         logger.warn(`Skip ${nextRelease.gitTag} tag creation in dry-run mode`);
       } else {
+        await fetchNotes(options.repositoryUrl, {cwd, env});
```
gr2m, line 125: Could you add a comment on why we fetch the notes here, with a reference to the pull request?

gr2m (commented):
> A fetch before adding a new note would reduce this risk.

could you verify that the patch is resolving the conflicts in your environment?

KillianHmyd:
@gr2m Sorry for the latency on this PR.
So we tested it from our side and it resolves only partially our conflicts. In our case we release using semantic release on each branches, so even if it lower the risk of conflicts we still have some.

I thought about another solution but which would introduce a breaking change. The solution would be to use a dynamic note reference, so instead of pushing them on `semantic-release` ref, do it on `semantic-release-${TAG_REFERENCE}`.
This way each tag would have their own note references and it would prevents conflicts between two unrelated releases.

It could be possible to keep retro-compatibility by fetching notes from the new dynamic reference and the current static one.

I can open a ticket or a WIP PR to discuss about it if you think it is relevant 😃

KillianHmyd:
Closed for this one : https://github.com/semantic-release/semantic-release/pull/2085

PR #2087, opened 2021-08-23
fix(marked): reverted upgrade to v3

travi:
since marked v3 requires node v12

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,7 +33,7 @@ jobs:
           node-version: ${{ matrix.node-version }}
           cache: npm
       - name: Ensure dependencies are compatible with the version of node
-        run: echo 'engine-strict=true' >> .npmrc
+        run: npx ls-engines
       - run: npm ci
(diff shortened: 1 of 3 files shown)


gr2m (commented):
should we add `marked` to be ignored by renovate? We'd need to configure it in `package.json`'s `"renovate"` object

travi:
> should we add `marked` to be ignored by renovate?

i'd rather renovate keep us aware that there is an upgrade, but fail the build if it would be incompatible. it looks like the build is failing properly now, but i missed something else that isnt compatible with node v10. tracking that piece down now

travi:
i'm reverting the switch to use `ls-engines` because it revealed that we have a few other engines compatibility issues to resolve before we can make that change. i will follow up with those adjustments in a separate PR

gr2m (approved):

ThibaultVlacich:
Wouldn't it be better to allow both versions? `^2.0.0 || ^3.0.0`

PR #2088, opened 2021-08-23
fix(engines): fixed defined node version to account for the higher requirement from the npm plugin

travi:
* updates `engines` definition to account for the fact that `@semantic-release/npm` already required `>=10.19`, making the definition in this package effectively incorrect
* updates the docs to refer to node version accounting for this adjustment
* enforces that all installed packages are within the range defined by this package's `engines` using `ls-engines`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,8 +35,11 @@ jobs:
       - run: npm ci
       - run: npm run test:ci
 
-  engines:
-    runs-on: macos-latest
+  # separate job to set as required in branch protection,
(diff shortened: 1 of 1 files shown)


travi:
well, this is frustrating. this is passing locally. i wonder if there is some sort of platform-dependent optional dependency that requires node v12 and doesn't get installed on macOS

travi:
the problem seems unlikely to be OS based. the `npx ls-engines` command passes successfully when run in a docker container using the `node:14` base image, which is Debian based.

PR #2102, opened 2021-09-03
docs: fix broken link on github actions recipe

roziscoding:
The link to the secrets and environment variables was broken since the documentation changed.
I pointed the link to the URL that made most sense, since users can choose to create the secret wherever they want.


gr2m (approved):
Thanks a lot!

PR #2103, opened 2021-09-03
fix(npm): upgraded to the beta version of the npm plugin

travi:


gr2m (approved):
I'm such a n00b when it comes to Docker, I'm glad you figured it out 👏🏼

PR #2132, opened 2021-09-17
feat(node-version): raised the minimum required version to v14.17 (#2063)

travi:
BREAKING CHANGE: the minimum required version of node is now v14.17


gr2m (approved):

masone:
The minimum required version seems to be `node@16` due to https://github.com/semantic-release/npm/pull/304 forcing `npm@7`, but `node@14` still shipping with `npm@6`. We're getting some odd dependency conflicts when installing with `npm@6`:

```
...
│ ├─┬ @semantic-release/npm@7.1.3
│ │ ├── @semantic-release/error@2.2.0 deduped
...
│ │ ├─┬ npm@7.24.1
│ │ │ ├─┬ UNMET DEPENDENCY @npmcli/arborist@2.8.3
│ │ │ │ ├─┬ UNMET DEPENDENCY @npmcli/installed-package-contents@1.0.7
│ │ │ │ │ ├─┬ UNMET DEPENDENCY npm-bundled@1.1.2
│ │ │ │ │ │ └── UNMET DEPENDENCY npm-normalize-package-bin@1.0.1
...
```

This applies to `semantic-release@17` as well, which ships with `semantic-release/npm@7.1.0` too because of the caret range https://github.com/semantic-release/semantic-release/blob/v17.4.7/package.json#L26

travi:
> The minimum required version seems to be `node@16` due to [semantic-release/npm#304](https://github.com/semantic-release/npm/pull/304) (released in semantic-release/npm@7.1.0) forcing `npm@7`, but `node@14` still shipping with `npm@6`.

`node@14` still ships with `npm@6` because of the breaking changes in `npm@7`. that is the only reason that it doesn't ship with node by default until v16. `npm@7` is [compatible with node versions all the way down to `v10`](https://github.com/npm/cli/blob/6ae8cbe7ed70c17b121bd9ea42f8d8af11823c18/package.json#L238). 

if you are still using `npm@6` in your project and having `npm@7` in your dependency tree because of `semantic-release` is causing trouble, we recommend not installing `semantic-release` as a local dependency of your project. since `semantic-release` should not be needed by your project until the release step of your CI process, you can either install it just before the release happens or use `npx` to install it just-in-time.

PR #2139, opened 2021-09-20
docs: add section on logger usage for plugins

jpoehnelt:
Based upon my research in https://github.com/semantic-release/semantic-release/discussions/2137.


gr2m (approved):
Thanks a lot!

PR #2141, opened 2021-09-20
docs: add new community plugin

jpoehnelt:
The [semantic-release-interval](https://www.npmjs.com/package/semantic-release-interval) plugin provides functionality to trigger a new release if the oldest commit since the last release is over a certain age.

This solves a particular use case where no fix/feat commits were made but many changes have built up in other areas (ci, build, chore) that may have some impact on the generated artifacts.

Example debug output:

```
[7:32:30 PM] [semantic-release] [semantic-release-interval] › ℹ  committerDate: 2021-09-20T17:58:24+00:00
[7:32:30 PM] [semantic-release] [semantic-release-interval] › ℹ  committerDate: 2021-09-20T19:31:05+00:00
[7:32:30 PM] [semantic-release] [semantic-release-interval] › ℹ  Oldest `committerDate` before threshold, 1 week(s).
  oldest commit: 2021-06-21T19:09:07+00:00
  threshold:     2021-09-13T19:32:30+00:00
[7:32:30 PM] [semantic-release] [semantic-release-interval] › ℹ  Setting release to 'patch' based upon interval.
```

### DIFF
--- a/docs/extending/plugins-list.md
+++ b/docs/extending/plugins-list.md
@@ -140,3 +140,5 @@
   - `verifyConditions`: Validate configuration and verify ```HEROKU_API_KEY```
   - `prepare`: Update the package.json version and create release tarball
   - `publish`: Publish version to heroku
+- [semantic-release-interval](https://www.npmjs.com/package/semantic-release-interval)
+  - `analyzeCommits`: Triggers a new release if the oldest commit since the last release is over a certain age.


jpoehnelt:
It minimizes drift between currently published artifacts from the prior release and the generated artifacts in the current repository that may have changed due to changes in dependencies. For example, default browserlists/Babel presets change over time and dependabot regularly updates related dependencies. In this scenario, no fix/bug/etc will trigger a release, but there will be drift from published artifacts. Thus the options are to patch on every chore/build/etc which adds significant noise, or use something like this.

gr2m:
> default browserlists/Babel presets change over time and dependabot regularly updates related dependencies

I think I would create a release when the output changes due to browserlist/Babel changes. If something breaks that you can at least pin it down to what cause it. If you do a scheduled release that might contain many of the changes it will be harder to dissect?

jpoehnelt:
> I think I would create a release when the output changes due to browserlist/Babel changes.

Do you diff artifacts then? It would be nice if these things were deterministic, but that's just not reality for most projects. I don't want to use a tool like Bazel for everything.

> If something breaks that you can at least pin it down to what cause it.

More frequent releases allow this. I have libraries where there is no functional change in code, but things break when I create a release months later because of some dependency. In the meantime, there could be hundreds of dependabot pull requests to sift through to figure out which one it was.

gr2m:
> Do you diff artifacts then

I do not. But I'd publish compiled code that might change based on my build tools, and it would be how users consume my library, then I would release new versions if the build tools updated, just in case.

---

To be honest I'm not convinced that this is a good plugin to list in the community section, it feels like it goes what we are trying to build with semantic-release. But when in doubt, I'd rather list it when not. Let me think about it a bit more and also chat with @travi

jpoehnelt:
> might change based on my build tools

But how do you ever know? 

image

Here are some updates to devDependencies. Some are innocuous and related to lint. Others such as `core-js` could have changes to compiled code.

travi:
> But how do you ever know?

i think this is the part that i struggle with the most in this case. if you don't know, i don't understand the value of releasing just because of the _possibility_ that _something might have_ changed. 

one of the most core concepts of `semantic-release` is to encourage communicating the impact of changes to the package consumers. not only is it unknown in this case whether there _are_ changes as a result of build tooling updates since the latest release, but it is also unknown what the _impact_ of those changes are to the consumers of the package. was a bug fixed? was a bug introduced? did the behavior stay the same from the perspective of the consumers, even though there were minor differences in the output? for example, i find it useful to run integration tests against the public interface of many of my packages _after_ building/bundling using [package-preview](https://npm.im/package-preview) to understand when the publicly exposed behaviors change. if those don't change, i personally avoid releasing.

while the implementation of `semantic-release` does provide significant flexibility in the workflows it can enable, there are some that can be enabled that are outside the list that we would recommend using `semantic-release` for. the goal of that flexibility is to enable extension points for release steps that are in line with the goals of releasing with `semantic-release`. `semantic-release` is not intended to be a general purpose workflow tool and this at least flirts with the grey area of being outside the intention of `semantic-release`. while this may enable a workflow that is useful for your team, i share @gr2m's concern about including it in the community list. inclusion does not mean official endorsement, so i'm not entirely against including it, but it feels like we would need to include some sort of note clarifying that it enables an unsupported workflow.

PR #2145, opened 2021-09-21
docs: added complex pre-release examples

nitzano:
This pull request is a follow up of discussion [#2142](https://github.com/semantic-release/semantic-release/discussions/2142), adding the first two cases from it a new document which shows complex pre-release examples. 

Reviews and suggestions are more than welcome

I can add the remaining ones (Recipes 3,4) in another pull request following this one

### DIFF
--- a/docs/recipes/pre-releases-scenarios.md
+++ b/docs/recipes/pre-releases-scenarios.md
@@ -1,0 +1,130 @@
+# Pre-releases scenarios
+
+This recipe will walk you through some real life scenarios of publishing pre-release versions alongside releasing on the default distribution version simultaneously.
+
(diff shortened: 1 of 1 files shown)


travi:
sorry for the delay getting back to you about this. these scenarios look great. the one question i have is what you think about navigation to this new page. do you have specific ideas of where we should link to this page that would have helped you find these details when you were looking for them?

nitzano:
@travi good question,I think a good place for it will be to maybe re-arrange this section:

https://semantic-release.gitbook.io/semantic-release/recipes/recipes

Something as:

* Release workflow
  *  Publishing on distribution channels
  * Publishing maintenance releases
  * Publishing pre-releases
    *  [Pre-releasing a breaking change](https://github.com/semantic-release/semantic-release/blob/1405b94296059c0c6878fb8b626e2c5da9317632/docs/recipes/pre-releases.md)
    *  Pre-releasing features and fixes (this article)

PR #2154, opened 2021-09-27
feat(get-config): add support for defining plugins as an object

XC-:
This implementation allows the plugins to be defined as an object instead of an array. The benefit
is readability and usability, especially when using yaml format, where empty objects can be left
out. In Javascript, the benefit is more limited since it is necessary to put in the empty objects,
but even then the configuration should be more intuitive. This is implemented in backwards
compatible way, where the original format is still used internally, so nothing should break.

### DIFF
--- a/package.json
+++ b/package.json
@@ -122,8 +122,8 @@
     "lint": "xo",
     "pretest": "npm run lint",
     "semantic-release": "./bin/semantic-release.js",
-    "test": "nyc ava -v",
-    "test:ci": "nyc ava -v"
+    "test": "LANG='en_US.UTF-8' nyc ava -v",
+    "test:ci": "LANG='en_US.UTF-8' nyc ava -v"
   },
   "xo": {
     "prettier": true,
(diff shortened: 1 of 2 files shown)


XC-:
Even though this is still a work in progress, or rather, especially because this is wip, I would appreciate thoughts from @travi , @gr2m and/or @pvdlg . What I do not understand is why so many tests are suddenly broken :thinking:

gr2m:
I'd usually start by creating an issue to have a discussion before sending a PR :)

Can you update the docs to account for your change? That's the best starting point to have a discussion

XC-:
> I'd usually start by creating an issue to have a discussion before sending a PR :)

Yea, I'm way too used to thinking through PoCs so, this kinda just happened ^^;

> Can you update the docs to account for your change? That's the best starting point to have a discussion

Will do. I'll update them later today after work.

XC-:
@gr2m I added some tests and documentation + examples for the new feature. There was an problem with the tests, where some of the `yargs` output was in Finnish because of my `LANG` variable, so I added `LANG=en_US.UTF-8` to fix it. I also encountered one possibly flaky test that can occasionally fail (though not often), I'll open a issue for that in case there isn't one already.

BinToss:
Is the plugin load/run order preserved? If it isn't, then something like `plugins: { '$pluginId': string, ...PluginSpecificOptions }[]` may be preferable if JSON and YML syntax allows it.

Issue #1696, opened 2020-11-18
Plugin options conflict with CLI options


gempain:
Hi there - I'll skip how amazing you lib is, by now you should have realized that it should be in every dev's toolbox,

When writing plugins, I realized that the plugin options override CLI options. Take this plugin, say `test.js`:

```js
module.exports = {
  verifyConditions: (options) => {
    console.log(options);
  }
}
```

and this config `.eslintrc.js`:

```js
module.exports = {
  branches: 'latest',
  plugins: [
    ['./test', {}],
  ],
};
```

and run

```
$ semantic-release -d --ci false
...
{
  ...
  ci: false,
  noCi: true
}
```

Now update your config to:

```js
module.exports = {
  branches: 'latest',
  plugins: [
    ['./test', {ci: 'tada !'}],
  ],
};
```

And you get:

```
$ semantic-release -d --ci false
...
{
  ...
  ci: 'tada !',
  noCi: true
}
```

This could lead to issues. I think plugin options should be passed in a different option.

What do you think ?

gr2m:
Thanks for the kind words :) 

hmm I agree, it feels wrong that CLI options get overwritten by plugin configuration ... there might be a good reason but I don't know what it could be. Could you investigate? Feel free to start a PR and we continue the discussion there?

EvanCarroll:
I investigated a different issue @gr2m. In my case, I was wondering why `noCi` was getting set when environment `CI=1` was set. For me `get-config.js` has

```js
if (options.ci === false) {
  options.noCi = true;
}
```

This sets the `noCi` option if `ci` isn't set and it ignores `env` entirely. 

All of this logic is overly complex. Why are there two options `ci` and `noCi`? Shouldn't this simply be 

* command-line options set CI options if set
* If not set, see if environmental variable CI is set
* If not set, default CI to off in options.

gr2m:
I agree with the sentiment. I haven't written the code, so I can't say why it was written the way. 

That being said, reducing complexity and maintenance overhead is @travi's and my top priority right now, so we'd much appreciate if anyone would like to look into it and find out if we could simplify the options without breaking common use cases

EvanCarroll:
I simplified the options in my own branch with this,

```js
if (options.ci === true || options.ci === false) {
  options.noCi = !options.ci;
}
else if (options.noCi === true || options.noCi === false) {
  options.ci = !options.noCi;
}
else if (env["CI"] !== undefined) {
[...]

PR #2169, opened 2021-10-01
fix: Resolve problems with setting options.ci and .noCi

EvanCarroll:
This was covered in #1696. This is an attempt to resolve all outstanding
known issues between ci and noCi from command line and env.

### DIFF
--- a/lib/get-config.js
+++ b/lib/get-config.js
@@ -74,9 +74,18 @@ module.exports = async (context, cliOptions) => {
     ...pickBy(options, (option) => !isNil(option)),
     ...(options.branches ? {branches: castArray(options.branches)} : {}),
   };
-
-  if (options.ci === false) {
-    options.noCi = true;
+  
+  // Not sure why we have noCi, for more information see GH #1696
+  if (options.ci === true || options.ci === false) {
+    options.noCi = !options.ci;
+  }
+  else if (options.noCi === true || options.noCi === false) {
+    options.ci = !options.noCi;
+  }
+  else if (env["CI"] !== undefined) {
+    let ci = env["CI"] ? true : false;
+    options.ci = ci;
+    options.noCi = !ci;
   }
 
   debug('options values: %O', options);


travi:
the current documentation for the `ci` and `no-ci` flags can be found at https://semantic-release.gitbook.io/semantic-release/usage/configuration#ci.

while it is true that `--no-ci` is equivalent `--ci false`, it is not intended to force the ci detection to be `true`. from those docs, it is mentioned that this is intended for to
> skip Continuous Integration environment verifications. This allows for making releases from a local machine.
(which is possible, but not recommended for normal use)

in short, it is used to _override_ the automatic CI service detection in areas like https://github.com/semantic-release/semantic-release/blob/0f0c650b41764d1a3deb33631147c7ca0e39fe59/index.js#L32-L34

in addition, the `CI` environment variable is not significant to `semantic-release` for similar reasons. 

instead, the intention is for [env-ci](https://github.com/pvdlg/env-ci) to [detect the CI environment](https://github.com/semantic-release/semantic-release/blob/0f0c650b41764d1a3deb33631147c7ca0e39fe59/index.js#L29). as described in https://semantic-release.gitbook.io/semantic-release/developer-guide/plugin#context, `env-ci` normalizes some of the CI provided information for consumption by semantic-release. if you are running in a CI environment that should be detected but is not, you could consult the env-ci project to better understand why your environment is not being detected properly.

EvanCarroll:
That's not entirely clear me,

> while it is true that --no-ci is equivalent --ci false, it is not intended to force the ci detection to be true

The behavior of `--no-ci false` is not documented anywhere on that page are we inferring desired behavior from that code block, or should we documented this somewhere? It's also not clear to me why without specifying `--ci` or `--no-ci` I would want to run in dry-run mode?

The part about not reading `env["ci"]` directly makes sense in light of that abstraction library. But it doesn't make sense to ignore that environmental information when you trying to synchronize a global app state into options. At some point your users are just going to be concerned with whether or not they're running in ci mode, and what should they check: (A) just `options.ci`, (B) both `options.ci` and `options.notCi`, (C) `options.ci` and `options.notCi` and `env-ci`? Where does this get synthesized into a global state? If people are supposed to write code like `(!isCi ... && !options.noCi)` then I think we should document this.

travi:
we have plenty of gaps in our documentation, so there is always room for improvement. if you would like to help us improve the documentation around this behavior, we would welcome a PR for review.

EvanCarroll:
I think I have to maintain my own fork at this point. Feel free to close out my other patches.

Issue #1825, opened 2021-02-28
Suggestion: New badge


chalkygames123:
## New feature motivation

Thanks to #1715, the semantic-release's logo was added in [simple-icon v4.12.0](https://github.com/simple-icons/simple-icons/releases/tag/4.12.0) (https://github.com/simple-icons/simple-icons/pull/4995) and reflected in [shields.io](https://shields.io/), so we can now use a new badge.

## New feature description

_Not a feature_

## New feature implementation

Current:

[![semantic-release](https://img.shields.io/badge/%20%20%F0%9F%93%A6%F0%9F%9A%80-semantic--release-e10079.svg)](https://github.com/semantic-release/semantic-release)

```markdown
[![semantic-release](https://img.shields.io/badge/%20%20%F0%9F%93%A6%F0%9F%9A%80-semantic--release-e10079.svg)](https://github.com/semantic-release/semantic-release)
```

Suggestion:

[![semantic-release](https://img.shields.io/badge/semantic--release-enabled-brightgreen?logo=semantic-release)](https://github.com/semantic-release/semantic-release)

```markdown
[![semantic-release](https://img.shields.io/badge/semantic--release-enabled-brightgreen?logo=semantic-release)](https://github.com/semantic-release/semantic-release)
```

BTW, does anyone know where the color `#e10079` used in the current one comes from?

chalkygames123:
Sorry, I fixed a typo and added previews.

chalkygames123:
While there are several possible variations, I prefer to place the brand name on the left and the status on the right, referring to a similar discussion, https://github.com/renovatebot/app-support/issues/38.

travi:
again, apologies for missing this until now. 

> While there are several possible variations, I prefer to place the brand name on the left and the status on the right

i [went back and forth](https://github.com/semantic-release/semantic-release/pull/2235#issue-1046939669) on this as well. i was mainly trying to get just the icon on the left and the project name on the right, but was unsuccessful. i didnt really consider another word on the right, like "enabled", but like the approach. from a branding perspective, keeping the logo with the "signature" is usually a better approach anyway. i'm not sure that "enabled" feels like the right word to me, though. "enabled" feels like a simple on/off toggle to me, and semantic-release is a little bit more involved than that. is there another word that might be more appropriate, or am i over-thinking this?

> does anyone know where the color `#e10079` used in the current one comes from?

[...]

PR #2235, opened 2021-11-08
docs(badge): switched to proper semantic-release logo

travi:
shields.io now has our official logo available through `simple-icons`. 

i was trying to get just the logo on the dark part of the left, similar to the previous version, but was unsuccessful without text content on both sections. i dont hate having `semantic` on the left section and `release` on the right section, but removing the `-` steps away from the official name. i'm interested in thoughts around how this is actually rendering.

also, i'm not sure what the history is around the color used for the badge. is it still the most appropriate for the project? i have no reason to change it, but thought this was a good chance to at least confirm.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,10 +142,10 @@ In order to use **semantic-release** you need:
 
 Let people know that your package is published using **semantic-release** by including this badge in your readme.
 
-[![semantic-release](https://img.shields.io/badge/%20%20%F0%9F%93%A6%F0%9F%9A%80-semantic--release-e10079.svg)](https://github.com/semantic-release/semantic-release)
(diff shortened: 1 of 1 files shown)


gr2m (approved):
Nice 👍🏼

chalkygames123:
@travi @gr2m Hi, I opened https://github.com/semantic-release/semantic-release/issues/1825 about eight months ago, and I happened to find this PR today :) Could you take a look at it?

And probably we should also update the badge at the top of the README.

travi:
thanks for the ping @chalkygames123! apologies for missing your original post. clearly it is a topic that we find value in. i'll respond in the other issue shortly.

> And probably we should also update the badge at the top of the README.

good catch. i obviously overlooked that in my update, but i agree that we should get it up to date. would be happy to accept a contribution to take care of that depending on where we land on the other discussion.

PR #2253, opened 2021-11-20
docs(recipes): fix path to recipes

mamachanko:
* also removes link to unknown “package managers and languages” recipes


travi (approved):
Thank you!

PR #2266, opened 2021-11-25
docs: update badges to include preset names

chalkygames123:
From https://github.com/semantic-release/semantic-release/issues/1825#issuecomment-978119923:

> @chalkygames123 are you interested in opening a PR that updates the readme badge and the docs suggesting the badge to use in projects using semantic-release?

@travi Well, about the latter, honestly, I found it kind of hard to write because English is not my first language 😅 So for the sake of accuracy, would you mind if I leave it to you?

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
     <img alt="Build states" src="https://github.com/semantic-release/semantic-release/workflows/Test/badge.svg">
   </a>
   <a href="#badge">
-    <img alt="semantic-release" src="https://img.shields.io/badge/%20%20%F0%9F%93%A6%F0%9F%9A%80-semantic--release-e10079.svg">
+    <img alt="semantic-release: angular" src="https://img.shields.io/badge/semantic--release-angular-e10079?logo=semantic-release">
   </a>
 </p>
 <p align="center">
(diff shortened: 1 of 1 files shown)


chalkygames123:
And it's worth mentioning that the alt text now also includes the preset name.

chalkygames123:
I might have misunderstood your question. Please just merge if there's no problem with the current changes 🙏

README.md
```diff
@@ -8,7 +8,7 @@
     <img alt="Build states" src="https://github.com/semantic-release/semantic-release/workflows/Test/badge.svg">
   </a>
   <a href="#badge">
-    <img alt="semantic-release" src="https://img.shields.io/badge/%20%20%F0%9F%93%A6%F0%9F%9A%80-semantic--release-e10079.svg">
+    <img alt="semantic-release: conventionalcommits" src="https://img.shields.io/badge/semantic--release-conventionalcommits-e10079?logo=semantic-release">
```
travi, line 11: the current default preset, which is used by this project, is the `angular` preset rather than `conventionalcommits`. i guess this highlights the value that this badge can add 😄

README.md
```diff
@@ -141,10 +141,11 @@ In order to use **semantic-release** you need:
 
 Let people know that your package is published using **semantic-release** by including this badge in your readme.
 
-[![semantic-release](https://img.shields.io/badge/semantic-release-e10079.svg?logo=semantic-release)](https://github.com/semantic-release/semantic-release)
+[![semantic-release: conventionalcommits](https://img.shields.io/badge/semantic--release-conventionalcommits-e10079?logo=semantic-release)](https://github.com/semantic-release/semantic-release)
```
travi, line 144: since the default preset is `angular` i think we should suggest that one here too. we do plan to update the default to `conventional-commits` at some point, but we can update that here at that point

travi (changes requested):
just an update to the convention mentioned. i think this is good to merge with that update

travi:
>  about the latter, honestly, I found it kind of hard to write because English is not my first language 😅 So for the sake of accuracy, would you mind if I leave it to you?

no worries as all. i'm happy to add a bit of detail once we get your changes merged in. i just wanted to make sure you had the first chance if you were interested. 

thank you so much for moving this forward for us!

README.md
chalkygames123: Oops sorry, fixed!

README.md
chalkygames123: Fixed too.

chalkygames123:
Thank you for your kind words ☺️

travi (approved):

PR #2283, opened 2021-12-02
docs(breaking-change): highlighted the need for `BREAKING CHANGE: ` to be in the commit footer

travi:
since this is a common confusion point for new folks


gr2m (approved):
good call 👍🏼

PR #2301, opened 2021-12-16
fix(docs): removing GITHUB_TOKEN from auth vars

YassinEldeeb:
GITHUB_TOKEN is an invalid secret name cause:
> Github: secret names must not start with GITHUB_

this is a github restriction.


gr2m:
> this is a github restriction.

Yes, but GitHub Actions is not the only CI environment where you can run semantic-release

YassinEldeeb:
Ooos missed that

PR #2307, opened 2021-12-21
docs: wrong prerelease example

maxiride:
The documentation says:

> If the `prerelease` property is set to `true` the `name` value will be used.

But then the example referring to this behaviour has the the `prerelease` value set to `'beta'` instead of `true`.


gr2m (approved):
Thank you!

PR #2365, opened 2022-02-15
Correct circleci workflow

cades:
when I tried to apply [example Circle CI config](https://github.com/semantic-release/semantic-release/blob/master/docs/recipes/ci-configurations/circleci-workflows.md) to my config.yml, I found it contains a YAML syntax error, which is fixed in d2bc910.

And in a2e6ff1, the circleci/node version is upgraded and an unnecessary step is removed.

### DIFF
--- a/docs/recipes/ci-configurations/circleci-workflows.md
+++ b/docs/recipes/ci-configurations/circleci-workflows.md
@@ -22,14 +22,12 @@ Finally, we call our release job with a `requires` parameter so that `release` w
 ```yaml
 version: 2.1
 orbs:
-  node: circleci/node@4.5
+  node: circleci/node@5.0.0
 jobs:
   release:
     executor: node/default
     steps:
       - checkout
-      - node/install:
-          lts: true
       - node/install-packages # Install and automatically cache packages
       # Run optional required steps before releasing
       # - run: npm run build-script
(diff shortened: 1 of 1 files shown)


docs/recipes/ci-configurations/circleci-workflows.md
```diff
@@ -22,14 +22,12 @@ Finally, we call our release job with a `requires` parameter so that `release` w
...
 version: 2.1
 orbs:
-  node: circleci/node@4.5
+  node: circleci/node@5.0.0
 jobs:
   release:
     executor: node/default
     steps:
       - checkout
-      - node/install
```
travi, line 31: the reason we include this step is to encourage using the latest lts version of node when releasing, even if that is not the version used to verify the project. does removing this step still accomplish that goal? if not would including a comment make that goal more clear here?

cades: Thanks for replying!

The reason why that step can be removed is, `circleci/node`'s default execute already use lts node docker image.

(There is a configurable tag, which defaults to 'lts')


Reference: https://circleci.com/developer/orbs/orb/circleci/node#executors-default

travi (approved):
thanks!

Issue #1681, opened 2020-11-09
remove limitation on number of release branches


admir86:
## New feature motivation

We are migrating from [TFVC](https://docs.microsoft.com/en-us/azure/devops/repos/tfvc/?view=azure-devops) to git and are also currently changing our workflow. We have decided that the [release flow](https://docs.microsoft.com/en-us/azure/devops/learn/devops-at-microsoft/release-flow) ([Trunk Based Development](https://trunkbaseddevelopment.com/)) suits us best.
I did some tests with semantic-release in our repo and apart from the restriction of a maximum of 3 release branches, semantic-release could cover everything we need. Even in the monorepo ([nx workspace](https://nx.dev/)) with some custom steps through the @semantic-release/exec plugin.

Using the [maintenance branches configuration](https://semantic-release.gitbook.io/semantic-release/usage/workflow-configuration#maintenance-branches) is not a solution for us. Put simply, our release branch is also the maintenance branch. We do not want to create a new release branch for a hotfix. And above all not from another release branch (release branches are only created from the main branch). For a hotfix we use cherry pick to get a fix from the main into an existing release branch e.g. releases/1.0.x.

That's how the branches config looks like:
```json
"branches": [
    {
      "name": "releases/*",
      "channel": ""
    },
    {
      "name": "main",
      "channel": "rc",
      "prerelease": "rc"
    }
  ]
```
We would always have to delete old release branches, but the last 3 releases is not enough for us.

What is the reason for this limitation? Would it be conceivable to overwrite the limitation via configuration if you do not want to remove it?

possibly related to https://github.com/semantic-release/semantic-release/issues/1529

klopfdreh:
This feature is important for us as well. It is opened since 11/09/2020 - anyone of the devs could have a look at it?

rasander:
+1 on this one. We are also strugling a bit with the 3 releasebranch limit.

travi:
what sorts of workflows are you all using that requires more active release branches? 

as far as the specific examples from OP, semantic-release officially does not support monorepos and, while trunk-based development is what we encourage, we specifically support [release from trunk](https://trunkbaseddevelopment.com/release-from-trunk/) and not [branch for release](https://trunkbaseddevelopment.com/branch-for-release/). 

[...]

PR #2405, opened 2022-03-30
fix: remove upper limit on number of release branches

tankers746:
https://github.com/semantic-release/semantic-release/issues/1681

Hi,

As per the issue above there are scenarios where you may want more than 3 release branches.
I'm not sure where the number 3 comes from as I was unable to find it in the documentation.

Thank you

### DIFF
--- a/lib/definitions/branches.js
+++ b/lib/definitions/branches.js
@@ -17,7 +17,7 @@ const prerelease = {
 
 const release = {
   filter: (branch) => !maintenance.filter(branch) && !prerelease.filter(branch),
-  branchesValidator: (branches) => branches.length <= 3 && branches.length > 0,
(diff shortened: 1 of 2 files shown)


klopfdreh:
Hey @tankers746 - I would add a configuration for the limit instead of removing it completely. We already got the branch object in there, so maybe something like:

```javascript
{
  "branches": [
    {"name": "release/*", "prerelease": false, "channel": false, limit: 3},
    {"name": "beta/*", "prerelease": true, "channel": false},
    {"name": "alpha/*", "prerelease": true, "channel": false}
  ],
...
}
```

and then add a convenient method that filters for the release branch definition and defaults to 3 if the option is not set.

travi:
closing for the time being. before moving forward with a code change proposal, we need to reach consensus on a goal in #1681

PR #2416, opened 2022-04-14
fix: get correct version if prerelease branch shares version with ...

### LINKED ISSUE #2415: Release picks wrong version when several previous (pre)releases share same HEAD
### Discussed in https://github.com/semantic-release/semantic-release/discussions/2414

<div type='discussions-op-text'>

<sup>Originally posted by **Tlepel** April 12, 2022</sup>
Hi everyone!

In my project there are 3 branches: `production`, `development` and `beta`. We've only recently started releasing on the alpha branch, in addition to the releases on the beta and production branches where we've been releasing for a few years now.

We've released a version to the alpha branch, let's say `v2.2.0-development.3`, which worked perfectly fine. However, when all seemed well we merged the alpha branch to beta and tried to release there. We don't use merge commits for this, so the `v2.2.0-beta.1` tag was created on the same commit as the `v2.2.0-development.3` tag. Now we've pushed a few commits to the beta branch that fixed some issues on that branch, and tried to release that. However, the tag that was generated was `v2.2.0-development.4` instead of `v2.2.0-beta.2`..

I checked the git notes to see if it went wrong there, and I tried to adapt it to reflect the situation, but that didn't change the version at all; none of these options worked:

```
{"channels": ["development","beta"]}
{"channels": ["beta"]}
{"channels": ["development"]}
```

### Schematic:

```
|
* - v2.2.0-development.1
|
* - v2.2.0-development.2
|
* - v2.2.0-development.3, v2.2.0-beta.1
| \
*  \   - v2.2.0-development.4 (on development branch)
|   |
|   *  - v2.2.0-development.4 (on beta branch, should be v.2.2.0-beta.2)
```

### Setup:

```js
branches: [
  'production',
  {
    name: 'beta',
    prerelease: true,
  },
  {
    name: 'development',
    prerelease: true,
  },
],
```

Can anyone help me out and see what we're doing wrong here?</div>

It seems like the current implementation doesn't account for multiple tags/versions using the same git-commit.

Tlepel:
…other branches

Fixes #2415

### DIFF
--- a/lib/get-last-release.js
+++ b/lib/get-last-release.js
@@ -30,7 +30,11 @@ export default ({ branch, options: { tagFormat } }, { before } = {}) => {
   const [{ version, gitTag, channels } = {}] = branch.tags
     .filter(
       (tag) =>
-        ((branch.type === "prerelease" && tag.channels.some((channel) => isSameChannel(branch.channel, channel))) ||
+        ((branch.type === "prerelease" &&
+          tag.channels.some((channel) => isSameChannel(branch.channel, channel)) &&
+          semver
+            .parse(tag.version)
+            .prerelease.includes(branch.prerelease === true ? branch.name : branch.prerelease)) ||
           !semver.prerelease(tag.version)) &&
         (isUndefined(before) || semver.lt(tag.version, before))
     )
--- a/test/get-last-release.test.js
+++ b/test/get-last-release.test.js
@@ -43,6 +43,43 @@ test("Get the highest prerelease valid tag, ignoring other tags from other prere
   });
 });
 
+test("Get the correct prerelease tag, when other prereleases share the same git HEAD", (t) => {
+  const testConfig = {
+    branch: {
+      name: "alpha",
+      prerelease: "alpha",
+      channel: "alpha",
+      tags: [
+        { version: "1.0.0-beta.1", gitTag: "v1.0.0-beta.1", gitHead: "v1.0.0-beta.1", channels: ["beta"] },
+        { version: "1.0.0-beta.2", gitTag: "v1.0.0-beta.2", gitHead: "v1.0.0-alpha.1", channels: ["alpha", "beta"] },
+        { version: "1.0.0-alpha.1", gitTag: "v1.0.0-alpha.1", gitHead: "v1.0.0-alpha.1", channels: ["alpha", "beta"] },
+      ],
+      type: "prerelease",
+    },
+    options: { tagFormat: `v\${version}`, debug: true },
+  };
+  const firstResult = getLastRelease(testConfig);
+
+  t.deepEqual(firstResult, {
+    version: "1.0.0-alpha.1",
+    gitTag: "v1.0.0-alpha.1",
+    name: "v1.0.0-alpha.1",
+    gitHead: "v1.0.0-alpha.1",
+    channels: ["alpha", "beta"],
+  });
+
+  testConfig.branch.prerelease = true;
+  const secondResult = getLastRelease(testConfig);
+
+  t.deepEqual(secondResult, {
+    version: "1.0.0-alpha.1",
+    gitTag: "v1.0.0-alpha.1",
+    name: "v1.0.0-alpha.1",
+    gitHead: "v1.0.0-alpha.1",
+    channels: ["alpha", "beta"],
+  });
+});
+
 test("Return empty object if no valid tag is found", (t) => {
   const result = getLastRelease({
     branch: {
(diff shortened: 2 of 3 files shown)


test/get-last-release.test.js
```diff
@@ -43,6 +43,31 @@ test('Get the highest prerelease valid tag, ignoring other tags from other prere
   });
 });
 
+test('Get the hightest prerelease valid tag, when other tags share the same HEAD', (t) => {
```
travi, line 46: this test name doesnt make the goal clear to me. semantic-release doesnt really consider the relationship between typical pre-release stages, so if "highest" is referring to promotion through those stages, i don't think that is worth highlighting here. the real goal is to ensure that the pre-release version is properly with the pre-release defined for that branch, correct?

Tlepel: Yeah, that's correct. The goal is more to ensure that the pre-release version corresponds with the pre-release for the branch indeed. I'll change it out to reflect that!

lib/get-last-release.js
```diff
@@ -30,7 +30,9 @@ module.exports = ({branch, options: {tagFormat}}, {before} = {}) => {
   const [{version, gitTag, channels} = {}] = branch.tags
     .filter(
       (tag) =>
-        ((branch.type === 'prerelease' && tag.channels.some((channel) => isSameChannel(branch.channel, channel))) ||
+        ((branch.type === 'prerelease' &&
+          tag.channels.some((channel) => isSameChannel(branch.channel, channel)) &&
+          semver.parse(tag.version).prerelease.includes(branch.name)) ||
```
limulus, line 35: Is it correct to use `branch.name` here? My concern is what if the branch is configured to use a different prerelease identifier for the version.

Would this make more sense?

```suggestion
          semver.parse(tag.version).prerelease.includes(branch.prerelease)) ||
```

limulus (commented):
Hey @Tlepel! Thanks for making this PR, I just happened to run into the same issue. I think I found a potential issue with this PR though. (See the code comment.)

lib/get-last-release.js
Tlepel: Thanks, good catch!  I totally missed the fact that the `prerelease` property could be anything else than a `boolean`, as that's the only way it is used in the docs and in my usecase.

I'll change it next week when I'll have access to my laptop again.

Tlepel: I changed the implementation accordingly, only had to add a test-case to check that it still would work if the prerelease-value was `true` instead of a `string` value, that's why the test might look messier than before.

shiipou (approved):
LGTM

Tlepel:
@travi whould you have time to review this PR? I try to rebase every time commits were merged to `master`, but to be fair I don't feel like keeping that going. I'd love to get some feedback if the PR is not up to standards, and improve the code if necessary, since it seems like more people might need this fix.

taorepoara:
We use two pre-release channels on our projects: alpha and beta.

We used to have a problem since channels are stored in the tag notes: when a commit is released to the alpha channel and then the beta channel, every thing is fine, but when another commit is released to the alpha channel (if the previous is not released in the main branch), the previous alpha release is not found and it try to create another alpha release with the same version...

This PR fixes our issues.

Langstra:
Could we get this reviewed and merged? 😇

taorepoara:
For those who are tired to wait for this merge I fond another project that could do the job: convco. They have just [added the prerelease management](https://github.com/convco/convco/pull/105). It does not work the same as Semantic Release, but it should work fine.

frnode:
I have same issue... Anyone if we can work around the problem temporarily?
thank you

taorepoara:
> I have same issue... Anyone if we can work around the problem temporarily? thank you

You can use the @Tlepel version instead of the official one.

Instead of installing semantic-release official version use the next code:

```bash
npm i "https://github.com/Tlepel/semantic-release.git#fix-versions-share-head"
```

The problem with this solution is that the this PR is **FINALLY** merged, you'll have to change your CI back to the official version of semantic release, since this branch will be closed.

Tlepel:
@gr2m Perhaps you can have a look at this PR? It seems that quite some projects run into this issue and the fix isn't too complicated. @travi is on the list of reviewers but he hasn't reviewed this PR after his initial review on the day I created this PR.

renestalder:
Is he going to break the record for the longest ever rebase run? :)

Tlepel:
> Is he going to break the record for the longest ever rebase run? :)

I sure hope not 😅

gr2m (approved):
Alright sorry for the wait! We hoped to find the time to do a deeper dive into potential unwanted side effects, but at this point all your contributions (and patience) outweigh the risk. Thank you all 💐💝

PR #2426, opened 2022-04-30
feat: add --skip-remote-check option

### LINKED ISSUE #2232: `--dry-run` shouldn't test push permissions
## New feature motivation

Or at least we should be able to disable the validation of push permissions.
When trying to validate a release in the context of a CI we do this with the least amount of permissions possible.

Running `semantic-release --dry-run` stops simply because it cannot push to the underlying repo. 

Requiring these kind of permissions in the current security landscape where we run with the least permissions possible. 

## New feature description
Separate the validation of credentials and the dry-running to allow for a reduced set of permissions passed in.

## New feature implementation

Separating the 2 allows the user to pass in both flags if they wish to retain current behavior.

OR we could introduce a new switch that disables push validation as part of the dry-run.

If the team feels that this will be helpful, I'm happy to take a stab at it.

thislooksfun:
This is a first draft of a new option to bypass checking remote access. It only works in dry run mode. There is still some work that needs to be done before it's ready to merge:

- [ ] Write a concise description
- [ ] Add tests
- [ ] Make `run` throw instead of returning `false`?

I took a first stab at writing some tests for this new flag, but I got lost in the tests and couldn't figure out how to make some of them work. If someone with more experience in this test suite could help out that would be greatly appreciated.

Closes #2232.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -34,6 +34,7 @@ Usage:
     .option('success', {...stringList, group: 'Plugins'})
     .option('fail', {...stringList, group: 'Plugins'})
     .option('debug', {describe: 'Output debugging information', type: 'boolean', group: 'Options'})
+    .option('skip-remote-check', {describe: 'TODO', type: 'boolean', group: 'Options'})
     .option('d', {alias: 'dry-run', describe: 'Skip publishing', type: 'boolean', group: 'Options'})
     .option('h', {alias: 'help', group: 'Options'})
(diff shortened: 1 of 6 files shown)


Pinappleplayer (approved):

codejedi365:
Any movement on this? I would love to see this come to fruition.

boxrick:
This would also help my current issue!

PR #2465, opened 2022-06-13
docs(configuration.md): `addChannel` step is skipped when `dryRun` is enabled

csantos1113:
After a lot of suffering trying to debug some issues with semantic-release, I finally noticed `addChannel` is also skipped on dryRun mode, but it isn't documented 

https://github.com/semantic-release/semantic-release/blob/master/lib/definitions/plugins.js#L85


gr2m (approved):
thank you! Sorry for all the headache that gave you

PR #2487, opened 2022-06-23
docs(plugins): add semantic-release-mirror-version to plugins

GarthDB:


gr2m (commented):
Happy to add your plugin! Looks similar to https://github.com/gr2m/semantic-release-plugin-update-version-in-files/, but no problem to have multiple plugins that are similar in functionality

Tunealb (approved):
Okay 👍

babblebey (approved):

PR #2498, opened 2022-07-06
docs: remove deprecated keyword in gitlab-ci.md

iczellion:
The usage of only/except in Gitlab pipelines is not recommended to use for newer pipelines according to this documentation: https://docs.gitlab.com/ee/ci/yaml/#only--except

We should update the documentation accordingly.


travi (approved):
thanks

PR #2502, opened 2022-07-09
docs(plugin-list): add semantic-release-space plugin

123FLO321:
The [semantic-release-space](https://github.com/123FLO321/semantic-release-space) plugin allows creating JetBrains Space Deployments with semantic-release.


gr2m (approved):
Thank you!

Issue #2478, opened 2022-06-17
Integration tests fail to execute properly on ARM CPU architecture


travi:
### Current behavior

the tests timeout do to docker failing to start the [docker-gitbox image](https://github.com/semantic-release/semantic-release/blob/9589a96239826abe9b07e8deffcc7d8aeb9c2e40/test/helpers/gitbox.js#L6) that we use. this is due to it not supporting the ARM architecture. 

I made some progress resolving the architecture limitation in https://github.com/semantic-release/docker-gitbox/issues/4, but have not resolved all of the issues to enable the integration tests to run successfully in this project yet.

```
✖ integration › before hook Rejected promise returned by test
✖ integration › after.always hook Rejected promise returned by test
✖ 14 tests remaining in test/integration.test.js
```

```
integration › before hook

  Rejected promise returned by test. Reason:

  Error {
    address: '/var/run/docker.sock',
    code: 'ENOENT',
    errno: -2,
    syscall: 'connect',
    message: 'connect ENOENT /var/run/docker.sock',
  }



  integration › after.always hook

  Rejected promise returned by test. Reason:

  TypeError {
    message: 'Cannot read properties of undefined (reading \'stop\')',
  }

  › Object.stop (test/helpers/gitbox.js:42:19)
  › test/integration.test.js:62:29

  ─

  2 hooks failed
```

### Expected behavior

the full test suite should execute without timing out. that means that the docker images must be usable on both x86 and ARM architectures. 

this could be enabled by updating https://github.com/semantic-release/docker-gitbox to work properly or by abandoning that image and finding a more modern one that is supported by the community rather than this team, but still satisfies the needs we have for these tests.

### `semantic-release` version

latest

### CI environment

local

### `semantic-release` configuration

N/A

### CI logs

see details from https://github.com/semantic-release/docker-gitbox/issues/4

travi:
since our maintainers are all using ARM based machines at this point, this is pretty blocking for making progress on the project. our docker skills to work through this are also quite rusty, so if someone with strength in docker would be interested in helping us work through this, it could be a big help to the project.

travi:
once #2516 merges, `mockserver` appears to be the next (and hopefully final) blocker for this. there is already an [open issue](https://github.com/mock-server/mockserver/issues/819) regarding building that image for arm.

fgreinacher:
Send a PR for arm64 support in `mockserver`: https://github.com/mock-server/mockserver/pull/1419

PR #2516, opened 2022-07-26
test(integration): upgrade the gitbox image to support arm architecture in addition to x86

travi:
while the gitbox image now appears to be working properly, the mockserver image is still having
problems, so this does not fully get the tests working on arm yet

for #2478


gr2m (approved):
thanks for moving forward on this <3

fgreinacher (approved):
LGTM 👍

travi:
> thanks for moving forward on this <3

big thanks to @jsoref in semantic-release/docker-gitbox#5 for making this step possible

PR #2523, opened 2022-08-13
docs: add semantic-release-react-native to plugin list

alexandermendes:
A new plugin for React Native versioning.


travi (approved):
Thanks!

alexandermendes:
Oops, in fact I just realised a small error there, `publish` should be `prepare`, updating 👀

PR #2547, opened 2022-08-31
docs: document that .cjs extensions are supported for config files

davidjgoss:
`.cjs` config files are supported since https://github.com/semantic-release/semantic-release/pull/1815, which is essential for ESM projects - just adding this to the docs.


travi (approved):
Thanks!

PR #2553, opened 2022-09-04
Remove jargon from README

getsnoopy:


travi:
Please explain your motivation for this change. From what you have provided, I do not see it as an improvement

getsnoopy:
It's summarized pretty well in the commit subject. "Impact" is jargon; the changes are not crashing into consumers. Even in the jargon sense of "strong/violent/marked effect", it doesn't make sense because the package is not only meant to summarize the strong/violent/marked effects, but *all* the effects of the changes on consumers. It's just poor word choice, and my change improves it to use non-jargon language.

travi:
I think I may better understand your concern with that description, but it was not at all clear from your original post. The use of "impact" here is intended to highlight that communicating the semantic impact is simplified. semantic-release simplifies the use of semantic versioning to communicate whether the combination of changes in a release add up to a fix, new feature, or breaking change. The places you are changing are not referring to listing the individual changes, but how those changes impact consumers. 

I don't see that as jargony, but instead as referring to the purpose of communicating to consumers with semantic versioning. I would be open to a contribution that clarified that meaning further (maybe "semantic impact" is better than "impact"?), but I think your current suggestion changes what is communicated in a way that no longer suggests the intended meaning.

getsnoopy:
Ah, I better understand what you mean now. But yes, that is the very thing I'm talking about: when you say "but how those changes impact consumers", I'm talking about changing that to use non-jargon language: "how those changes *affect* consumers". The changes are not crashing into consumers, and even in the jargon sense of "strongly/violently/markedly affect", it makes no sense for the package description to describe how strongly the changes affect the consumers, since that is a judgement the consumers can make for themselves. That's what I mean by jargon; just replacing the jargon word "impact" with the non-jargon word "affect" (in the verb sense) or "effect" (in the noun sense).

travi:
thanks for the contribution, but i dont agree that the use of "impact" in these cases is incorrect enough to justify a change here. as i mentioned before, i would possibly be open to accepting a contribution that makes it more clear that these statements refer to the semantic version level, but i think that should be a separate PR at this point if someone would like to contribute it.

getsnoopy:
I'm not sure what there is to disagree about, seeing as it's the [definition of the word](https://www.merriam-webster.com/dictionary/impact), or why there's resistance to remove jargon from the documentation, but OK.

Issue #2543, opened 2022-08-27
ES Module


travi:
Part of our efforts to migrate `semantic-release` to native ES Modules: https://github.com/semantic-release/semantic-release/issues/2133

travi:
the updates for core are complete as far as we know at this point, so we plan to promote the latest beta to stable very soon. if anyone is eager to try out the changes before we promote to latest, give the latest beta version a shot and let us know here  if you find any problems.

PR #2558, opened 2022-09-12
refactor(plugins): switched from `require` to `await import()` when loading plugins

travi:
to isolate the conversion ahead of #2543


travi:
this obviously still needs clean up, but hoping to get some feedback. i'm not 100% i have my head around all that needs to change here, but assuming we change from `require` to `await import()`, the majority of the work should be handling the additional promises that are introduced. i think i have most of that handled, but may have missed some.

however, the part that i did not anticipate is that the result does not appear to satisfy [`_.isPlainObject()`](https://lodash.com/docs/3.10.1#isPlainObject) the same way that loading with `require` did. it does not appear to be an issue with the added `default` property either.

logging one of the example plugins from the tests shows the following, which i would expect to both satisfy `_.isPlainObject` (at least because of the `null` prototype) and satisfy the expected shape if it were to make it far enough through the logic
```
[Module: null prototype] {
  analyzeCommits: [AsyncFunction: analyzeCommits],
  default: { analyzeCommits: [AsyncFunction: analyzeCommits] }
}
```
but it appears that the loaded module has some detail that i havent yet figured out that prevents it from satisfying `_.isPlainObject()`

travi:
this also got me thinking about how this situation of loading plugins potentially gets more complicated as we progress further. currently, only cjs plugins are expected. once we move to esm, we will want to support loading plugins that are also esm. loading with `await import()` likely normalizes that situation for us, but we will want to confirm it once we are past this part.

this partially comes to mind because i did consider if it could be worth considering using [`module.createRequire`](https://nodejs.org/api/module.html#modulecreaterequirefilename) to avoid all of the complexity, but that would prevent supporting esm plugins. i dont think that even makes sense as an intermediate step since we will want to convert official plugins to esm as soon as converting core to esm unblocks converting the plugins.

lib/plugins/index.js
```diff
@@ -6,36 +6,41 @@ const {validatePlugin, validateStep, loadPlugin, parseConfig} = require('./utils
...
-                });
-                plugins[type] = [...(plugins[type] || []), [func, config]];
-              }
-            });
-          } else {
-            errors.push(getError('EPLUGINSCONF', {plugin}));
-          }
+        console.log({plugins: options.plugins})
+        console.dir(plugin, {depth: null})
+        console.log(isPlainObject(plugin))
```
gr2m, line 22: just a reminder to remove the debug logs when the PR is ready :)

lib/plugins/index.js
```diff
@@ -6,36 +6,41 @@ const {validatePlugin, validateStep, loadPlugin, parseConfig} = require('./utils
...
   let {options, logger} = context;
   const errors = [];
 
   const plugins = options.plugins
-    ? castArray(options.plugins).reduce((plugins, plugin) => {
-        if (validatePlugin(plugin)) {
-          const [name, config] = parseConfig(plugin);
-          plugin = isString(name) ? loadPlugin(context, name, pluginsPath) : name;
+    ? await castArray(options.plugins).reduce(async (plugins, plugin) => {
+      const plugins2 = await plugins;
```
gr2m, line 15: I wonder if there is a better name for `plugins2`? Maybe `normalizedPlugins` or `awaitedPlugins`?

gr2m (commented):
changes look good to me thus far

lib/plugins/index.js
travi: yep, this is definitely part of the clean up that is necessary once this is working :)

lib/plugins/index.js
travi: yep 👍🏼

travi:
i was able to resolve the previous issues by fixing the paren wrapping that i broke (this was the main problem) and using the `.default` property of the imported module (which i had tried previously, but wasnt working fully because of the former problem

travi:
however, `xo` now reports a lint issue for the dynamic import expression:
```
lib/plugins/utils.js:52:43
  ✖  52:43  import() expressions are not supported yet.  node/no-unsupported-features/es-syntax
```

however, all tests, including integration tests pass in our supported node range from my checking. all scenarios within the dynamic import expression seem to be covered by tests as well.

based on https://github.com/mysticatea/eslint-plugin-node/issues/250, it sounds like the rule does not report correctly for this situation and is unlikely to be fixed soon. i hate to disable a rule like this and lose protection from other scenarios, but i'm unsure of a better approach for now. i'm open to suggestions

travi:
i went ahead and disabled the rule, since it seems like the best option, but open to alternative suggestions

gr2m:
> i hate to disable a rule like this and lose protection from other scenarios, but i'm unsure of a better approach for now. i'm open to suggestions

+1 for disabling. We can create a follow up issue to review once we concluded the ESM transition?

Great work @travi 👏🏼

PR #2566, opened 2022-09-20
ci(permissions): provided only the necessary permissions to the `GITHUB_TOKEN`

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.


travi:
i'm supportive of restricting the permissions along these lines, but i think this is currently incomplete. 

for both workflows, the global permissions should be `contents: read` to only allow checkout. 

the permissions needed for the release should be limited to that job specifically. also, there are other permissions needed to enable all of the release behaviors. please refer to the docs for the github plugin for more details: https://github.com/semantic-release/github#github-authentication

.github/workflows/release.yml
```diff
@@ -6,8 +6,14 @@ name: Release
...
       - beta
       - "*.x"
+permissions:
+  contents: read  #  for checkout
 jobs:
   release:
+    permissions:
+      contents: write       #  to be able to publish a GitHub release
+      issues: write         #  to be able to comment on released issues
+      pull-requests: write  #  to be able to comment on released pull requests
```
gr2m, line 16: why split up permissions, there is only one job. I'd combine the permissions all under the top level `permissions` key

gr2m: okidoke, apologies, totally makes sense, I missed Matt's message

gr2m (approved):

travi:
it looks like prettier isnt happy with some of the formatting in the release workflow definition. would you mind taking a look at resolving that @sashashura. once we get the build fully passing, we can get this merged in

travi (approved):
Thanks!

PR #2606, opened 2022-11-11
forced plugins to use beta version of semantic-release

travi:
for #2543

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,6 +75,9 @@
     "testdouble": "3.16.6",
     "xo": "0.32.1"
   },
+  "overrides": {
+    "semantic-release": "20.0.0-beta.1"
+  },
   "engines": {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -75,6 +75,9 @@
     "testdouble": "3.16.6",
     "xo": "0.32.1"
   },
+  "overrides": {
+    "semantic-release": "20.0.0-beta.1"
+  },
```
travi, line 80: i think we should target removing this before promoting the pre-release of core to stable

gr2m: TIL! I didn't know `overrides` was a thing! It's only supported in npm 8.3+ though, right?
https://www.stefanjudis.com/today-i-learned/how-to-override-your-dependencys-dependencies/

Or was it backported?

gr2m (approved):

PR #2607, opened 2022-11-11
feat(esm): convert to esm (#2569)

travi:
for #2543

remaining tasks:
- [x] get remaining tests passing
		two tests remain:
		* ~~one in `test/index.test.js` related to plugin name using a string~~
		* ~~one in `test/integration.test.js` that might be just related to a change in output from aggregate-error, which was upgraded~~
- [x] xo: remove in favor of prettier?
- [x] raise minimum supported node version to v18
- [x] promote beta of `env-ci` to latest: https://github.com/semantic-release/env-ci/pull/248
- [x] update dependency on env-ci to latest
- [x] remove `overrides` definition in `package.json`
- [x] verify real usage for some period of time
~~- [ ] consider if any plugins that are direct dependencies can/should also be converted to esm (not a requirement, but an option if it makes sense to do while we are waiting to verify enough real usage)~~
~~- [ ] should we raise the minimum required version of git? (search issues for features that might be motivating for this)~~


travi:
successful publishes using the beta:
* `@travi/cli`
	* [v9.0.30](https://github.com/travi/cli/releases/tag/v9.0.30): [build](https://github.com/travi/cli/actions/runs/3448165422/jobs/5754950843)

travi:
Reminder that this is a release branch, so we should not squash when merging into the mainline

travi:
looks like the commit for the node version minimum lost the BREAKING CHANGE token, so we'll want to make sure we note that in the release notes when we promote to stable

travi:
since we get lots of questions about unsupported project structures or workflows, i think it could be worth considering documenting these as unsupported more clearly and maybe call them out as "breaking changes". these wouldnt actually change what is supported, but would give something more clear to point to for those types of requests.

unsupported:
* monorepos
* gitflow and other workflows that include releasing in order to enable testing in consumers before stable releases

supported:
* microrepos
* trunk-based development
* github flow

encourage:
* continuous integration
* continuous releases/continuous deployment
* confidence to integrate and release 

is this going to far to include as part of this release?

travi:
> is this going to far to include as part of this release?

we decided to consider this and the potential raising of the minimum git requirement after promoting the changes from this PR, so those could be considered for a future breaking change rather than the current one

hongaar:
> raise minimum supported node version to v18

If possible it would be nice for users to be able to run this on maintenance + LTS versions (Node 14, 16 and 18). Adoption of latest LTS generally takes a while and CI systems also lag behind in docs/images, e.g.:
- [README for setup-node](https://github.com/actions/setup-node ) uses node 16 in primary example
- Latest [ubuntu image on GH Actions](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#language-and-runtime) is running Node 16

travi:
> If possible it would be nice for users to be able to run this on maintenance + LTS versions (Node 14, 16 and 18)

appreciate the feedback, @hongaar. however, reducing the supported node versions helps the maintenance team significantly and is in line with our [node support policy](https://semantic-release.gitbook.io/semantic-release/support/node-support-policy). most CI services enable releasing with a different version of node than is used for the rest of the pipeline, which we [outline here](https://semantic-release.gitbook.io/semantic-release/support/node-version#recommended-solution). 

if the options outlined there are not possible, our recommendation is to simply continue using the previous version since it has been stable for some time.

gr2m (approved):
Let's gooooo

travi:
After discussion, we've decided to wait until after the holidays to promote this to stable/latest so we don't surprise consumers with the breaking changes during the holidays.

PR #2619, opened 2022-11-19
fix(secrets-masking): used the proper named import from hook-std to enable masking for stderr

travi:
for #2543


gr2m (approved):
Ha! That was a tough one! Glad you figured it out!

PR #2620, opened 2022-11-19
feat(node-versions): raised the minimum required node version to v18

travi:
in accordance with our node support policy:
https://semantic-release.gitbook.io/semantic-release/support/node-support-policy

BREAKING CHANGE: Node v18 or greater is now required. Please refer to
https://semantic-release.gitbook.io/semantic-release/support/node-version for recommended options
for releasing with a supported node version.

for #2543

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,9 +17,8 @@ jobs:
     strategy:
       matrix:
         node-version:
-          - 14.17
-          - 16.0.0
-          - 17
+          - 18.0.0
+          - 19
 
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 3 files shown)


gr2m (approved):
Feels great

benmccann:
This feels a bit too aggressive. I would understand not wanting to test on older versions, but using `engines` to ban people from running on a supported LTS is a bit much in my opinion. Our project still supports Node 16 as its an LTS that's supported until September. Would it be possible to add back Node 16 support in the `engines` field (or alternately to remove the `engines` field and simply drop testing older versions)?

travi:
We recommend releasing from a separate job from the one you do your verification in. This enables verifying in a matrix of the versions supported by your package and then releasing after all have passed. 

https://semantic-release.gitbook.io/semantic-release/support/node-version#recommended-solution

PR #2631, opened 2022-12-05
fix: aggregate error v4

hongaar:
This change makes `utils.extractErrors` compatible with aggregate-error v4, which changes its implementation to store the errors in an `.errors` property instead of making the Error instance iterable. 

See https://github.com/sindresorhus/aggregate-error/commit/02342f97aa10c73fc12a909bb71e1db8caca0791

_Note: could not run the linter due to `Error: Cannot find module 'prettier'`_

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -3,7 +3,11 @@ const semver = require('semver');
 const hideSensitive = require('./hide-sensitive');
 
 function extractErrors(err) {
-  return err && isFunction(err[Symbol.iterator]) ? [...err] : [err];
+  return err.errors && isFunction(err.errors[Symbol.iterator])
+    ? [...err.errors]
+    : err && isFunction(err[Symbol.iterator])
+    ? [...err]
+    : [err];
 }
 
 function hideSensitiveValues(env, objs) {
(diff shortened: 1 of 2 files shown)


travi:
our work toward converting the core project to esm in https://github.com/semantic-release/semantic-release/pull/2607 enables us to upgrade to aggregate-error v4. we are very close to being ready to promote that work to latest. do you see this change as being necessary after that change?

hongaar:
The change itself would still be needed but the test should probably be changed:

```js
// This will now test for the `isFunction(err.errors[Symbol.iterator])` path
t.deepEqual(extractErrors(new AggregateError(errors)), errors);

// Add something like
// Test aggregate-error < v4 which stores errors by making object iterable
const legacyAggregateError = new Error();
legacyAggregateError[Symbol.iterator] = function* () { 
  for (const error of errors) {
    yield error;
  }
}
t.deepEqual(extractErrors(aggregateErrorv4), errors);
```

hongaar:
Closing since latest supports aggregate-error v4. Not sure if there is value in adding backwards compatibility with earlier aggregate-error versions? May be useful when using plugins which depend on earlier versions of aggregate-error. If wanted, I will reopen another PR and add the test from https://github.com/semantic-release/semantic-release/pull/2631#issuecomment-1345395372.

PR #2662, opened 2023-01-10
docs(plugins): set analyzeCommits as not required

aryzing:
The `analyzeCommits` step was marked requiring a plugin, which conflicts with the default behaviour of using `@semantic-release/commit-analyzer` by default. Hence, it is not required.


aryzing:
Closing due to lack of interest

PR #2668, opened 2023-01-12
docs(pre-releases): Correct @latest version in semantic-release

owanturist:
There is a tiny typo in the version mentioned in `Working on another future release` section. It says that installing with `npm install example-module` will still receive version `1.0.0`, but in fact, there was a bug fix made on the default distribution channel one step before.


CharliesAngel1 (approved):

gr2m (approved):
cheers!

PR #2669, opened 2023-01-12
docs: update node LTS version for npx command

shaneog:
The [lts tag](https://www.npmjs.com/package/node?activeTab=versions) refers to Node.js 16.13.2. To use v18 of Node you need to use the tag `v18-lts`.


shaneog:
Confirmation from the maintainer that the `lts` tag will no longer be used: https://github.com/aredridel/node-bin-gen/issues/188#issuecomment-1381207399

gr2m (approved):
Cheers!

PR #2670, opened 2023-01-13
fix(deps): bump `@semantic-release/npm` to `^10.0.0-beta.1`

gr2m:
- fix(deps): bump `@semantic-release/npm` to `^10.0.0-beta.1`
- build(package): lock file

### DIFF
--- a/lib/plugins/utils.js
+++ b/lib/plugins/utils.js
@@ -53,7 +53,19 @@ export async function loadPlugin({ cwd }, name, pluginsPath) {
     : __dirname;
 
   // See https://github.com/mysticatea/eslint-plugin-node/issues/250
-  return isFunction(name) ? name : (await import(resolveFrom.silent(basePath, name) || resolveFrom(cwd, name))).default;
+  if (isFunction(name)) {
+    return name;
+  }
+
+  const { default: cjsExport, ...esmNamedExports } = await import(
+    resolveFrom.silent(basePath, name) || resolveFrom(cwd, name)
+  );
+
+  if (cjsExport) {
+    return cjsExport;
+  }
+
+  return esmNamedExports;
 }
 
 export function parseConfig(plugin) {
--- a/lib/plugins/utils.js
+++ b/lib/plugins/utils.js
@@ -1,7 +1,7 @@
-import {dirname} from 'node:path';
-import {fileURLToPath} from 'node:url';
-import {castArray, isArray, isFunction, isNil, isPlainObject, isString} from 'lodash-es';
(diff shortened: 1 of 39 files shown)


lib/plugins/utils.js
```diff
@@ -47,14 +47,25 @@ export function validateStep({required}, conf) {
   return conf.length === 0 || validateSteps(conf);
 }
 
-export async function loadPlugin({cwd}, name, pluginsPath) {
+export async function loadPlugin({ cwd }, name, pluginsPath) {
   const basePath = pluginsPath[name]
     ? dirname(resolveFrom.silent(__dirname, pluginsPath[name]) || resolveFrom(cwd, pluginsPath[name]))
     : __dirname;
 
   // See https://github.com/mysticatea/eslint-plugin-node/issues/250
```
travi, line 55: i think this comment can be removed at this point

lib/plugins/utils.js
```diff
@@ -47,14 +47,25 @@ export function validateStep({required}, conf) {
...
+
+  const { default: cjsExport, ...esmNamedExports } = await import(
+    resolveFrom.silent(basePath, name) || resolveFrom(cwd, name)
+  );
+
+  if (cjsExport) {
+    return cjsExport;
+  }
+
+  return esmNamedExports;
```
travi, line 68: nice work!

travi (approved):
noticed one really minor detail to clean up, but otherwise looks good to go

lib/plugins/utils.js
travi: i'm wondering, should we release this as an enhancement against v20 rather than wait for this beta to be promoted to stable? seems like it would be beneficial. maybe we try it here first by releasing as part of this upcoming beta, but could be worth backporting ahead of this becoming stable

gr2m: yeah let's do that later, if needed

lib/plugins/utils.js
gr2m:
```suggestion
```

travi (approved):

travi:
looks like we didnt capture that the npm plugin upgrade was breaking. worth trying to re-release a new beta, @gr2m?

gr2m:
oops yes! I'll do an empty commit to `beta` with this breaking change message:

> legacy authentication using `NPM_USERNAME` and `NPM_PASSWORD` is no longer supported. Use `NPM_TOKEN` instead.

gr2m:
there we go: https://github.com/semantic-release/semantic-release/releases/tag/v21.0.0-beta.1

Issue #2658, opened 2023-01-07
Semantic Release 20 can't detect git installed in Github Action Windows


Belphemur:
### Current behavior

Crashes complaining can't find git binary.

Relating to this dependabot PR: https://github.com/Belphemur/SoundSwitch/pull/1085

### Expected behavior

Find the git binary as it was the case before in the 19.0.5 version.

### `semantic-release` version

20.0.0

### CI environment

GitHub Action

### Plugins used

git

### `semantic-release` configuration

https://github.com/Belphemur/SoundSwitch/blob/dev/package.json

### CI logs

Run npx semantic-release -d
[semantic-release]: Git version 2.7.1 is required. No git binary found.
TypeError: Invalid Version: undefined
    at new SemVer (D:\a\SoundSwitch\SoundSwitch\node_modules\semver\classes\semver.js:19:13)
    at compare (D:\a\SoundSwitch\SoundSwitch\node_modules\semver\functions\compare.js:3:3)
    at lt (D:\a\SoundSwitch\SoundSwitch\node_modules\semver\functions\lt.js:2:29)
    at file:///D:/a/SoundSwitch/SoundSwitch/node_modules/semantic-release/bin/semantic-release.js:29:9
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
Error: Process completed with exit code 1.

webstech:
Wow.  I just hit this as well.  

`Git for Windows` may not fully follow the semver standard (I will open an issue there to ask about it) but the switch of checkers did break things.  

Is this test still needed?  Git 2.7.1 came out in early 2016.  Existing users of this package will be past that version.  New users will likely have an updated tool chain given all the attacks in the past few years.  I would vote to remove the test.

Rossbro2:
So what's the solution to this?  My build is currently failing due to this 😞

webstech:
> So what's the solution to this?

Not sure.  After removing the version test, other `windows` errors occurred.

```
[9:35:47 a.m.] [semantic-release] » ℹ  Running semantic-release version 20.0.2
[9:35:47 a.m.] [semantic-release] » ✘  An error occurred while running semantic-release: Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'
    at new NodeError (node:internal/errors:400:5)
[...]

PR #2672, opened 2023-01-15
fix(windows): fixed issues preventing execution from windows

webstech:
Updates to support running on Windows post the esm conversion.

1. [fix: Detect git version on windows](https://github.com/semantic-release/semantic-release/commit/301687fc0cb9e8caf0999b3926132168826d47be) 
 git does not fully match semver, especially on the windows version. Using loose checking in findVersions is able to match the version.
2. [fix: use file:// url loading plugins](https://github.com/semantic-release/semantic-release/commit/3ec47a33fe98c857d773865187253ebadf766875)
 The `file://` url is required for the `win32`platform but supported on all platforms.

### DIFF
--- a/bin/semantic-release.js
+++ b/bin/semantic-release.js
@@ -25,7 +25,7 @@ See https://github.com/semantic-release/semantic-release/blob/master/docs/suppor
 
 execa("git", ["--version"])
   .then(({ stdout }) => {
-    const gitVersion = findVersions(stdout)[0];
+    const gitVersion = findVersions(stdout, { loose: true })[0];
     if (lt(gitVersion, MIN_GIT_VERSION)) {
       console.error(`[semantic-release]: Git version ${MIN_GIT_VERSION} is required. Found ${gitVersion}.`);
       process.exit(1);
--- a/lib/plugins/utils.js
+++ b/lib/plugins/utils.js
@@ -52,9 +52,14 @@ export async function loadPlugin({cwd}, name, pluginsPath) {
     ? dirname(resolveFrom.silent(__dirname, pluginsPath[name]) || resolveFrom(cwd, pluginsPath[name]))
     : __dirname;
 
-  // See https://github.com/mysticatea/eslint-plugin-node/issues/250
-  // eslint-disable-next-line node/no-unsupported-features/es-syntax
-  return isFunction(name) ? name : (await import(resolveFrom.silent(basePath, name) || resolveFrom(cwd, name))).default;
+  if (!isFunction(name)) {
+    const file = resolveFrom.silent(basePath, name) || resolveFrom(cwd, name);
+    // See https://github.com/mysticatea/eslint-plugin-node/issues/250
+    // eslint-disable-next-line node/no-unsupported-features/es-syntax
+    name = (await import(`file://${file}`)).default;
+  }
+
+  return name;
 }
 
 export function parseConfig(plugin) {


webstech:
The tests fail on windows (mostly - have not checked them all yet) due to [testdouble #491](https://github.com/testdouble/testdouble.js/issues/491).  The issue has been open for awhile.

webstech:
@travi Can you approve the workflow run?  I am researching how the tests fail on windows unrelated to these changes.  That should not stop this PR from moving forward.

travi:
just to confirm, do you mean v20+ in the PR title?

webstech:
Looks like one of the test jobs is hung up somewhere.

lib/plugins/utils.js
```diff
@@ -52,9 +52,14 @@ export async function loadPlugin({cwd}, name, pluginsPath) {
...
     : __dirname;
 
-  // See https://github.com/mysticatea/eslint-plugin-node/issues/250
-  // eslint-disable-next-line node/no-unsupported-features/es-syntax
-  return isFunction(name) ? name : (await import(resolveFrom.silent(basePath, name) || resolveFrom(cwd, name))).default;
+  if (!isFunction(name)) {
+    const file = resolveFrom.silent(basePath, name) || resolveFrom(cwd, name);
+    // See https://github.com/mysticatea/eslint-plugin-node/issues/250
+    // eslint-disable-next-line node/no-unsupported-features/es-syntax
+    name = (await import(`${ file[0] != '.' ? "file://" : "" }${file}`)).default;
```
travi, line 59: so i dont make incorrect assumptions, could you highlight the details of this change?

bin/semantic-release.js
```diff
@@ -25,7 +25,7 @@ See https://github.com/semantic-release/semantic-release/blob/master/docs/suppor
 
 execa("git", ["--version"])
   .then(({ stdout }) => {
-    const gitVersion = findVersions(stdout)[0];
+    const gitVersion = findVersions(stdout, { loose: true })[0];
```
travi, line 28: so we capture details that we can look back to in the future, could you mention the details that lead to this change?

travi (commented):
thanks a lot for the contribution. would it be possible to include some tests that simulate the situations that require these changes?

bin/semantic-release.js
webstech: The output of `git -v` on Windows at my current install is: `git version 2.39.0.windows.2`.  This is not fully semver compliant.  Issue #https://github.com/git-for-windows/git/issues/4086 was opened (and closed) recently regarding this (tl;dr it is not going to change). It appears the semver package works and the find-versions package does not.  The `loose` option extracts the version info from the string correctly.  The non-Windows versions could have similar issues in the future that this will hopefully avoid.  `find-versions` could be updated but a rejection would not be unexpected.

lib/plugins/utils.js
webstech: This change uses the `file://` url schema for import if the file path is not relative (starts with '.').  Node doc indicates this works for import on all platforms.  This is a cheap test instead of using `path.isAbsolute`.  If plugins can be referenced using `http:` schema, this may need to be changed.

webstech:
> would it be possible to include some tests that simulate the situations that require these changes?

Simply running on Windows tests these changes.  Running elsewhere shows they do no harm.  Not sure how the version test can be simulated.

b12k:
Looking forward for this PR to be merged soon.
I'm facing the same issues with git versions and plugins import.

webstech:
Tidying up loose ends.  This is in support of issue #2658.  The issue will track related changes.

webstech:
> Running elsewhere shows they do no harm. 

Is there any reason to hold this back when the tests run on non-Windows?  Not sure what tests I could do for this.  As noted in the related issue, patches will be coming to allow Windows as part of the CI tests.

lib/plugins/utils.js
travi: > This is a cheap test instead of using `path.isAbsolute`.

the benefit of `path.isAbsolute` is that it is self-documenting about why the check is in place. i think i actually prefer that approach here for that reason. any reason not to go that route?

travi:
> Simply running on Windows tests these changes. Running elsewhere shows they do no harm. Not sure how the version test can be simulated.

sorry for the delay responding to this question. i was hopeful that we'd be able to add tests that were specific to the new code paths based on these scenarios, but i finally had a chance to review our existing tests and it looks like the only points to observe those particular behaviors are currently at the system level. so, without executing the tests on windows where the underlying behavior is actually different, i dont think we could actually test these things w/o bigger changes to our tests. 

based on that, lets handle the path.isAbsolute point that i mentioned above and then i think we're ready to get this merged and released.

lib/plugins/utils.js
webstech: > any reason not to go that route?

After looking at the code some more, it appears `file` is always absolute and no check is needed.  I have removed the check so `file://` is always prepended.  Tests under Windows ran fine.

travi: this question did go through my mind, but i dismissed it thinking that loading by package name might be the other case. assuming we can get passing builds on this PR, i'll get it this merged.

travi (approved):

travi:
i'm going to move ahead with merging this even though we havent seen all of the checks pass yet. the tests are not existing correctly in v19 after timing out, which appears to be happening pretty consistently and i have tried restarting several times. those problems are not unique to this branch since they also exist on the mainline and in other PRs.

webstech:
> i'm going to move ahead with merging this

Thank you.

travi:
> > i'm going to move ahead with merging this
> 
> Thank you.

Thank you for this contribution. It was a huge help getting us past these issues.

PR #2679, opened 2023-01-19
docs(workflow-configuration): improved sentence structure

khaled-badenjki:
A correct english sentence needs a verb, which is "be" in this case


travi (commented):
thanks!

travi (approved):

PR #2688, opened 2023-01-25
feat(plugins): add support for loading ESM plugins

travi:
this cherry-picks commits that originally landed in the current beta branch in order to backport the esm plugin loading capability into the mainline ahead of the beta branch landing.

### DIFF
--- a/lib/plugins/utils.js
+++ b/lib/plugins/utils.js
@@ -52,14 +52,18 @@ export async function loadPlugin({cwd}, name, pluginsPath) {
     ? dirname(resolveFrom.silent(__dirname, pluginsPath[name]) || resolveFrom(cwd, pluginsPath[name]))
     : __dirname;
 
-  if (!isFunction(name)) {
-    const file = resolveFrom.silent(basePath, name) || resolveFrom(cwd, name);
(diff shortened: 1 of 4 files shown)


lib/plugins/utils.js
```diff
@@ -52,14 +52,18 @@ export async function loadPlugin({cwd}, name, pluginsPath) {
...
 
-  return name;
+  const file = resolveFrom.silent(basePath, name) || resolveFrom(cwd, name);
+  const { default: cjsExport, ...esmNamedExports } = await import(`file://${file}`);
+
+  if (cjsExport) {
+    return cjsExport;
+  }
+
+  return esmNamedExports;
```
travi, line 66: this implementation has changed slightly because of merging it with the fix for windows support of v20

gr2m (approved):
code changes look good to me 👍🏼

travi:
we're still having some trouble with our test suite completing successfully, especially in node v19. i dont believe the changes in this PR impact that situation and I would want to address that separately anyway, once we can figure out the root cause. 

short of understanding that, i'm going to go ahead and merge this change since we have successful verification from the parts that completed.

PR #2719, opened 2023-02-27
test(integration): fix package name conflict

webstech:
Two tests had the same package name, causing random concurrency failures during CI testing.

An example of the conflict is in [this](https://github.com/semantic-release/semantic-release/actions/runs/4191434175/jobs/7265855238) CI run.

### DIFF
--- a/test/integration.test.js
+++ b/test/integration.test.js
@@ -716,7 +716,7 @@ test("Use the repository URL as is if none of the given git credentials are vali
 });
 
 test("ESM Plugin with named exports", async (t) => {
-  const packageName = "log-secret";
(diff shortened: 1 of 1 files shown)


webstech:
@travi The test did not complete but the [log](https://github.com/semantic-release/semantic-release/actions/runs/4285474077/jobs/7463815230#step:8:172) shows the tests with the two formerly conflicting package names finishing about the same time.  Not a definitive proof but the change is minor.

This patch is for the problem I mentioned [here](https://github.com/semantic-release/semantic-release/pull/2681#issuecomment-1437971323).

travi (approved):
thanks!

PR #2720, opened 2023-02-28
ci: set test workflow timeout to 10 minutes

webstech:
Issues are preventing ci jobs from completing.  This patch reduces the timeout from the default timeout of six hours.  The logs are not always viewable until the timeout so this should help in problem examination.


travi (approved):
worth a shot. thanks!

PR #2721, opened 2023-02-28
test: update docker config for windows support (#2721

webstech:
There is legacy support for PortBindings and Binds at the root level of the config.  Currently (including Windows) these reside under the HostConfig.  ExposedPorts is now used to identify the externally available ports.

Tests run when sideloaded for Linux testing [here](https://github.com/webstech/testteeny/actions/runs/4290352828/jobs/7474281208).  On Windows, it runs locally  but fails on GitHub server with a `docker image not found` error (not related to config change).  Researching the Windows error.

### DIFF
--- a/test/helpers/gitbox.js
+++ b/test/helpers/gitbox.js
@@ -23,7 +23,10 @@ export async function start() {
   container = await docker.createContainer({
     Tty: true,
     Image: IMAGE,
-    PortBindings: {[`${SERVER_PORT}/tcp`]: [{HostPort: `${HOST_PORT}`}]},
+    HostConfig: {
+        PortBindings: {[`${SERVER_PORT}/tcp`]: [{HostPort: `${HOST_PORT}`}]}
+    },
+    ExposedPorts: {[`${SERVER_PORT}/tcp`]: {}}
   });
   await container.start();
 
--- a/test/helpers/mockserver.js
+++ b/test/helpers/mockserver.js
@@ -19,7 +19,10 @@ export async function start() {
   container = await docker.createContainer({
     Tty: true,
     Image: IMAGE,
-    PortBindings: {[`${MOCK_SERVER_PORT}/tcp`]: [{HostPort: `${MOCK_SERVER_PORT}`}]},
+    HostConfig: {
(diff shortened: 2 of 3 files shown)


travi:
i see that this passes, but it isn't clear to me what this change accomplishes. you may have mentioned elsewhere, but could you help me connect the dots?

> On Windows, it runs locally but fails on GitHub server with a `docker image not found` error (not related to config change). Researching the Windows error.

is this related to https://github.com/semantic-release/semantic-release/issues/2658#issuecomment-1382167715?

webstech:
> i see that this passes, but it isn't clear to me what this change accomplishes. you may have mentioned elsewhere, but could you help me connect the dots?

This brings the config up to the current format.  This format allows the CI to run on Linux and Windows (which does not support the legacy format).  It removes the risk of an older format becoming unsupported and allows Windows users to contribute.

> is this related to [#2658 (comment)](https://github.com/semantic-release/semantic-release/issues/2658#issuecomment-1382167715)?

No.  TIL Windows server does not support Linux Docker containers.  Integration testing under Windows can not be part of the CI workflow.  It works fine for devs using Docker on Windows 10/11.

(an aside):
Sorry for the delayed update on this today.  It took some time but I think Windows testing can be done for all the tests except the integration tests using the docker containers.  This [job](https://github.com/webstech/testteeny/actions/runs/4300466766/workflow) is a sample implementation.  That job has a couple of other small patches I will be pushing.  All testing related (like this one).

webstech:
Change details:
`ExposedPorts` is needed for the docker daemon on Windows.  When it is present, `PortBindings` has to be part of `HostConfig`.  If `HostConfig` is present, `Binds` must be part of it and is _ignored_ at the root level.  This last bit originally led to this [change](https://github.com/semantic-release/semantic-release/blob/86cf2645af5e47a38b3bd2ae5374e5c05b3876e6/test/integration.test.js#L69) before I figured out the real problem.

travi:
thanks for the extra context. this really helps. i was confusing independent pieces, so you got me back on the right page. i really appreciate the steps to move this forward and get past legacy approaches 👍🏼

travi (approved):

PR #2722, opened 2023-03-01
test: Windows support in integration tests

webstech:
A pair of patches to the integration tests. 

1. ci: win32 escape slash as doubleslash for test
    Pathing in stderr output has double backslash in the filename.  The test will now use double backslash in the existence test.

2. ci: avoid Windows pop-up in credential tests
    Windows uses a `credential.helper` that will open a pop-up for credentials if they are missing.  A git config parameter will now be set in the environment to avoid the pop-up.

### DIFF
--- a/test/integration.test.js
+++ b/test/integration.test.js
@@ -598,7 +598,7 @@ test("Log unexpected errors from plugins and exit with 1", async (t) => {
   // Verify the type and message are logged
   t.regex(stderr, /Error: a/);
   // Verify the the stacktrace is logged
-  t.regex(stderr, new RegExp(pluginError));
+  t.regex(stderr, new RegExp(process.platform != "win32" ? pluginError : pluginError.replace(/\\/g, "\\\\")));
   // Verify the Error properties are logged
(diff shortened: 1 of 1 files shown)


test/integration.test.js
```diff
@@ -598,7 +598,7 @@ test("Log unexpected errors from plugins and exit with 1", async (t) => {
   // Verify the type and message are logged
   t.regex(stderr, /Error: a/);
   // Verify the the stacktrace is logged
-  t.regex(stderr, new RegExp(pluginError));
+  t.regex(stderr, new RegExp(process.platform != "win32" ? pluginError : pluginError.replace(/\\/g, "\\\\")));
```
travi, line 601: Could we make this a strict comparison?  Also, I think I'd prefer reversing to avoid the negation

webstech: Sorry, ligature fonts may not be as helpful as I had hoped.  It was supposed to be strict.  I will reverse.  I had used negation to keep the non-windows side more visible.

travi (approved):

PR #2723, opened 2023-03-01
ci: index test should reset `testdouble`

webstech:
The testdouble doc recommends resetting after each test.

The first test is now serial to ensure there is no conflict with future tests since it uses mocking.


webstech:
Background:
These two patches do not address any specific problem.  They are the result of researching concurrency issues running the CI.  Other tests using `testdouble` do the reset and run the tests serially.  

I'm not really sure it all works as expected with a high level of concurrency since the different test files are running at the same time, even if the tests in the file are serial.  On the github runners, there are two cpu so the level of concurrency is quite low.  

In my local tests, the biggest concurrency problem is access to the global `.gitconfig`.  This may be  `Git for Windows` related - the config is frequently updated for credential handling (I think).

travi (approved):
thanks!

travi:
thanks for this. we should definitely be reseting between tests, so appreciate closing this gap. i do somewhat consider forcing the tests to be executed serially to be a bit of a test smell, but it is necessary in their current state. i hope to resolve that over time, but this is the right move for now.

PR #2747, opened 2023-03-25
docs(extending): add semantic-release-jira-notes plugin

iamludal:


docs/extending/plugins-list.md
```diff
@@ -179,6 +179,11 @@
 - [semantic-release-coralogix](https://github.com/adobe/semantic-release-coralogix)
   - `verifyConditions` Verified that required credentials are provided and API is accessible
   - `publish` add a release tag to Coralogix
+- [semantic-release-jira-notes](https://github.com/iamludal/semantic-release-jira-notes)
+  - `verifyConditions`: Validate the config options.
+  - `generateNotes`: Generate the release notes with links to JIRA issues.
+- [semantic-release-tags](https://github.com/iamludal/semantic-release-tags)
+  - `success`: Update `major` and `minor` tags.
```
gr2m, line 186: could you add a bit more documentation to that plugin? I'm not quite sure what its use case is?

iamludal: I think that the `semantic-release-major-tag` plugin below should be enough in this list, I will remove my plugin as I can't find any use-case when a user might want to pin to a `vX.Y` tag. I just implemented in case someone might need it.

gr2m (approved):

gr2m (approved):

gr2m:
Hmm the CI errors seem unrelated 🤔

PR #2786, opened 2023-05-06
docs(npx-version): mentioned using a renovate regex manager to update the version

travi:
confirmed in https://github.com/travi-test/semantic-release-tester/commit/0c14764598290d9869fa1a0f6e4d3b1885d23ee5

### DIFF
--- a/docs/usage/installation.md
+++ b/docs/usage/installation.md
@@ -47,5 +47,5 @@ $ npx semantic-release
      ]
    }
    ```
-4. `npx` is a tool bundled with `npm@>=5.2.0`. You can use it to install (and run) the semantic-release binary.
+4. `npx` is a tool bundled with `npm@>=5.2.0`. You can use it to install (and run) the **semantic-release** binary.
    See [What is npx](../support/FAQ.md#what-is-npx) for more details.
--- a/docs/usage/installation.md
+++ b/docs/usage/installation.md
@@ -32,7 +32,7 @@ $ npx semantic-release
    This also means you, or a bot, must upgrade **semantic-release** when a new major version is released.
 2. Pinning **semantic-release** to an exact version makes your releases even more deterministic.
    But pinning also means you, or a bot, must update to newer versions of **semantic-release** more often.
-   You can use [Renovate's regex manager](https://docs.renovatebot.com/modules/manager/regex/) to get updates for **semantic-release**.
+3. You can use [Renovate's regex manager](https://docs.renovatebot.com/modules/manager/regex/) to get automatic updates for **semantic-release** in either of the above scenarios.
    Put this in your Renovate configuration file:
    ```json
    {
@@ -47,5 +47,5 @@ $ npx semantic-release
      ]
    }
    ```
-3. `npx` is a tool bundled with `npm@>=5.2.0`. You can use it to install (and run) the semantic-release binary.
+4. `npx` is a tool bundled with `npm@>=5.2.0`. You can use it to install (and run) the semantic-release binary.
    See [What is npx](../support/FAQ.md#what-is-npx) for more details.
--- a/docs/usage/installation.md
+++ b/docs/usage/installation.md
@@ -24,9 +24,25 @@ For other type of projects we recommend installing **semantic-release** directly
 $ npx semantic-release
 ```
 
-**Note**: For a global installation, it's recommended to specify the major **semantic-release** version to install (for example with `npx semantic-release@18`).
-This way your build will not automatically use the next major **semantic-release** release that could possibly break your build.
-You will have to upgrade manually when a new major version is released.
+### Notes
 
-**Note**: `npx` is a tool bundled with `npm@>=5.2.0`. It is used to conveniently install the semantic-release binary and to execute it.
-See [What is npx](../support/FAQ.md#what-is-npx) for more details.
+1. For a global installation, it's recommended to specify the major **semantic-release** version to install (for example with `npx semantic-release@18`).
+   This way your build will not automatically use the next major **semantic-release** release that could possibly break your build.
+   You will have to upgrade manually when a new major version is released.
+2. Specifying an exact version will make your releases even more deterministic, but requires regular maintenance to stay up to date when new versions of **semantic-release** become available.
+   In projects using [Renovate](https://docs.renovatebot.com/), a [regex manager](https://docs.renovatebot.com/modules/manager/regex/) can be defined to recommend version updates when new **semantic-release** versions are published:
+   ```json
+   {
+     "regexManagers": [
+       {
+         "description": "Update semantic-release version used by npx",
+         "fileMatch": ["^\\.github/workflows/[^/]+\\.ya?ml$"],
+         "matchStrings": ["\\srun: npx semantic-release@(?<currentValue>.*?)\\s"],
+         "datasourceTemplate": "npm",
+         "depNameTemplate": "semantic-release"
+       }
+     ]
+   }
+   ```
+3. `npx` is a tool bundled with `npm@>=5.2.0`. It is used to conveniently install the semantic-release binary and to execute it.
+   See [What is npx](../support/FAQ.md#what-is-npx) for more details.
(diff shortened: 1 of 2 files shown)


docs/usage/installation.md
```diff
@@ -24,9 +24,25 @@ For other type of projects we recommend installing **semantic-release** directly
...
-
-**Note**: `npx` is a tool bundled with `npm@>=5.2.0`. It is used to conveniently install the semantic-release binary and to execute it.
-See [What is npx](../support/FAQ.md#what-is-npx) for more details.
+### Notes
+
+1. For a global installation, it's recommended to specify the major **semantic-release** version to install (for example with `npx semantic-release@18`).
+   This way your build will not automatically use the next major **semantic-release** release that could possibly break your build.
+   You will have to upgrade manually when a new major version is released.
+2. Specifying an exact version will make your releases even more deterministic, but requires regular maintenance to stay up to date when new versions of **semantic-release** become available.
+   In projects using [Renovate](https://docs.renovatebot.com/), a [regex manager](https://docs.renovatebot.com/modules/manager/regex/) can be defined to recommend version updates when new **semantic-release** versions are published:
```
HonkingGoose, line 33:
```suggestion
   You can use [Renovate's regex manager](https://docs.renovatebot.com/modules/manager/regex/) to get updates for **semantic-release**.
   Put this in your Renovate configuration file:
```

docs/usage/installation.md
```diff
@@ -24,9 +24,25 @@ For other type of projects we recommend installing **semantic-release** directly
...
+         "description": "Update semantic-release version used by npx",
+         "fileMatch": ["^\\.github/workflows/[^/]+\\.ya?ml$"],
+         "matchStrings": ["\\srun: npx semantic-release@(?<currentValue>.*?)\\s"],
+         "datasourceTemplate": "npm",
+         "depNameTemplate": "semantic-release"
+       }
+     ]
+   }
+   ```
+3. `npx` is a tool bundled with `npm@>=5.2.0`. It is used to conveniently install the semantic-release binary and to execute it.
```
HonkingGoose, line 47:
```suggestion
3. `npx` is a tool bundled with `npm@>=5.2.0`. You can use it to install (and run) the semantic-release binary.
```

HonkingGoose (commented):
My goals with this rewrite are to:

- Clarify _who_ is recommending things here
- Explain that pinning to a specific version of **semantic-release** means users (or a bot) needs to check for updates often
- Use simple words and phrases
- Use `you` to put the reader into the text
- Use `must` to indicate requirements

docs/usage/installation.md
```diff
@@ -24,9 +24,25 @@ For other type of projects we recommend installing **semantic-release** directly
...
-This way your build will not automatically use the next major **semantic-release** release that could possibly break your build.
-You will have to upgrade manually when a new major version is released.
-
-**Note**: `npx` is a tool bundled with `npm@>=5.2.0`. It is used to conveniently install the semantic-release binary and to execute it.
-See [What is npx](../support/FAQ.md#what-is-npx) for more details.
+### Notes
+
+1. For a global installation, it's recommended to specify the major **semantic-release** version to install (for example with `npx semantic-release@18`).
+   This way your build will not automatically use the next major **semantic-release** release that could possibly break your build.
+   You will have to upgrade manually when a new major version is released.
```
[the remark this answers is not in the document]
travi: i have not tested the regex manager config suggested in this PR against an npx command that specifies only the major version. it is unclear if it would update only the major version, replace the partial definition with a full one, or fail in some way. is there a difference in this case that is worth clarifying in some way, or should both scenarios be expected to work with the suggested config?

travi:
thanks for taking the time to provide feedback, @HonkingGoose! much of this change was just moving existing content into a more appropriate structure, but your suggestions do improve the content, so much appreciated.

docs/usage/installation.md
```diff
@@ -24,9 +24,28 @@ For other type of projects we recommend installing **semantic-release** directly
...
+   But pinning also means you, or a bot, must update to newer versions of **semantic-release** more often.
+   You can use [Renovate's regex manager](https://docs.renovatebot.com/modules/manager/regex/) to get updates for **semantic-release**.
+   Put this in your Renovate configuration file:
+   ```json
+   {
+     "regexManagers": [
+       {
+         "description": "Update semantic-release version used by npx",
+         "fileMatch": ["^\\.github/workflows/[^/]+\\.ya?ml$"],
+         "matchStrings": ["\\srun: npx semantic-release@(?<currentValue>.*?)\\s"],
```
HonkingGoose, line 43: Is this matchStrings valid for both _pinned_ and _normal_ versions? Or just for one of those?

travi: good question. i wondered the similar. i tested with:
* [just the major version](https://github.com/travi-test/semantic-release-tester/pull/418/files)
* [a fully pinned version](https://github.com/travi-test/semantic-release-tester/pull/417/files)

both were recognized, since the expression ends with the whitespace after the version, and provided updates as i would have expected. i think this covers the scenarios suggested in this portion of the docs.

out of curiosity, i've also tried [specifying the version as simply `beta`](https://github.com/travi-test/semantic-release-tester/commit/d359f6051719a644270cae7542be59532702f06f) since i wasnt sure how that would be handled with no additional configuration. i did not receive an update PR, which is good since i normally use that when testing a pre-release channel while working through a series of updates. i'm still curious what would happen if i specified a precise pre-release version, like `semantic-release@22.0.0-beta.1`, but have not tried that out yet. i think that curiosity goes beyond the goals of this PR, though, so i can experiment further later.

HonkingGoose: > both were recognized, since the expression ends with the whitespace after the version, and provided updates as i would have expected. i think this covers the scenarios suggested in this portion of the docs.

Can you explain in the docs what works/doesn't work? That way readers don't have to guess.

> i'm still curious what would happen if i specified a precise pre-release version, like `semantic-release@22.0.0-beta.1`, but have not tried that out yet. i think that curiosity goes beyond the goals of this PR, though, so i can experiment further later.

I don't know if the `matchStrings` "version regex" is "wide" enough to include `-beta` versions. I don't know enough RegEx to check this. I think you should try it, and see what happens. 😄 

Also if you made it "wide" enough, so it captures `-beta` you should then find a way to avoid the `--` CLI flags.

gr2m (approved):
nice!

docs/usage/installation.md
travi: > Can you explain in the docs what works/doesn't work? That way readers don't have to guess.

i decided to split the mention of the regex manager to a separate note so that i could clarify that it was an automation option for either pinning strategy. i think that provides better clarity for those scenarios

> I don't know if the matchStrings "version regex" is "wide" enough to include -beta versions. I don't know enough RegEx to check this. I think you should try it, and see what happens. 😄

the `matchStrings` regex does match versions that include the pre-release version. i think those scenarios are beyond the scope of what these particular docs need to cover, so i'm comfortable investigating those details after getting these updates merged.

> Also if you made it "wide" enough, so it captures -beta you should then find a way to avoid the -- CLI flags.

since the regex only matches up to the next whitespace character, other trailing flags would already be excluded

HonkingGoose: Thanks for responding to me again `@travi`, and for improving the docs again.

I'll let you resolve this comment. 😉 I don't have the rights to do that, for whatever reason.

docs/usage/installation.md
```diff
@@ -24,9 +24,28 @@ For other type of projects we recommend installing **semantic-release** directly
...
+         "datasourceTemplate": "npm",
+         "depNameTemplate": "semantic-release"
+       }
+     ]
+   }
+   ```
+4. `npx` is a tool bundled with `npm@>=5.2.0`. You can use it to install (and run) the semantic-release binary.
```
HonkingGoose, line 50: I think **semantic-release** needs to be bold:

```suggestion
4. `npx` is a tool bundled with `npm@>=5.2.0`. You can use it to install (and run) the **semantic-release** binary.
```

HonkingGoose (approved):

docs/usage/installation.md
travi: good catch

PR #2794, opened 2023-05-11
docs: add triggering semantic-release on private repositories

JIMBambao:


travi:
could you clarify the use case that you are trying to capture with this contribution? what do you mean by private in this case?  

setting `"private": true` in the `package.json` prevents publishing a package altogether. in a way, that property is named poorly because it truly means something closer to "this project is not a package". if you are targeting projects that are not packages with this contribution, i dont think that is clear with this iteration.

PR #2796, opened 2023-05-12
ci(release): pinned the version of semantic-release

travi:
the goal with this is to make releases more deterministic based on commit rather than a point in time. in combination with https://github.com/semantic-release/.github/pull/18, we should continue to release with the latest version, but without that dependence on time.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,7 +25,7 @@ jobs:
           node-version: lts/*
       - run: npm clean-install
       - run: npm audit signatures
-      - run: npx semantic-release
+      - run: npx semantic-release@21.0.2
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -25,7 +25,7 @@ jobs:
           node-version: lts/*
       - run: npm clean-install
       - run: npm audit signatures
-      - run: npx semantic-release
+      - run: npx semantic-release@21.0.2
```
gr2m, line 28: Should we add comment explaining that it's updated automatically by renovate and how? People might look at our release workflow for reference

gr2m (approved):

.github/workflows/release.yml
travi: good idea. i'll get something added and link to our docs about it

PR #2814, opened 2023-05-27
refactor: use makeTag from utils rather than lodash template

TheWolfNL:


lib/utils.js
```diff
@@ -1,4 +1,4 @@
-import { isFunction, template, union } from "lodash-es";
+import { template, union } from "lodash-es";
```
TheWolfNL, line 1: `isFunction` was imported but not used.

lib/branches/get-tags.js
```diff
@@ -1,8 +1,9 @@
-import { escapeRegExp, template } from "lodash-es";
+import { escapeRegExp } from "lodash-es";
 import semver from "semver";
 import pReduce from "p-reduce";
 import debugTags from "debug";
 import { getNote, getTags } from "../../lib/git.js";
+import { makeTag } from "../../lib/utils.js";
```
TheWolfNL, line 6: I copied the line above, but this could also be:
```suggestion
import { getNote, getTags } from "../git.js";
import { makeTag } from "../utils.js";
```
though in that manner it would lose the mention of libs and might not be as explicit / obvious.

lib/utils.js
TheWolfNL: Incase the unused function is desired
```suggestion
import { isFunction, template, union } from "lodash-es";
```

PR #2842, opened 2023-06-20
test(integration): properly waited for the image pulls to complete

travi:
this uses the technique for waiting for the pull to complete [from the dockerode readme](https://github.com/apocas/dockerode/tree/edf29ccb2c2c7bfcdd1cf3cacbe861bd0f4bc87a#helper-functions) and mostly gets rid of our dependence on get-stream.

the [remaining get-stream usage](https://github.com/semantic-release/semantic-release/blob/45eb11ae9cb46e6167ecfa255ae5c62ab0d016ff/lib/git.js#L49) is the `.array()` that is [removed in v7](https://github.com/sindresorhus/get-stream/releases/tag/v7.0.0). i suggest we deal with that usage separately from this PR

### DIFF
--- a/test/helpers/gitbox.js
+++ b/test/helpers/gitbox.js
@@ -1,5 +1,4 @@
 import Docker from "dockerode";
-import getStream from "get-stream";
 import pRetry from "p-retry";
 import { gitShallowClone, initBareRepo } from "./git-utils.js";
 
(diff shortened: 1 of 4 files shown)


travi:
it appears that the pulls are happening before each test in the integration file. would be better to do the pulls before the overall suite run so that they only happen once

travi:
> it appears that the pulls are happening before each test in the integration file. would be better to do the pulls before the overall suite run so that they only happen once

i was wrong. it is using `test.before`, not `test.beforeEach`, so this should already be correct. since we are actually fully waiting for the pulls, we may need to extend the timeout for the integration test, though

travi:
it looks like this change does not fix the flakiness, but i think it is still a worthwhile change to get merged

gr2m (approved):

PR #2848, opened 2023-06-23
ci(ossf-scorecard): added workflow calculating scorecard metrics

travi:
the changes to detect semantic-release in github actions workflows was [released yesterday](https://github.com/ossf/scorecard/releases/tag/v4.11.0), so it feels like the right time to add this to our workflows. 

our [current score is 7.7](https://securityscorecards.dev/viewer/?uri=github.com/semantic-release/semantic-release), but i expect that to go up once this is merged and the score is recalculated with the new detection for semantic-release. i'm hopeful that the failure to detect our license might also be fixed, but i haven't dug in to confirm that one 🤞🏼

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,9 @@
   <a href="https://github.com/semantic-release/semantic-release/actions?query=workflow%3ATest+branch%3Amaster">
     <img alt="Build states" src="https://github.com/semantic-release/semantic-release/workflows/Test/badge.svg">
   </a>
+  <a href="https://securityscorecards.dev/viewer/?uri=github.com/semantic-release/semantic-release">
+    <img alt="OpenSSF Scorecard" src="https://api.securityscorecards.dev/projects/github.com/semantic-release/semantic-release/badge">
+  </a>
   <a href="#badge">
     <img alt="semantic-release: angular" src="https://img.shields.io/badge/semantic--release-angular-e10079?logo=semantic-release">
   </a>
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -7,6 +7,9 @@
   <a href="https://github.com/semantic-release/semantic-release/actions?query=workflow%3ATest+branch%3Amaster">
     <img alt="Build states" src="https://github.com/semantic-release/semantic-release/workflows/Test/badge.svg">
   </a>
+  <a href="https://securityscorecards.dev/viewer/?uri=github.com/semantic-release/semantic-release">
```
travi, line 10: we've kept badges to a minimum, but this one feels worth including. open to leaving it out if it feels inconsistent

travi:
an example of the new semantic-release detection working and rewarding a 10 for the packaging score: https://securityscorecards.dev/viewer/?uri=github.com/travi-test/semantic-release-tester#section-Packaging

travi:
> i'm hopeful that the failure to detect our license might also be fixed

according to https://github.com/ossf/scorecard/blob/4cd5446862ea4c470810fea81fc7f45a36d04dec/docs/checks.md#license, the scorecard relies on GitHub's understanding of our license. it is true that GitHub does not recognize the license, but i do not understand why. comparing the contents of https://github.com/semantic-release/semantic-release/blob/master/LICENSE to https://github.com/form8ion/javascript/blob/master/LICENSE from another of my projects where the license is recognized correctly, i dont spot obvious differences beyond the ones that would be expected to differ between projects.

gr2m:
> i dont spot obvious differences beyond the ones that would be expected to differ between projects.

maybe we can try to just update the LICENSE file? That might resolve it

gr2m (approved):

PR #2856, opened 2023-06-28
ci: added a codeql workflow

travi:
which will help us understand if our code has any security issues that can be found through static analysis and will also satisfy the SAST check for the ossf scorecard


travi:
closing in favor of the default setup enabled through the repo settings

Issue #2313, opened 2021-12-31
Error while publishing to npm "EINVALIDNPMTOKEN Invalid npm token."


senshiii:
I am trying to publish a package using semantic release and this is the error I am receiving.

Versions :- 
1. semantic-release : 18.0.1
2. node : 16.13.1
3. npm : 8.3.0

I have first run the `npx semantic-release-cli setup` with a valid GitHub token. This GitHub Token is stored as `MY_GITHUB_TOKEN` in my repo secrets. This command created a "publish" token and created a `NPM_TOKEN` secret in my [GitHub Repo](https://github.com/senshiii/immitate). 

I am using GitHub Actions. This is my ci yml file.

```yml
  name: CI
on:
  push:
    branches: [master]
jobs:
  build:
    name: Build

    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        node: [12.x, 14.x, 16.x]
        os: [windows-latest, ubuntu-latest]
        include:
          - os: ubuntu-latest
            node: 16.x
            upload: true

    steps:
      - name: Checkout repo
        uses: actions/checkout@v2

      - name: Use Node ${{ matrix.node }}
        uses: actions/setup-node@v2
        with:
          node-version: ${{ matrix.node }}

      - name: Install dependencies ( clean )
        run: npm ci --ignore-scripts

      - name: Build
        run: npm run build --if-present

      - name: Upload Artifacts
        if: matrix.upload
        uses: actions/upload-artifact@v2
        with:
          name: dist
          path: dist

  publish:
    name: Publish to npm
    runs-on: ubuntu-latest
    needs: [build]

    steps:
      - name: Checkout
        uses: actions/checkout@v2

      - name: Download build artifact
        uses: actions/download-artifact@v2
        with:
          name: dist
          path: dist

      - name: Setup Node
        uses: actions/setup-node@v2
        with:
          node-version: 16.x
          registry-url: https://registry.npmjs.org/

      - name: Install
        run: npm ci

      - name: Semantic Release
        run: npm run semantic-release
        env:
          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
          GITHUB_TOKEN: ${{ secrets.MY_GITHUB_TOKEN }}
```   

Logs of semantic release step

```bash
[4:11:02 AM] [semantic-release] › ℹ  Running semantic-release version 18.0.1
[...]

PR #2875, opened 2023-07-15
docs: update example to use NODE_AUTH_TOKEN env

rensjaspers:
The example shows the wrong env key in the release job. It says `NPM_TOKEN`, but it should be `NODE_AUTH_TOKEN`. 

See: https://github.com/semantic-release/semantic-release/issues/2313

### DIFF
--- a/docs/recipes/ci-configurations/github-actions.md
+++ b/docs/recipes/ci-configurations/github-actions.md
@@ -57,7 +57,7 @@ jobs:
       - name: Release
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
         run: npx semantic-release
 ```
 


travi:
appreciate the contribution, but this should not be necessary. I have lots of projects configured to use `NPM_TOKEN` and all of them are working fine. i would rather investigate the scenarios where `NPM_TOKEN` is not working as expected to understand what the trouble is. would you be willing to help investigate that?

rensjaspers:
I might be mistaken, but it seems that you're writing the literal string `${NPM_TOKEN}` to the `.npmrc` here instead of its actual value: https://github.com/semantic-release/npm/blob/master/lib/set-npmrc-auth.js#L31

Could this be the source of the issue?

travi:
that is written to the generated `.npmrc` so that the value is read from the environment at publish time. see https://docs.npmjs.com/cli/v9/configuring-npm/npmrc#files for details of how that works from an `.npmrc` file

rensjaspers:
I see. Looks like I need to spend some more time to figure out what's going on. I'll gladly help investigate.

tsi:
I also had a similar experience, and replacing `NPM_TOKEN` with `NODE_AUTH_TOKEN` helped, what should be investigated on our side?
Seems like [many others had the same issue](https://github.com/semantic-release/semantic-release/issues/2313), maybe that issue should not be marked as closed?

tsi:
@travi thanks, seems like it does.

travi:
Thanks for confirming. I'm going to close this PR then. Would accept a PR that highlights these details to help others avoid this situation.

PR #2916, opened 2023-08-21
feat: Added TypeScript definitions

seebeen:
These type definitions are feature complete and include all of the interfaces, options and callbacks from the main semantic-release package.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,711 @@
+declare interface AggregateError extends Error {
+  errors: any[];
+}
+
+declare module "semantic-release" {
+  import { Signale } from "signale";
+  export interface EnvCi {
+    /**
+     * Boolean, true if the environment is a CI environment
+     */
+    isCi: boolean;
+
+    /**
+     * Commit hash
+     */
+    commit: string;
+
+    /**
+     * Current branch name
+     */
+    branch: string;
+  }
+
+  /**
+   * Base context used in every semantic release step.
+   */
+  export interface BaseContext {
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -0,0 +1,711 @@
+declare interface AggregateError extends Error {
+  errors: any[];
+}
```
gr2m, line 3: do we still need to define it? It's a global object now:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError

gr2m (commented):
should we add type tests with `tsd`, at least some basic ones that we can add to as needed?

index.d.ts
travi: we're still [using the library](https://github.com/semantic-release/semantic-release/blob/915f36cda7ed9bd2487d6dd5d7b32ca9b32d3dd3/package.json#L34) rather than the built-in. i think there is something that would make the switch a breaking change, but i do think we should migrate

seebeen:
> should we add type tests with `tsd`, at least some basic ones that we can add to as needed?

I can add those later in. The typedef file added in this PR is also in @semantic-release/wordpress, and in a PR [https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66223](here).

Since that PR won't be merged anytime soon, let's roll our own, and I can implement a full typechecking test later on.

gr2m (approved):

travi (approved):

PR #2934, opened 2023-09-01
feat(conventional-changelog-presets): supported new preset format

### LINKED ISSUE #2929: Tracking issue for conventional-changelog preset updates
### My release workflow is broken with an error about conventional-changelog-*

new versions of conventional-changelog presets were released over the weekend with breaking changes. these updates were properly released as major versions, but you may be installing the latest version as part of your release pipeline.

Note: if you are using the default angular preset through the core package without installing independently, you should not be experiencing problems because the previous major is still defined as the installed dependency in that situation

### What can i do to fix my release pipeline?

you will need to specify a version or version range for the preset package that you are installing. for example, if you are installing with npm before executing, you could update like:

```diff
- npm install -g semantic-release conventional-changelog-conventionalcommits
+ npm install -g semantic-release conventional-changelog-conventionalcommits@6
```

### What will the semantic-release team do to handle this issue going forward?

this is a breaking change for us too. this is a change to how presets are loaded and we will be updating to handle the change. however, we want to make sure we handle the update properly, so we are not making the change immediately. we will likely treat this change as a breaking change for our consumers by dropping support for loading the presets that used the previous loading approach. 

### Other issues exposed by this situation

we are seeing feedback about some other issues that need attention that arent directly related to this issue, but are exposed/highlighted by this situation. we are paying attention to that feedback, but will need to follow up on those issues after the primary complications are addressed. 

* #2932 

### How can i help?

if you are looking to resolve problems with your releases, focus on specifying the version of the preset you are loading. however, if you would like to help us investigate these issues, we welcome you help and participation in the conversation.

travi:
BREAKING CHANGE: the new preset format is a breaking change when compared to the previous preset format. updating to support the new format means that the old preset format is no longer supported. update your preset to the latest version to maintain compatibility

closes #2929


travi:
related work in the related plugins:
* https://github.com/semantic-release/commit-analyzer/pull/519
* https://github.com/semantic-release/release-notes-generator/pull/526

loginglive:
Daniel ramos navarro
Dependency: RAMSALM

travi:
we need an integration test to verify the behavior of handling reverted commits where the original commit was created within this release (it should be excluded). the order of handling commits has changed in conventional-changelog with the releases that this PR is addressing, so we want to make sure we are accounting for that change appropriately

loginglive:
https://www.ramsalm.com

felipecrs:
Maybe it will be a good idea to gate https://github.com/semantic-release/semantic-release/pull/1836 against the same major release?

felipecrs:
I see, it makes sense. Thanks for taking the time to consider it.

travi:
> I see, it makes sense. Thanks for taking the time to consider it.

this one is still high on my list, too. once we get this one out, it is a good excuse to give it a bit more attention. thanks for keeping us honest about it :)

travi:
> we need an integration test to verify the behavior of handling reverted commits where the original commit was created within this release (it should be excluded). the order of handling commits has changed in conventional-changelog with the releases that this PR is addressing, so we want to make sure we are accounting for that change appropriately

i ran some manual tests in https://github.com/travi-test/semantic-release-tester before the beta, with beta.2, and with beta.3. beta.2 did not correctly exclude the reverted commits, but beta.3 did. beta.3 includes the conventional-changelog packages that were previously not yet included in the commit-analyzer and release-notes-generator betas. this gives me more confidence that the commit order changes are handled properly. i still want to get a test in place, but might consider that addition after getting the initial version promoted to stable.

travi:
i'm wondering if we should raise the required node version to 18.17.0 to accommodate the new requirement from npm v20. see https://github.com/semantic-release/npm/pull/671

felipecrs:
In case you are looking for some user feedback

travi:
> > we need an integration test to verify the behavior of handling reverted commits where the original commit was created within this release (it should be excluded). the order of handling commits has changed in conventional-changelog with the releases that this PR is addressing, so we want to make sure we are accounting for that change appropriately
> 
> i ran some manual tests in [travi-test/semantic-release-tester](https://github.com/travi-test/semantic-release-tester) before the beta, with beta.2, and with beta.3. beta.2 did not correctly exclude the reverted commits, but beta.3 did. beta.3 includes the conventional-changelog packages that were previously not yet included in the commit-analyzer and release-notes-generator betas. this gives me more confidence that the commit order changes are handled properly. i still want to get a test in place, but might consider that addition after getting the initial version promoted to stable.

i captured https://github.com/semantic-release/semantic-release/issues/2959 for this since i dont want the addition of that test and refactoring the integration test file to further delay getting this release out.

gr2m:
> i'm wondering if we should raise the required node version to 18.17.0 to accommodate the new requirement from npm v20. see [semantic-release/npm#671](https://github.com/semantic-release/npm/pull/671)

works for me 👍🏼

travi:
looks like loaders changed after node v20, so that results in trouble between ava and testdouble. looks like we'll no longer be able to specify the loader as part of ava's nodeArguments, but [need to use NODE_OPTIONS instead](https://github.com/avajs/ava/discussions/3213#discussioncomment-6268035). that adjustment gets things working for most of the tests, but causes testdouble (maybe the loader from testdouble?) to not be found within the integration test.

using node 20.6 and quibble v0.8.0 without specifying the loader explicitly seems to get past most of the integration test issues, but that hasnt yet been integrated into a testdouble release yet. that is in progress here: https://github.com/testdouble/testdouble.js/pull/516. 

i considered disabling the node v20 tests to get this merged since testing against node v20 is new to this pipeline, but the testdouble loader changes make me want to investigate a bit more and maybe wait for the next testdouble release, which i expect will be soon. the main reason i'd like to wait in that case is that using the automatic loaders requires node v20.6 and i'd want to raise our supported v20 range to have that minimum instead of the current proposal in this PR.

worst case, if i dont get the tests working with node v20.6 soon, i may still disable that run from the pipeline and raise the minimum anyway, but would love to see it work before setting the minimum requirement

travi:
this ended up being an interesting rabbit hole.

> maybe the loader from testdouble?

this ended up being the case for this issue. since we werent using `NODE_OPTIONS` to supply node options during the ava test runs before, (instead using the `nodeArguments` ava config) we were [not ignoring `NODE_OPTIONS` from the environment](https://github.com/semantic-release/semantic-release/blob/42f7b8269ee1c2594785a54d478975a8ae90fe44/test/integration.test.js#L33) passed into the `semantic-release` executions in the integration tests. with that now specifying to use the `testdouble` loader, the executions attempted to wire up the loader that was already wired up and reported `testdouble` as unable to be found.


> that hasnt yet been integrated into a testdouble release yet

[v3.19.0](https://github.com/testdouble/testdouble.js/commit/baa2f9acee77ac917f3595b69401ec14baadeb6b) of testdouble was released today and i got that working with node v20.6.0 without specifying `NODE_OPTIONS`. 

that left the node v18 test execution in the state mentioned above. i eventually realized that i could switch back from `NODE_OPTIONS` to the `nodeArguments` ava config in that node version and that it would be ignored in node v20. that is actually a pretty nice combination since it allows us to avoid conditionally passing the loader option depending on which node version the tests are running in.

.github/workflows/test.yml
```diff
@@ -20,8 +20,9 @@ jobs:
     strategy:
       matrix:
         node-version:
-          - 18.0.0
-          - 19
+          - 18.17.0
+          - 20.6.0
```
travi, line 24: since we now have two minimum versions with the or-ed supported ranges, this prevents accidentally breaking those lower versions

.github/workflows/test.yml
```diff
@@ -20,8 +20,9 @@ jobs:
...
+          - 20
```
travi, line 25: this enables testing the upper bound of the v20 major. i did not include similar for the v18 range since the risk of breaking that upper bound is lower since we are also testing in v20

package.json
```diff
@@ -8,6 +8,7 @@
     "files": [
       "test/**/*.test.js"
     ],
+    "failFast": true,
```
travi, line 11: i think this will help our situation where we have tests failing/timing out but not exiting in our pipeline. it does make the issue of not cleaning up after the integration test a bit worse, but that appears to be an issue with [how `test.after.always` works in ava](https://github.com/avajs/ava/issues/928) and already existed in some other situations. that should only be a problem with local test executions.

test/integration.test.js
```diff
@@ -35,7 +35,7 @@ let env;
 
 // Environment variables used only for the local npm command used to do verification
 const npmTestEnv = {
-  ...process.env,
+  ...processEnvWithoutGitHubActionsVariables,
```
travi, line 38: not directly impactful now that i backed away from specifying the `testdouble` loader with `NODE_OPTIONS`, but realized that we were not doing this here while debugging that situation

package.json
```diff
@@ -8,6 +8,7 @@
...
     "nodeArguments": [
       "--loader=testdouble",
```
travi, line 13: once we drop node v18 support, we'll want to remember to remove this from this list

gr2m (approved):
Looks good to me!

Just for reference: there is known problem with v20.6.0 and circulate dependencies in CJS files, not sure if that could affect us, maybe plugins? See https://github.com/actions/toolkit/issues/1523 I think we should be fine keeping the tests in 20.6.0 as our test suite and code is all ESM now

travi:
> there is known problem with v20.6.0 and circulate dependencies in CJS files

good info! i'll just bump the minimum for the v20 range to v20.6.1. v20 isnt even LTS yet, so i'm not too concerned about being restrictive on the supported minimum at this point

Issue #2191, opened 2021-10-14
Opt-out from the Terminal markdown rendered in dry-run mode


n0th1ng-else:
## New feature motivation

Hello. I wonder if we can opt-out from the Terminal markdown rendered when we run the semantic-release in --dry-run mode?

We use dry-run mode to render the upcoming release notes once developer opens a pull request. It actually does two things - first, it shows the expected version once the PR is merged into the main branch and secondly, it breaks the pipeline if the commit message was in the wrong format and semantic-release could not fetch the proper commit data. So it is kind of secures us from making mistakes while we specify the commit message.

The --dry-run mode transforms the release notes from the markdown into terminal-like text which is nice, looks a bit off in our case. When I put the result of the dry-run into the GitHub comment it looks not that nice. 

So I wonder if we can introduce the flag that allows to output of the markdown release notes instead of getting it formatted with [TerminalRendered](https://github.com/semantic-release/semantic-release/blob/master/index.js#L24)? I understand using the markdown terminal formatted is for human readability, but I think we can find more examples of when the release notes can be useful in the raw markdown format

<img width="863" alt="Screenshot 2021-10-14 at 14 01 53" src="https://user-images.githubusercontent.com/46719800/137313552-e6091639-ee69-4658-83a3-e7a1e3baddc6.png">
<img width="866" alt="Screenshot 2021-10-14 at 14 02 18" src="https://user-images.githubusercontent.com/46719800/137313557-1dbad6ac-bd85-413d-a00d-e5063539c3bc.png">

## New feature description

Basically, I am thinking of an extra CLI flag that can be used in pair with `--dry-run` flag. Something, like `npx semantic-release --dry-run --markdown-output` would be enough I guess

## New feature implementation

The idea is to introduce an additional flag (which is not a breaking change) and once we are in dry run mode, output the release notes with no changes [here](https://github.com/semantic-release/semantic-release/blob/master/index.js#L210)

PR #2937, opened 2023-09-04
feat: make printing Markdown to the Terminal optional

diogokiss:
This commit adds a new configuration option called 'marked'. When set to true, the output of Semantic Release would print Markdown to the terminal, whenever supported, using the 'marked-terminal' library as usual.

https://www.npmjs.com/package/marked-terminal

When set to false, the output will be printed in plain text. This feature is particularly useful when the output or the command will be consumed by another Markdown renderer, like GitHub.

This addresses the following issue.
https://github.com/semantic-release/semantic-release/issues/2191


ducanhpham0312 (approved):

jkuball:
Why was this never merged? It is still an issue. I don't see any way on getting the correctly rendered release notes in a dry run, and I'd really love to. What can we/I do to get this merged? :)

Code looks fine for me, it is backwards compatible, and it solves a real problem.

diogokiss:
@ducanhpham0312 or @gr2m , do would think this Pull Request might get a review any time soon? 🥺

travi:
we can understand the desire for this, but i dont think we are prepared to handle this with the current form of semantic-release. first, i dont think this is the best approach to capturing the output for display in github. unfortunately, we dont currently support that feature, but working around the current limitation like this leaves us with more complexity that we would need to support long into the future. i would much rather have first class-support for that functionality.

we do have plans to make progress in this direction, but can't promise a timeline. the best way to follow our plan is to track https://github.com/semantic-release/semantic-release/issues/4072

PR #2963, opened 2023-09-17
ci(matrix): prevent skipping the test job when test_matrix fails

travi:
for #semantic-release/.github#19

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,6 +15,10 @@ on:
 permissions:
   contents: read # to fetch code (actions/checkout)
 
+env:
+  FORCE_COLOR: 1
+  NPM_CONFIG_COLOR: always
+
 jobs:
   test_matrix:
     strategy:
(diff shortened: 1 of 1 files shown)


travi:
this prevents auto merge from happening when one of the matrix jobs fails, resulting in the `test` job being skipped, which is treated as passing.

as mentioned in https://github.com/semantic-release/.github/issues/19, i hope to restructure the pipeline a bit as a result of this change, but want to get the basic change in place first. the restructuring i have in mind would benefit from a bit more conversation

.github/workflows/test.yml
```diff
@@ -15,6 +15,11 @@ on:
 permissions:
   contents: read # to fetch code (actions/checkout)
 
+
+env:
+  FORCE_COLOR: 1
+  NPM_CONFIG_COLOR: always
```
travi, line 21: this has nothing to do with the required check, but will enable more colorful output in the workflow logs

gr2m (approved):

Issue #2973, opened 2023-09-21
SyntaxError: Named export 'execa' not found; "semantic-release" exited with code 1 (SIGHUP)


tomerh2001:
### Current behavior

```
<1s
+ bunx semantic-release
file:///opt/atlassian/pipelines/agent/build/node_modules/semantic-release/bin/semantic-release.js:6
import { execa } from "execa";
         ^^^^^
SyntaxError: Named export 'execa' not found. The requested module 'execa' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:
import pkg from 'execa';
const { execa } = pkg;
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:131:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:213:5)
    at async ModuleLoader.import (node:internal/modules/esm/loader:308:24)
    at async loadESM (node:internal/process/esm_loader:42:7)
    at async handleMainPromise (node:internal/modules/run_main:66:12)
Node.js v20.6.1
error: "semantic-release" exited with code 1 (SIGHUP)
```

### Expected behavior

For semantic release to start running

### `semantic-release` version

"semantic-release": "^22.0.0",

### CI environment

Bitbucket Pipelines

### Plugins used

  - '@semantic-release/commit-analyzer'
  - '@semantic-release/release-notes-generator'
  - '@semantic-release/npm'
  - '@semantic-release/changelog'
  - '@semantic-release/git'

### `semantic-release` configuration

branches:
  - develop
  - main
plugins:
  - '@semantic-release/commit-analyzer'
  - '@semantic-release/release-notes-generator'
  - '@semantic-release/npm'
  - '@semantic-release/changelog'
  - '@semantic-release/git'

### CI logs

```
Build




Build setup
20s
apt-get update
1s
apt-get install -y git
5s
npm i -g bun
4s
bun install
<1s
bunx semantic-release
<1s
+ bunx semantic-release
file:///opt/atlassian/pipelines/agent/build/node_modules/semantic-release/bin/semantic-release.js:6
import { execa } from "execa";
         ^^^^^
SyntaxError: Named export 'execa' not found. The requested module 'execa' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:
import pkg from 'execa';
const { execa } = pkg;
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:131:21)
[...]

PR #2982, opened 2023-09-23
fix: make engine version exception for Bun

jamonholmgren:
I've tested this and it works fine, and semantic-release can run (as far as I can tell) with no issues in a Bun-only environment (no Node). [Example -- running on CircleCI](https://app.circleci.com/pipelines/github/jamonholmgren/bluebun/34/workflows/742b2662-d60f-4ff2-8f97-ab9b93d9b70c/jobs/39).

However, Bun reports its Node version as 18.15. So, when you try to run semantic-release without this change, you get this error:

```
 ~/Code/ bunx --bun semantic-release --version
[semantic-release]: node version ^18.17 || >=20.6.1 is required. Found v18.15.0.

See https://github.com/semantic-release/semantic-release/blob/master/docs/support/node-version.md for more details and solutions.
```

Given that semantic-release works fine with Bun, we should not exit if we are running in the Bun runtime. This tiny PR allows semantic-release to do that.

### DIFF
--- a/bin/semantic-release.js
+++ b/bin/semantic-release.js
@@ -14,7 +14,7 @@ const { satisfies, lt } = semver;
 
 const MIN_GIT_VERSION = "2.7.1";
 
-if (!satisfies(process.version, engines.node)) {
+if (!satisfies(process.version, engines.node) && typeof Bun === "undefined") {
   console.error(
     `[semantic-release]: node version ${engines.node} is required. Found ${process.version}.
 


gr2m (approved):
I'm okay given how minimal the change is. I am worried of additional maintenance overhead from Bun users though, to be honest.

If we officially support Bun, we should also run CI tests with Bun, but we can do this in a follow up PR. Let's see what @travi thinks

bin/semantic-release.js
```diff
@@ -14,7 +14,7 @@ const { satisfies, lt } = semver;
 
 const MIN_GIT_VERSION = "2.7.1";
 
-if (!satisfies(process.version, engines.node)) {
+if (!satisfies(process.version, engines.node) && typeof Bun === "undefined") {
```
gr2m, line 17: I wonder if we should reverse the check, instead of checking if the runtime environment is Bun, we could check if the runtime environment is node

travi:
I've been really debating with myself about how to respond to this request. I'll start by saying that I haven't yet spent any time to understand bun, so I could use some education around some details where my assumptions are likely wrong.

> I am worried of additional maintenance overhead from Bun users though, to be honest.

This is my biggest concern as well. The primary reason we hard fail with that check is specifically to minimize the potential runtime environments that can result in support requests. We know there are node versions outside of our supported ranges that can work, but we choose to prevent them from resulting in more variation for user issues. 

in addition, i'm concerned that support requests that do come in will leave us without the knowledge as maintainers around how to investigate. we already had a [recent request](https://github.com/semantic-release/semantic-release/issues/2973) come in that turned out to be an issue with an invalid dependency tree. i still dont know if the underlying problem was something different about how bun builds the dependency tree (especially since the version of execa by the part of the tree that defined it was not honored), or just how that team was managing their project.

> Bun reports its Node version as 18.15

this confuses/concerns me. if this is "a Bun-only environment (no Node)", this seems really misleading. part of me wants to solve this by defining a node support range that covers what bun reports so that we could avoid defining bun support specifically. however, that leaves us in a dangerous spot of breaking bun support the next time the supported node ranges are restricted further. protecting from that would require a ci run that tests bun compatibility, which puts us right back in the situation of semi-officially adding bun support

does bun happen provide a way to define supported version ranges along the lines of `engines.bun`? if we do decide to add bun support, i think i'd at least prefer to define supported range(s) and enforce similarly to how we do node.

> The primary reason we hard fail with that check is specifically to minimize the potential runtime environments that can result in support requests.

our normal recommendation when our required node ranges are more restrictive than our users' projects is to have a specific release job that defines a node version that is supported by semantic-release, even if it might be different than other jobs in the pipeline. would something similar be possible in a bun project where only the release job uses node?

seebeen:
Bun released its first "stable" version couple of weeks back?  (By stable - I mean v1)

Implementing Bun "support" in this manner invites all sorts of exceptions to make further down the line (for other libs, runtimes, envs, etc...).

Maybe get someone from Bun in order to find a common solution that goes both ways?

jamonholmgren:
Hey thanks for giving my request some consideration! I'm a big fan of semantic-release.

> The primary reason we hard fail with that check is specifically to minimize the potential runtime environments that can result in support requests.

I absolutely understand this. As an open source maintainer myself, I'm very much on your side.

> does bun happen provide a way to define supported version ranges along the lines of engines.bun?

This I don't know, to be honest. Summoning @Jarred-Sumner, as I feel like he'd probably want to know about this sort of situation and may have some thoughts.

> would something similar be possible in a bun project where only the release job uses node

I'm really reluctant to lose the speed savings that I'm getting by using `semantic-release` via bun in CI. It's really working well for me and the job runs so much faster.

Honestly: I don't mind doing the hacky workaround that I'm doing. I can maintain this hack and am willing to indefinitely, and I've had zero issues with it.

However, if there could be a less hacky way to override the node version, I'd use it. Something like 

```
"overrideVoidsWarrantyDoNotFileIssuesIfYouUseThis": {
  "node": "xxxx"
}
```

Regarding Bun just being hype: maybe, but I'm having way more fun writing for Bun than I ever have for Node. Take that for what you will. :-)

Jarred-Sumner:
> > I am worried of additional maintenance overhead from Bun users though, to be honest.
> 
> This is my biggest concern as well. The primary reason we hard fail with that check is specifically to minimize the potential runtime environments that can result in support requests.

Bun aims to be a drop-in replacement for Node, which means that if the code doesn't work in Bun but does work in Node, it's a bug in Bun. If people file issues in any of your repositories where something is broken in Bun but works in Node, please tell them to file an issue in Bun's repo to fix it – it's our job to fix bugs in Bun. We don't want to increase the maintenance burden of libraries.

> > in addition, i'm concerned that support requests that do come in will leave us without the knowledge as maintainers around how to investigate. we already had a [recent request](https://github.com/semantic-release/semantic-release/issues/2973) come in that turned out to be an issue with an invalid dependency tree. i still dont know if the underlying problem was something different about how bun builds the dependency tree (especially since the version of execa by the part of the tree that defined it was not honored), or just how that team was managing their project.

This sounds like a bug in `bun install`

> > Bun reports its Node version as 18.15
> 
> this confuses/concerns me. if this is "a Bun-only environment (no Node)", this seems really misleading.

This is unfortunately necessary - many npm packages break if the Node.js version is unexpected or non-existant. It's a little like relying on a `User-Agent` header in frontend code.  There is no "Bun-only environment". Bun implements many Node.js globals like `process`, `Buffer` (along with web globals like `fetch`, `Headers`, `Response`, etc)


> does bun happen provide a way to define supported version ranges along the lines of `engines.bun`? if we do decide to add bun support, i think i'd at least prefer to define supported range(s) and enforce similarly to how we do node.
> 
> > The primary reason we hard fail with that check is specifically to minimize the potential runtime environments that can result in support requests.
> 
> our normal recommendation when our required node ranges are more restrictive than our users' projects is to have a specific release job that defines a node version that is supported by semantic-release, even if it might be different than other jobs in the pipeline. would something similar be possible in a bun project where only the release job uses node?

We have been planning on adding support for the `"engines"` field, but haven't gotten around to it yet.

> https://dev.to/thejaredwilcurt/bun-hype-how-we-learned-nothing-from-yarn-2n3j - an honest criticism

Here is my response to that blog post: https://twitter.com/jarredsumner/status/1704261410136752529

jamonholmgren:
@Jarred-Sumner mentioned that he might bump the Bun "node" version to 20.x, which would solve this issue. If not, I do also have the workaround mentioned above.

I'm going to close this PR. I appreciate the attention, and I appreciate the work you do on semantic-release!

travi:
thanks for the additional feedback and understanding, @jamonholmgren and @Jarred-Sumner!

> Bun aims to be a drop-in replacement for Node, which means that if the code doesn't work in Bun but does work in Node, it's a bug in Bun. If people file issues in any of your repositories where something is broken in Bun but works in Node, please tell them to file an issue in Bun's repo to fix it – it's our job to fix bugs in Bun. We don't want to increase the maintenance burden of libraries.

really appreciate you sharing these details. this removes my concern about any additional support coming from use in a Bun context.

> We have been planning on adding support for the `"engines"` field, but haven't gotten around to it yet.

i do think this would be helpful if we get to a point of adding official support for Bun. however, with the above understanding, i think we can avoid the need since we can silently support it without needing to be overly restrictive.

> > > Bun reports its Node version as 18.15
> > 
> > 
> > this confuses/concerns me. if this is "a Bun-only environment (no Node)", this seems really misleading.
> 
> This is unfortunately necessary - many npm packages break if the Node.js version is unexpected or non-existant. It's a little like relying on a `User-Agent` header in frontend code. There is no "Bun-only environment". Bun implements many Node.js globals like `process`, `Buffer` (along with web globals like `fetch`, `Headers`, `Response`, etc)

appreciate this explanation and it does make sense. i've worked in the front-end for long enough to completely understand this need related to complications with browser support.



> @Jarred-Sumner mentioned that he might bump the Bun "node" version to 20.x, which would solve this issue

this feels like the ideal solution here. we are admittedly pretty aggressive with our required node versions, but there are reasons behind our minimum decisions. for example, v18.17 is because of the minimum defined by npm v10, which is a direct dependency of our npm plugin. v20.6.1 is a combination of support for auto-registering of loaders and a node bug for v20.6.0. as long as Bun can stay ahead of (or respond to user requests for updates) what we define our requirements to be, I see no issue with just continuing to use semantic-release as-is in a Bun context.

PR #2993, opened 2023-10-01
docs: correct FAQ note about 0.x major versions

### LINKED ISSUE #2639: Docs: incorrect phrasing about "best practice" and 0.X.Y versions
It seems from https://github.com/semantic-release/semantic-release/issues/1507#issuecomment-671514974 that the reason for closing #1507 was more that it's too big of an ask for the maintenance team. I'd also like to amplify @jcornaz's https://github.com/semantic-release/semantic-release/issues/1507#issuecomment-605420041:

> @gr2m, @travi, Since that FAQ states "it's not considered a good practice", could it provide sources about that?
> 
> Right now, the links provided by the FAQ are:
> 
> * a blog post that indeed seems to be written by someone who doesn't like `0.y.z` versions
> * the semantic versioning specification that says to start version at `0.1.0`
>   * By the way, the title of the FAQ is incorrect, semantic versioning starts at `0.1.0`, not `0.0.1`
>   * And fun fact, even semantic versioning specification itself started at `0.1.0`: https://semver.org/spec/v0.1.0.html
> 
> Unless other sources are provided, I don't see a reason to favor a random blog post over an official specification.
> 
> Could you maybe provide more sources of your statement? Or maybe reword the FAQ answer, so that we understand that it is an opinionated point of view that `semantic-release` takes?
> 
> And of course, don't forget to remove the link to the semantic versioning specification in that FAQ. Since it says the opposite, it only adds to the confusion ;-)

Semver itself allows for -and even seems to encourage!- a leading 0 per https://semver.org/#how-should-i-deal-with-revisions-in-the-0yz-initial-development-phase:

> The simplest thing to do is start your initial development release at 0.1.0 and then increment the minor version for each subsequent release.

Can we update the docs to not put down the idea of a 0.1.0?

```diff
## Can I set the initial release version of my package to 0.0.1?

- This is not supported by semantic-release as it's not considered a good practice, mostly because [Semantic Versioning](https://semver.org/) rules applies differently to major version zero.
+ This is not supported by semantic-release. [Semantic Versioning](https://semver.org/) rules apply differently to major version zero and supporting those differences is out of scope.
```

JoshuaKGoldberg:
Fixes #2639.

Switches the FAQ page from taking a stance on the issue to explicitly noting it as out of scope.

### DIFF
--- a/docs/support/FAQ.md
+++ b/docs/support/FAQ.md
@@ -192,7 +192,7 @@ If you need more control over the timing of releases, see [Triggering a release]
 
 ## Can I set the initial release version of my package to `0.0.1`?
 
-This is not supported by semantic-release. [Semantic Versioning](https://semver.org/) rules apply differently to major version zero and supporting those differences is out of scope.
+This is not supported by semantic-release. [Semantic Versioning](https://semver.org/) rules apply differently to major version zero and supporting those differences is out of scope and not one of the goals of the semantic-release project.
 
 If your project is under heavy development, with frequent breaking changes, and is not production ready yet we recommend [publishing pre-releases](../recipes/release-workflow/pre-releases.md#publishing-pre-releases).
 
--- a/docs/support/FAQ.md
+++ b/docs/support/FAQ.md
@@ -192,7 +192,7 @@ If you need more control over the timing of releases, see [Triggering a release]
 
 ## Can I set the initial release version of my package to `0.0.1`?
 
-This is not supported by **semantic-release** as it's not considered a good practice, mostly because [Semantic Versioning](https://semver.org) rules applies differently to major version zero.
+This is not supported by semantic-release. [Semantic Versioning](https://semver.org/) rules apply differently to major version zero and supporting those differences is out of scope.
 
 If your project is under heavy development, with frequent breaking changes, and is not production ready yet we recommend [publishing pre-releases](../recipes/release-workflow/pre-releases.md#publishing-pre-releases).
 


gr2m (approved):
I like it! I'll merge in but @travi / @seebeen if you have any objections we can revise. 

Thank you @JoshuaKGoldberg always nice to see you around 😁

travi:
i dont have any objections. i think this is an improvement over what we had in place. 

however, my hesitation is that i believe the origin of the previous version hinted at the idea of semantic-release being opinionated about the topic, but maybe approached the communication of the stance too strongly. i think there is value in highlighting that semantic-release is not simply an implementation of the semver spec, but an opinionated guide about how certain aspects about what a change _is_ results in an appropriate bump to communicate the impact of the change to consumers. 

thats maybe a long winded way to say that "out of scope", while not inaccurate, sounds like we are choosing not to implement it simply because of the work involved. instead, we are actually recommending against the practice, even though it is allowed by the semver spec. 

subtle difference and not enough to avoid moving forward with this improvement, especially since our recommendation still follows. i just wonder if strengthening the rewording slightly might help us avoid folks still asking for it to be added or preparing a PR to add it for us when i would still consider it outside the goals of the project.

docs/support/FAQ.md
```diff
@@ -192,7 +192,7 @@ If you need more control over the timing of releases, see [Triggering a release]
 
 ## Can I set the initial release version of my package to `0.0.1`?
 
-This is not supported by **semantic-release** as it's not considered a good practice, mostly because [Semantic Versioning](https://semver.org) rules applies differently to major version zero.
+This is not supported by semantic-release. [Semantic Versioning](https://semver.org/) rules apply differently to major version zero and supporting those differences is out of scope.
```
JoshuaKGoldberg, line 195:
```suggestion
This is not supported by semantic-release. [Semantic Versioning](https://semver.org/) rules apply differently to major version zero and supporting those differences is out of scope and not one of the goals of the semantic-release project.
```

@travi following https://github.com/semantic-release/semantic-release/pull/2993#issuecomment-1743837952, what do you think of this addition?

travi: That works for me. Thanks for humoring my long winded justification :)

gr2m: I like it 👍🏼

JoshuaKGoldberg: Yay! I like having the goals clarification a lot, +1 that without it it's kind of unclear _why_ pre-0 is out of scope.

(sorry for responding to three paragraphs with a rushed answer - but yes agreed all around, and I really appreciate the write-up!)

travi (approved):

travi:
Thanks!

PR #3064, opened 2023-11-18
fix: explicitly specify git tag message

sheerlox:
Fixes https://github.com/semantic-release/semantic-release/issues/3065.

### DIFF
--- a/lib/git.js
+++ b/lib/git.js
@@ -223,7 +223,7 @@ export async function verifyAuth(repositoryUrl, branch, execaOptions) {
  * @throws {Error} if the tag creation failed.
  */
 export async function tag(tagName, ref, execaOptions) {
-  await execa("git", ["tag", tagName, ref], execaOptions);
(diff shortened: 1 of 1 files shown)


travi:
this is a bigger change than it may seem. this changes the resulting tags from lightweight tags to annotated tags. while it could be argued that annotate tags are better for releases, i think that change would be far more than just a bug fix. in addition, enabling annotated tags was attempted previously, but it resulted in problems. i dont have time to track that down right now, but it would be good for one of us to link that previous attempt to this conversation.

i'm not against making this possible at some point in the future, but would want to do it very carefully. if there is a different solution to the referenced problem, i'd prefer a change that isnt quite this impactful

sheerlox:
I wasn't aware of the difference between annotated and lightweight tags, thanks for pointing out that adding the `-m` option changes the type of tag being created!

Closing this PR as it's not the right solution (at least without a more thoughtful approach), let's track this in #3065.

Issue #3085, opened 2023-12-03
Deprecated dependency `read-pkg-up`


McLaynV:
### Current behavior

```
> npm install

npm WARN deprecated read-pkg-up@11.0.0: Renamed to read-package-up
```

### Expected behavior

Switch from `read-pkg-up` to the new package [`read-package-up`](https://www.npmjs.com/package/read-package-up)

### `semantic-release` version

22.0.8

### CI environment

any

### `semantic-release` configuration

no

### CI logs

```
npm WARN deprecated read-pkg-up@11.0.0: Renamed to read-package-up
```

travi:
thanks for reporting!

i've submitted [a PR](https://github.com/renovatebot/renovate/pull/26101) to renovate to automate replacement for this rename. assuming that is accepted, we will get a PR from renovate to make the change.

travi:
Haven't seen PRs come in from renovate about this yet, so opened https://github.com/renovatebot/renovate/discussions/26469 to discuss what might be preventing it.

travi:
thanks for linking the update! that at least makes sense. i expect that we'll get a PR to handle this once the version with that fix gets deployed.

ardunster:
Any news on this?

ardunster:
Apparently this fix got applied in this commit: https://github.com/semantic-release/semantic-release/commit/4980cba57169048ae1cd6ba869170da980b30fe3 and this thread was never updated.

travi:
unfortunately, the PRs from renovate related to this still seem to be getting stuck in situations like this. i havent had a chance to follow up to help the renovate team understand the situation, but yes, this case has been resolved. i'll close this issue because of that

PR #3127, opened 2024-01-01
fix(deps): rename read-pkg-up to read-package-up

kirkeaton:
Package was renamed from read-pkg-up to read-package-up


travi:
thanks for the nudge on these. i'm going to close as a duplicate of https://github.com/semantic-release/semantic-release/issues/3085, but i need to investigate why the renovate replacement hasn't worked.

Issue #3144, opened 2024-01-16
Allow Configuration of when tags are applied


hanseltime:
### New feature motivation

I recently ran into an issue where our npm registry was down and failed to deploy our packages.  "No Problem, I thought.  I'll just re-run the commit since there shouldn't be a tag yet".

However, when I ran the pipeline again, the release was skipped because there was already a tag applied that indicated we had successfully published.

I checked the source and realized that we are tagging our commit before publish succeeds.  This leads to "false positive" releases.

I would like for there to be a semantic-release config option that could be opt-in:

```typescript
interface Config {
  releaseTagAfter: 'publish' | 'prepare'
// The rest
}
```

### New feature description

A user could add this configuration field to their config if they wanted to change when the git tag was created.  If they chose 'publish' only after a successful publish stage would the tag get added.  Likewise with prepare.

To keep parity with the legacy config, we would just default the releaseTagAfter field to be 'prepare'.

This also has the benefit of allowing releasers to handle any plugins that make assumptions about the tag being applied during prepare - so the user can always go back to the original.

### New feature implementation

The implemenation would involve making the publish methods a function and then evaluating the new config field to determine when to call it.  We would then apply 'prepare' as the default.

travi:
> I checked the source and realized that we are tagging our commit before publish succeeds.

currently, this is necessary to support the flexibility that our plugin system enables. some existing plugins require the tag to already exist so they can carry out their parts of a release

> No Problem, I thought. I'll just re-run the commit since there shouldn't be a tag yet

as you likely figured out, the current necessary approach is to manually remove the tag before trying again. we would love to get to a point where semantic-release can make release steps be fully atomic, but it is a difficult feat to accomplish when coordinating so many independent systems. it would at least require having each step configure how to undo a step if a later step fails. this gets more difficult when considering that some systems (npm for example) do not enable an undo that could be followed up by a redo (even if you unpublish a version from npm in the short window it is allowed, you can never republish that version again in the future). 

we are aware of the value of handling this and get plenty of requests to handle it better. we may get to a point where we approach solving this better, but would be very particular about how we expose it. we are very cautious about adding more config options in any form to our already complex config system, so we would likely prefer to approach this in a way where semantic-release can handle it automatically without additional config

we appreciate the feedback. since we arent at a point where we can accept a change like this and have other issues tracking the feature request here, i'm going to close this issue

PR #3145, opened 2024-01-16
feat(config): add tagReleaseAfter to config

hanseltime:
# Summary

This allows the configurers of semantic-release to determine after which lifecycle, the tag for a release should get added.

Addresses: https://github.com/semantic-release/semantic-release/issues/3144

* introduces a '--tag-release-after' and 'tagReleaseAfter' config parameter.
* Defaults to the legacy 'prepare' stage
* Provides documentation with a suggestion to not use prepare for the sake of replay but with an acknowledgement that some plugins may not play nice
* Creates a new Error State for verification
* Adds some constants for lifecycle methods

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -37,6 +37,7 @@ Usage:
     .option("debug", { describe: "Output debugging information", type: "boolean", group: "Options" })
     .option("d", { alias: "dry-run", describe: "Skip publishing", type: "boolean", group: "Options" })
     .option("h", { alias: "help", group: "Options" })
+    .option("tag-release-after", { describe: "After when to apply the release tag", type: "string", group: "Options" })
     .strict(false)
     .exitProcess(false);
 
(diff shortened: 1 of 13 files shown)


hanseltime:
@travi - please let me know if I'm abusing @'s on this repo.  I see that you have been commenting on some PRs and issues and was wondering if there is a step I'm missing in contribution to get PRs into a visible spot.

Apologies if this is the wrong way to go about it and thanks!

hanseltime:
Thanks for the information @travi.  It wasn't too much time on my part luckily and is always worth the cost of getting associated with an open source project.  Hopefully, I can contribute in the future.  

Unfortunately, the principle of marking a change as finished before it actually is, is a pretty critical bug for longevity of use of this product - since it will subsequently report to users of the product that it deployed successfully (even though all it did was just skip the deployment entirely due to the tag).  I will perhaps try to write a plugin that will remedy this bug before trying to maintain a fork with an opt-in flag.

My one reservation about such a plugin is that, fundamentally, the plugin needs to untag or edit a tag on the branch when a failure is detected, which feels far more dangerous than allowing users to acknowledge a config set of "tag after the publish command - note some of our old plugins don't play well with this so keep this in mind when you do your switch".

I'll give it some thinking over and hopefully get back to fixing this bug in a different way in the near future.

PR #3159, opened 2024-01-25
Draft: Add config to provide git push options

MatthewJohn:
I've created this change for myself (currently in testing) and created the PR to see if you believe it would be useful.

The rational behind this:
 * I use Gitlab primarily but replicate to Github.
 * Some projects use both Gitlab CI/CD and Github actions (mainly for testing/tagging etc. in Gitlab and then producing artifacts in Github).

When I merge into Gitlab, semantic release runs on main and performs the release. The release commit, by default, uses `[skip ci]`, which is appropriate for Gitlab. Unfortunately, when the release commit is replicated to Github, it _also_ respects this and skips Github actions.

This change means that I change the commit message, removing [skip ci] and use a git push option `ci.skip` (see https://docs.gitlab.com/ee/user/project/push_options.html), which will only skip the CI/CD pipeline in Gitlab.

I'm creating this PR to get feedback if it's a feature that you believe would be useful and, if so, I can tidy it up after testing, add tests etc. ready for proper review :)

Many thanks
Matt

### DIFF
--- a/index.js
+++ b/index.js
@@ -127,8 +127,8 @@ async function run(context, plugins) {
           cwd,
           env,
         });
-        await push(options.repositoryUrl, { cwd, env });
-        await pushNotes(options.repositoryUrl, nextRelease.gitTag, {
+        await push(options.repositoryUrl, options.gitPushOptions, { cwd, env });
+        await pushNotes(options.repositoryUrl, options.gitPushOptions, nextRelease.gitTag, {
           cwd,
           env,
         });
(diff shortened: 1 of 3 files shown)


maximgubar:
@MatthewJohn since this PR is still in draft I assume you have some temporary solution atm?
would be really interesting to see it, cause for me this 2 pipelines on each release is a bit annoying :)

MatthewJohn:
> @MatthewJohn since this PR is still in draft I assume you have some temporary solution atm? would be really interesting to see it, cause for me this 2 pipelines on each release is a bit annoying :)

Hey @maximgubar,

I'm afraid I really can't remember - I don't think it was working exactly as I wanted it to. I'll take a look to see if I can remember which project I was trying to use this for and if I did have a workaround. Prior to making this change, the main thing I did was to adjust the commit message format to no longer include '[skip ci]' (https://github.com/DockStudios/jmon/blob/main/.releaserc#L20) and then added rules in the Gitlab workflow to ignore commit messages that were release-like (https://github.com/DockStudios/jmon/blob/main/.gitlab-ci.yml#L57).

I'll take a look to see if I can find evidence of a project using the changes in this PR to determine if they worked :D 

Matt

Issue #3046, opened 2023-11-03
feature request: consider supporting a workflow to enable 


travi:
### New feature motivation

the conversation in https://github.com/semantic-release/semantic-release/issues/2503 resulted in the realization that the true desire is support for an additional workflow beyond those currently officially supported by semantic-release. we've heard plenty of other requests for similar workflows in other previous threads as well.

while the officially supported [pre-release workflow](https://semantic-release.gitbook.io/semantic-release/recipes/release-workflow/pre-releases) can be used in a way that can accomplish a similar outcome, but it is worth noting that pre-releasing every release isn't the intended use of that officially supported workflow.

### New feature description

verification for some projects requires testing beyond the scope of the ci pipeline of the project itself. some example use cases are described:
* by @dominykas: https://github.com/semantic-release/semantic-release/issues/2503#issuecomment-1788538343
* by @hutchic: https://github.com/semantic-release/semantic-release/issues/2503#issuecomment-1789259320

our typical response to such requests is that semantic-release is opinionated in ways that strongly favor continuous integration and continuous deployment. part of being successful with those practices is isolating the project and having strong test coverage to ensure that the behaviors intended to be available from the public interface of the project are fulfilled before release or integration into consumers. a goal of this thread is to have a conversation about where that stance might miss real world needs and what accounting for that might mean for the semantic-release project

travi:
> The goal here is to pre-release without pushing a commit at all - sort of like CITGM. This is completely automated.

i really like this sort of description for this sort of workflow. calling it something like a canary helps distinguish this desire from the intended use of the existing pre-release workflow

travi:
one question that i think needs to be clarified with the desired workflow here is how a pre-release/canary gets promoted to be a stable release. in the official pre-release workflow, the merge to the release branch accomplishes this goal. all releases before merge are pre-releases. merging to the release branch promotes to stable

in https://github.com/semantic-release/semantic-release/issues/2503#issuecomment-1789259320, @hutchic mentions:

> A human is required to trigger the GitHub Action workflow for a proper release, which uses a separate releaserc.json configuration to manage this process.

is something similar needed for the use case you describe @dominykas?

travi:
> our typical response to such requests is that semantic-release is opinionated in ways that strongly favor continuous integration and continuous deployment.

this comment from @hutchic is worth considering in this context:

[...]

PR #3198, opened 2024-02-17
feat: added override to prerelease version numbers

JTeeuwissen:
Fixes: https://github.com/semantic-release/semantic-release/issues/1793

Modified version of: https://github.com/semantic-release/semantic-release/pull/1894

The aim of this PR is to allow prerelease version overrides, without resorting to build tags. Which, although supported by SemVer, are not supported by NuGet (which removes the info, causing problems when uploading the same version with different build info).
Overriding this version allows e.g. the pipeline id to be used for increasing versions, even after force pushing to the prerelease branch.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -37,6 +37,7 @@ Usage:
     .option("debug", { describe: "Output debugging information", type: "boolean", group: "Options" })
     .option("d", { alias: "dry-run", describe: "Skip publishing", type: "boolean", group: "Options" })
     .option("h", { alias: "help", group: "Options" })
+    .option("prerelease-version", { describe: "Prerelease version number override", type: "string", group: "Options" })
     .strict(false)
     .exitProcess(false);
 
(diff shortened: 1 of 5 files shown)


L3o-pold:
@JTeeuwissen can you show an example of `prerelease-version` value? I don't get it reading the doc and code.

JTeeuwissen:
@L3o-pold Sure, Azure Devops names their pipelines like so: `#RepoName_CI_2024.02.19.1`. Thus, we can use the following value:
`${build.replace(/^.*_(\d*).(\d*).(\d*).(\d*)$/, '$1-$2-$3.$4')}` to turn the prerelease version into something like this: `2024-02-19.1`. With version it could look like: `v1.2.3-beta.2024-02-19.1`. Which works across force pushes and on NuGet while remaining ordered chronologically.

L3o-pold:
@JTeeuwissen from where `commit` and `build` and populated?

JTeeuwissen:
envCi is populated using the [env-ci](https://www.npmjs.com/package/env-ci) package, which is given the env and cwd. But that already was a part of the existing context.

travi:
closing based on https://github.com/semantic-release/semantic-release/issues/1793#issuecomment-1952627408

this could potentially be a better fit related to #3046 than related to the existing pre-release workflow. worth noting that we have not yet accepted that request as a type of workflow that we would be willing to support within semantic-release

PR #3205, opened 2024-02-22
feat: Add support for a single config file in a monorepo

RenatoMPestana:
# Issue

The new version of `cosmiconfig` defaults the `searchStrategy` to `none`, which breaks support for a single root config file for monorepos, when running semantic-release on individual packages, because it no longer recursively looks for the file on parent folders.

# Solution

`cosmiconfig` still allows for  a `stopDir` path to be set, wich will set the `searchStrategy` to `global` and allow for the config to be searched on parent folders, we just need a way to pass that through `semantic-release` cli first, so its then passed as an option to `cosmiconfig`

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -36,6 +36,11 @@ Usage:
     .option("fail", { ...stringList, group: "Plugins" })
     .option("debug", { describe: "Output debugging information", type: "boolean", group: "Options" })
     .option("d", { alias: "dry-run", describe: "Skip publishing", type: "boolean", group: "Options" })
+    .option("stopDir", {
+      describe: "Sets the stop path for recursive search of release config files",
+      type: "string",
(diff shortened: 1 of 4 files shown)


travi:
> which breaks support for a single root config file for monorepos

monorepos are not officially supported by the semantic-release project. in the future, please gain maintainer acceptance of a new feature before creating a PR. we are very hesitant to add more configuration options to our already complex configuration, especially when they are specifically for adding functionality that is only necessary in the context of a monorepo.

the best approach that i can recommend for accomplishing your goal would be to define a release config for each released directory in your monorepo and use the [extends](https://semantic-release.gitbook.io/semantic-release/usage/configuration#extends) option to point to your shareable config, wherever that might be.

PR #3207, opened 2024-02-23
docs: updated installation docs so that user can copy paste without error

jonmejia:
…into terminal without editing after pressing the copy button

I frequent these docs and use semantic release in multiple projects, it's mildly annoying that whenever I click the copy button it comes with the '$' character that my terminal does not recognize. I think it would be a much better user experience if users could simply click the copy button and paste it into their terminal without any issues.


travi (approved):
thanks!

PR #3242, opened 2024-03-30
docs: add community plugin semantic-release-pull-request-analyzer

bobvanderlinden:
This plugin is for teams that like to avoid conventional commits, but do want to make use of the plugin ecosystem of semantic-release.

`semantic-release-pull-request-analyzer` replaces `analyzeCommits` and `generateNotes`, so that it can find the release-type based on merged pull request labels, instead of commit subjects.


gr2m (approved):
Nice 👍🏼

PR #3264, opened 2024-04-19
wip(supported-workflows-docs): captured a start to a structure

travi:
clarifying these considerations is long overdue. obviously, this isnt yet in a form that we would release, but i started to capture thoughts that have been shared consistently in support discussions to start refining into something more helpful that should be easier to point to when questions continue to come up.

### DIFF
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -2,11 +2,13 @@
 
 ## Usage
 
+- [Considerations](docs/usage/considerations.md)
 - [Getting started](docs/usage/getting-started.md)
 - [Installation](docs/usage/installation.md)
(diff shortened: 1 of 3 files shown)


babblebey:
This will make for a good new sections, I've got in mind a "Foundations" section which should hold sub pages that explains concepts, ideas and thoughts that makes semantic-release. 

Things like...

- Plugins
- Shareable Configurations
- Release Workflows

These are mostly items, that links back to extending and using of semantic-release, the what, why and hows.

- Foundation
  - Considering 
  - Plugins
  - Shareable Configurations 
  - Release Workflows

What'd you think? 🤔

PR #3275, opened 2024-04-26
docs(ci-configurations): update github token URL

ChristopherBull:
Updated a URL for `GITHUB_TOKEN` documentation (which no longer describes how tokens are auto-generated for CI), and replaced it with another appropriate GitHub documentation URL.


travi (approved):
thanks!

PR #3291, opened 2024-05-06
docs: Remove non-existing plugin from  plugins-list.md

Piwero:
There isn't such a plugin in https://github.com/m1pl and link is broken.


gr2m (approved):
Thank you!

PR #3377, opened 2024-07-02
ref: Remove usage of aggregate-error library

AbhiPrasad:
ref https://github.com/semantic-release/semantic-release/discussions/3376
ref https://github.com/es-tooling/module-replacements/issues/80

As per the `package.json` `engines` of this package, `semantic-release` supports Node.js `>=20.8.1`.

As per https://github.com/sindresorhus/aggregate-error

> Note: With [Node.js 15](https://medium.com/@nodejs/node-js-v15-0-0-is-here-deb00750f278), there's now a built-in [AggregateError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError) type.

This means we can remove the usage of this package and use the JavaScript built-in `AggregateError`.

### DIFF
--- a/docs/developer-guide/plugin.md
+++ b/docs/developer-guide/plugin.md
@@ -54,8 +54,6 @@ module.exports = { verifyConditions };
 Then, in your `src` folder, create a file called `verify.js` and add the following
 
 ```javascript
-const AggregateError = require("aggregate-error");
-
 /**
  * A method to verify that the user has given us a slack webhook url to post to
  */
--- a/index.js
+++ b/index.js
@@ -4,7 +4,6 @@ import * as marked from "marked";
 import envCi from "env-ci";
 import { hookStd } from "hook-std";
 import semver from "semver";
-import AggregateError from "aggregate-error";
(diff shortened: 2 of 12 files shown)


travi:
we need to consider whether this is a breaking change, especially since we use it in core and across plugins to aggregate for a full execution. see https://github.com/semantic-release/gitlab/pull/469#issuecomment-1408281435 for us considering this previously

AbhiPrasad:
@travi from looking at the existing usages, I don't believe this is breaking. This is because:

1. I can't find any plugin that relies on `aggregate-errors` behaviour to create errors from non-error objects.

https://github.com/sindresorhus/aggregate-error?tab=readme-ov-file#errors

>  If a non-Error object, a new Error is created with all properties from the object copied over.

2. I can't find any plugin that relies on multiple `error.stack` being part of the overall `error.message` (https://github.com/sindresorhus/aggregate-error/issues/4#issuecomment-488356468)

> aggregate-error puts all the stacks of the errors into message, but imo message should only contain the aggregated messages

Are there specific things that you are looking from `AggregateError`?

PR #3389, opened 2024-07-15
feat: capture output of prepare plugin

rossreicks:
I would like a way to generate additional notes in my plugin that utilizes the prepare stage.

This way I can update my github plugin config like the following:

```js
releaseNameTemplate:
                    "<%= nextRelease.name %> (<%= nextRelease.prepareOutput.subscriberPackageVersionId %>)"
```

if there is a different way I can tie into that I would appreciate knowing how. I have attempted to mutate the context but it seems that the context is not a global reference one it reaches the plugin.

### DIFF
--- a/index.js
+++ b/index.js
@@ -199,7 +199,7 @@ async function run(context, plugins) {
 
   nextRelease.notes = await plugins.generateNotes(context);
 
-  await plugins.prepare(context);
+  nextRelease.prepareOutput = await plugins.prepare(context);
 
   if (options.dryRun) {
     logger.warn(`Skip ${nextRelease.gitTag} tag creation in dry-run mode`);


travi:
could you expand on your use case? we've never had a request for this before, so i'm pretty hesitant to accept a change like this for a limited use case. understanding your goal might help us consider alternatives

rossreicks:
> could you expand on your use case? we've never had a request for this before, so i'm pretty hesitant to accept a change like this for a limited use case. understanding your goal might help us consider alternatives

Understandable! My use case is that I would like to have a deep link to my published package in my github release notes. For example, lets say the npm plugin deploys a package to the registry [here]( https://www.npmjs.com/package/semantic-release/v/22.0.12), I would like to deep link to this in the changelog.

I'm sure there is a better way, but right now im thinking if the prepare part of the plugin (an unoffical custom plugin) can set the UUID of the release version, I could tie into that to build the deep link

rossreicks:
I have a workaround; I can just write to an HTML file or something in my plugin and then use the assets array in the github plugin to link to that html file. Should be good enough!

Issue #3478, opened 2024-10-16
Allow release branches to have names like `N.N`  (`N` is a number)


abichinger:
### New feature motivation

I'm currently trying to use `semantic-release` for an Odoo module. To publish a module on the Odoo Apps store you have to ensure that:

> The branch name exactly matches the series name for which your modules are meant, so `17.0` for  Odoo version 17.0.

When I try to release from branch `17.0` I get the following error:

```log
[9:08:25 PM] [semantic-release] › ✘  ERELEASEBRANCHES The release branches are invalid in the `branches` configuration.
```

Because branch `17.0` is treated as a maintenance branch.

My current configuration:
```js
/**
 * @type {import('semantic-release').GlobalConfig}
 */
export default {
  branches: [{ name: "17.0", range: false }],
  tagFormat: "v17.0.${version}",
  plugins: [
    "@semantic-release/commit-analyzer",
    "@semantic-release/release-notes-generator",
    [
      "@semantic-release/git",
      {
        assets: ["CHANGELOG.md", "__manifest__.py", "**/static/**"],
      },
    ],
    [
      "@semantic-release/exec",
      {
        prepareCmd:
          "python ./scripts/version.py ${branch.name}.${nextRelease.version}",
      },
    ],
  ],
};
```

### New feature description

Allow the user to define release branches with names like `N.N` by explicitly setting `BranchObject.range` to `false`

abichinger:
Looking at the test case [`A "maintenance" branch is identified by having a "range" property or a "name" formatted like "N.x", "N.x.x" or "N.N.x"`](https://github.com/semantic-release/semantic-release/blob/c87596ee1603840d50822e002cea837209b3a243/test/definitions/branches.test.js#L4) this seems to be a bug, not a feature request.

travi:
The current implementation is not a bug. As you can see from that test description, it is intended to describe a range, not a specific version. The purpose of our support for maintenance branches is to enable releasing additional changes that are before the latest, such as v2.x after v3.0.0 has been released. 

use of our maintenance workflow does not support the workflow you describe and will not be modified to add support. to support what you describe would require us to add support for a new workflow. Since our goal is to promote a continuous deployment style workflow, we would be very hesitant to add a workflow that targets every release to be a branch rather than from the default branch. 

You are welcome to attempt to convince us that this would be valuable, but it is likely that the workflow required by that ecosystem will remain incompatible with semantic-release

abichinger:
[...]

PR #3479, opened 2024-10-16
fix: regex to identify maintenance branches

abichinger:
Fixes: #3478

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -27,7 +27,7 @@ export function isMajorRange(range) {
 }
 
 export function isMaintenanceRange(range) {
-  return /^\d+\.(?:\d+|x)(?:\.x)?$/i.test(range);
+  return /^\d+(?:\.(?:\d+|x))?\.x$/i.test(range);
 }
 
 export function getUpperBound(range) {
--- a/test/definitions/branches.test.js
+++ b/test/definitions/branches.test.js
@@ -15,6 +15,7 @@ test('A "maintenance" branch is identified by having a "range" property or a "na
   t.false(maintenance.filter({ name: "some-name", range: false }));
   t.false(maintenance.filter({ name: "some-name" }));
   t.false(maintenance.filter({ name: "1.0.0" }));
+  t.false(maintenance.filter({ name: "1.0" }));
   t.false(maintenance.filter({ name: "x.x.x" }));
   /* eslint-enable unicorn/no-fn-reference-in-iterator */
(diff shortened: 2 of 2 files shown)


lib/utils.js
```diff
@@ -27,7 +27,7 @@ export function isMajorRange(range) {
 }
 
 export function isMaintenanceRange(range) {
-  return /^\d+\.(?:\d+|x)(?:\.x)?$/i.test(range);
+  return /^\d+(?:\.(\d+|x))?(?:\.x)$/i.test(range);
```
travi, line 30:
```suggestion
  return /^\d+(?:\.(?:\d+|x))?\.x$/i.test(range);
```

lets keep the groups as non-capturing. also, the last segment no longer needs to be a group since it is no longer optional

travi, line 30: to help explain this change as we look back at it in the future, the key detail about this change is that it makes the second segment the optional one rather than the third segment. that ensures that the resulting version-ish branch name is a range when considered to be a match for a maintenance branch

travi (changes requested):
just a couple small tweaks that i think can clean up the regex slightly. open to further thoughts if you have some.

thanks for digging in far enough to identify this and sticking with it to get this resolved 👍🏼

lib/utils.js
abichinger: Why should we use non-capturing groups? Is it just because of better performance or is there another reason?

travi: just no need to do any capturing. mostly thinking about consistency so that we dont wonder in the future why one group is non-capturing and one group is capturing within a regex used for just testing if there is a match or not. i tend to get rusty with regex, so removing potential questions for future maintenance is my main goal. again, open to suggestions if there are reasons to do anything beyond what i'm thinking about

abichinger: > so removing potential questions for future maintenance is my main goal

Got it! Thank you for the explanation. I have nothing to add. :+1:

abichinger: I'm going to add this explaination to the commit message

travi (approved):

lib/utils.js
travi: if you are updating the commit message, lets also make the change summary a little more targeted at the reason for this. something like `fix(branch-naming): prevent non-range versions from being identified as maintenance branches` seems more descriptive to me

travi:
thanks again!

PR #3513, opened 2024-11-21
docs: update broken link to gitlab pipelines

el-memer:
Link to Gitlab documentation is broken, a little PR to fix it :)


el-memer:
Also, wercker link does not work either and point to nowhere but I don't know the good one

travi (approved):
thanks!

PR #3521, opened 2024-11-26
fix(ci): allow for publishing on allowed tag names

Killusions:
In my team we use a 2-step release, where we first commit and create the tag locally, then push the tag and run a GitLab CI job on the tag to publish to NPM and GL releases. This is currently not possible with semantic-release.

When a tag checked is out locally, env-ci.branch is the branch it is on, however, in CI environments it returns the tag name. This makes it impossible to publish from a tag, as tags will never be included in context.branches, no matter the user configuration.

This proposes adding tags to the list of possible branches in CI (locally this is not necessary and will fail for non-checked out branches), allowing for configurations like `branches: ['main', 'vx.x.x']`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -65,7 +65,7 @@ async function run(context, plugins) {
   await verify(context);
 
   options.repositoryUrl = await getGitAuthUrl({ ...context, branch: { name: ciBranch } });
-  context.branches = await getBranches(options.repositoryUrl, ciBranch, context);
+  context.branches = await getBranches(options.repositoryUrl, ciBranch, context, isCi);
   context.branch = context.branches.find(({ name }) => name === ciBranch);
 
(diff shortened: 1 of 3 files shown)


travi:
It seems like you are using semantic-release in a way that it is not intended. Automatically handling the version bump based on semantic commits is a key piece to the semantic-release approach. If you are pre-versioning, it seems like semantic-release might not be the tool you are looking for. 

Because this proposal deviates so much from the philosophy of semantic-release I'm going to decline the proposal.

Killusions:
@travi It seems you maybe misunderstood what I meant, I am still using semantic release for that, but running it locally. Semantic release locally creates the commit and tag, which is then pushed and creates a release on the tag pipeline

Issue #1592, opened 2020-06-30
Support custom location of the configuration file (`package.json`/`.releaserc`)


jcornaz:
## New feature motivation

When I have a complex release process with many scripts being invoked by `@semantic-release/exec`, then I'd like to put everything related to the release process into a `.release` folder of my repository. That way, I can keep everything neat and organized, rather that having everything on the root of the repository.

Yes I know, I can put my script anywhere I want. But, it turns out the `package.json` and `package.lock` (or `.releaserc`) are *also* release-related (in non-javascript project). So it would make sense, and improve organization to put these configuration files `.release` folder too. Otherwise, I am just confusing my teammates, who'd see a `.release` folder not containing the release process config...

## New feature description

It is already possible to use different [file name and format](https://semantic-release.gitbook.io/semantic-release/usage/configuration#configuration-file).

It would be nice to be also able to chose the location of the file.

Example: 

```
semantic-release --config .release/package.json
```

or 

```
semantic-release --config .release/.releaserc.yml
```

abbottdev:
Can you use the `--extends` parameter? You can pass it a filepath 😄

jcornaz:
> Can you use the --extends parameter? You can pass it a filepath 😄

Can you show a complete example? Sorry but it isn't clear to me.

I looked at https://semantic-release.gitbook.io/semantic-release/usage/configuration#options and thought `--extends` was for *shared configurations*. (which is not what I want)

Nevertheless from https://semantic-release.gitbook.io/semantic-release/usage/shareable-configurations I understood that a shared configurations is: 

> is an npm package that exports a semantic-release configuration object

But I don't want to create a npm package that and export an object.


Maybe what I'm asking for, is indeed already available, but only need to be clarified in the documentation.

dominykas:
> But I don't want to create a npm package.

The docs say:

> List of modules or file paths

It can be a simple module (i.e. a js file), if you give it a full path.

jcornaz:
>The docs say:
>
>> List of modules or file paths
>
> It can be a simple module (i.e. a js file), if you give it a full path.

The docs actually say:

> List of modules or file paths containing a shareable configuration.

(It seems like you missed the *containing a shareable configuration* part of the sentence).

Reading about [shareable configuration](https://semantic-release.gitbook.io/semantic-release/usage/shareable-configurations), it doesn't look like what I'm after.

[...]

Issue #1623, opened 2020-08-28
Prefixed maintenance branches


paulgessinger:
I'm not entirely sure if this is a bug or intended to be this way. Basically, I want my maintenance branches to be named like `release/v1.x.x`.

## Current behavior

Currently, a config like

```yml
branches:
  - master
  - 'release/v+([0-9])?(.{+([0-9]),x}).x'
```

I believe results in something like 

```yml
branches:
  - name: master
  - name: release/v1.x.x
```

when expanded and a branch `release/v1.x.x` exists. 
It does not, however, set a range on this second entry, so I guess it's treated like a pre-release branch?

## Expected behavior

I can make the configuration explicit such that `release/v+([0-9])?(.{+([0-9]),x}).x` can be expanded to maintenance branches. Maybe this could be done by applying an extracting regex on the name matched from the branch-name pattern above?

## Environment

I don't believe this depends on my environment.

gr2m:
If I recall correctly, the order of the branches configuration matters. The regex for maintenance branches has to be listed before your latetest release branch.

Can you try this instead?

```yml
branches:
  - 'release/v+([0-9])?(.{+([0-9]),x}).x'
  - master
```

dominykas:
I think the regex to set the range only works with the built-in regex or smth like that. IIRC it was hardcoded?

For any non-built-in regex matcher you have to set the range manually, unfortunately.

dominykas:
Iirc, this is somewhere in the codepath: https://github.com/semantic-release/semantic-release/blob/431d571a7b7284b2029a55da68a44c65d7c16451/lib/utils.js#L36

paulgessinger:
Hm, I see. Would you be ok to share (the relevant part of) the release.config.js you use to work around this, maybe?

dominykas:
We're running on Jenkins, so we have a `GIT_BRANCH` env var available, but you could read it by calling git.

Once done, we do roughly this (we use `release-N.x` as naming convention):

```
        const maintenanceMatches = gitBranch.match(/^release-(\d+).x$/);
        if (maintenanceMatches) {
            maintenance = maintenanceMatches[1]; // get the major version of the maintenance branch
        }
```

We then do roughly this:

```
    if (maintenance) {
        branches.push({
            name: 'release-+([0-9])?(.{+([0-9]),x}).x', // regex from default semantic-release config: '+([0-9])?(.{+([0-9]),x}).x'
            type: 'maintenance'
[...]

PR #3573, opened 2025-01-21
feat: allow manual config file selection

jonapich:
Add the `--config` CLI argument to specify the configuration file to load.

This is useful for users who require more control over their environment.

My use case is described in [this comment](https://github.com/semantic-release/semantic-release/issues/1592#issuecomment-2471211029).

Closes [#1592](https://github.com/semantic-release/semantic-release/issues/1592)


babblebey (commented):
Great stuff @jonapich 👍🏾 

I've got a follow observation/questions though but, I've got to ask this to @semantic-release/contributors first.... What are your thoughts about this?

<details>
<summary>Here's my own thought 🤔</summary>

I think that the [configuration options](https://semantic-release.gitbook.io/semantic-release/usage/configuration#options) CLI arguments already addresses this (as a way of consuming a certain config overriding the default `.releaserc`) to a good extent, but it requires writing a long line of command with arguments if you're looking to override more stuffs than just few properties.
</details>

sheerlox:
If the use case is limited to [working around prefix limitations on maintenance branches](https://github.com/semantic-release/semantic-release/issues/1592#issuecomment-2471211029) as mentioned by OP, is there anything preventing the use of [the current solution](https://github.com/semantic-release/semantic-release/issues/1623#issuecomment-871700509) for that specific issue?

Reading through #1592, being aware of this project's philosophy to limit the number of available options, and taking into account @babblebey's comment, I don't see a justification for implementing the `--config` flag. However, this is dependent on the answer to my question above.

jonapich:
Thanks for the feedback!

I believe the --extends is merged on top of the auto discovered configuration file?

We really want to skip any user-provided configuration and use our own. For instance, if the user has a package.json with the release key, we want to ignore it in favor of our own.

I linked a comment in my PR body that mentions this in other words. One additional challenge is a limitation that forces us to use a .mjs file, which is late in the order of auto-discovered files.

sheerlox:
> I believe the --extends is merged on top of the auto discovered configuration file?

I believe it's actually the opposite: the auto-discovered file is merged on top of the extends, meaning it takes precedence in case of conflicts. This makes sense as shared configs could be (for example) organization-wide defaults, and individual repositories might need to adjust based on their own needs.

> We really want to skip any user-provided configuration and use our own. For instance, if the user has a package.json with the release key, we want to ignore it in favor of our own.
> 
> I linked a comment in my PR body that mentions this in other words. One additional challenge is a limitation that forces us to use a .mjs file, which is late in the order of auto-discovered files.

Ok thanks for the added context, I didn't pick up what "golden path" meant in the first place and focused on the prefixed maintenance branch issue.

However, I still do not understand how you provide your `semantic-release` configuration file to your users. Are we talking about a library that handles `semantic-release` installation and/or configuration to enforce sensible defaults?

I'm trying to understand the exact use case here so we can make sure there isn't any existing/better solution to your issue. The more information you can provide on the exact use case the better :wink:

jebeaudet:
Hey!

Thanks for your questions and inputs! I work with @jonapich so I'll fill in for our use case of golden path.

Basically, we're working on implementing a secure and standard CI solution for our users internally. To do so, we have this setup : 
### `golden-ci` repository
This repository contains shared workflows and actions that are tailored to what you're trying to deploy (a python service, a java lib, etc.). Each workflow has the same general steps, adapted per language and build tools we decided "golden" : 
1. checkout the code
2. build & run tests
3. Fetch the secure artifactory token
4. Install semantic-release
5. Put the "golden" semantic-release configuration file
6. With semantic-release, determine the next version
7. Publish the artifacts to a artifact repository (github packages, jfrog, etc.)

With this, there is 2 problems we have by not being able to specify the configuration file : 
1. If the repo has, for some reason, a `.releaserc.json` file commited, it will have precedance over our `release.config.mjs`. This is a poor user experience, but it's not the end of the world.
2. The main problem we want to avoid here is that if we can't control the configuration file of semantic-release, a malicious actor to be able to commit a `package.json` file that will have a `release` target with `echo $ARTIFACTORY_SECURE_TOKEN` which would get the secure token printed in the logs.

Does that explain our use case better?

Thanks!

sheerlox:
Thank you for taking the time to describe your use case in detail, that makes a lot of sense.

It indeed looks like there's no practical way of going around this issue without this new option, so I'm rooting to merge this :heavy_check_mark:

sheerlox (approved):

babblebey:
> should we wait for some more feedback from @semantic-release/maintainers?

Yea, would be nice!

> I've got a follow observation/questions though

Now, I guess I can proceed to this with @jonapich..... I'd assume that the inline CLI argument still overrides the properties in the config file from the custom path - _straight drop-in_ - if so can we get a test coverage for this... it'd be great to have 😉

jonapich:
Sure! I'll give it a shot.

travi:
> Thank you for taking the time to describe your use case in detail

i agree. thank you, this information is really helpful. a suggestion for the future, please make sure this is clear to us and the maintainers of the project agree to the path forward _before_ creating a PR. a PR is really helpful to getting a feature out the door once it has been agreed to, but we want to help you avoid wasted effort if we arent aligned. 

> * If the repo has, for some reason, a `.releaserc.json` file commited, it will have precedance over our `release.config.mjs`. This is a poor user experience, but it's not the end of the world.
> * The main problem we want to avoid here is that if we can't control the configuration file of semantic-release, a malicious actor to be able to commit a `package.json` file that will have a `release` target with `echo $ARTIFACTORY_SECURE_TOKEN` which would get the secure token printed in the logs

i'm honestly having a hard time with this. i'll start with a reminder that you can find pretty consistently throughout our issues. semantic-release is an opinionated tool for simplifying releases that adhere to semantic versioning and not a general purpose workflow tool, so there are some practices that we tend to encourage and others that we tend to discourage. i'm still processing a bit, so i'm not completely sure where i think this lands, but i'll try to share thoughts that i'm trying to work through.

> 1. If the repo has, for some reason, a `.releaserc.json` file commited, it will have precedance over our `release.config.mjs`. This is a poor user experience, but it's not the end of the world.

this is a precedence concern, both tactically as you describe, but also philosophically around what semantic-release should support and encourage. semantic-release, and cosmiconfig more generically, are not intended for use where multiple configs exist. the primary reason cosmiconfig has a precedence order is for performance so that it can check for the most likely names and stop before the less common names if something is found. that order is not for enabling overrides or similar. unless i overlooked it, i dont find a mention of where the `release.config.mjs` is coming from, but it sounds like it is from somewhere outside of the repository. when a config exists in the repository, it could be very confusing why that config is being ignored in favor of a config from somewhere else. at minimum, i think we'd need to warn and maybe even fail if multiple configs are found, which i dont think would be simple through cosmiconfig.

from this perspective, at best it seems like you are wanting semantic-release to help you overcome debt of outdated configs. i would consider that sort of debt outside of the scope of responsibility of a tool like semantic-release. instead, i would encourage using a tool like [octoherd](https://github.com/octoherd), to script cleaning up such debt across your list of repos. the other potential desire that this sounds like is a desire to control these repos from a top-down approach, restricting the flexibility of the maintainers of these projects. we believe in configuration being close to the projects that use it and versioned together with the project over time, which is a big part of why semantic-release uses cosmiconfig in the first place. rather than pushing the config onto projects top-down, we'd instead suggest packaging up the config so that is easy to consume as a dependency, whether direct or installed during the pipeline execution, from the project and extending it through existing means.

> 2\. The main problem we want to avoid here is that if we can't control the configuration file of semantic-release, a malicious actor to be able to commit a `package.json` file that will have a `release` target with `echo $ARTIFACTORY_SECURE_TOKEN` which would get the secure token printed in the logs.

this feels even more outside the scope of responsibility of a tool like semantic-release. why do you have scenarios where malicious actors have write access to the repos that would contain these configs? i understand that echoing of secrets is just provided as an example, but many modern ci services will automatically obfuscate attempts to do so to known secrets. more generically though, if a malicious actor has write access to perfom such an attack through a release config, they also have access to many other configs within the repo. this seems like an access control or code review problem rather than a need for config override from semantic-release.

> We really want to skip any user-provided configuration and use our own.

while i understand the use case, i dont agree with the solution in this case. i'm not even saying that i dont think a `--config` or `--config-file` option will never make sense, but i'm currently not sold on this use case being the reason to add it and the ongoing maintenance costs that could come with it. i think it would be a much better investment to clean up your existing config debt, possibly in an automated fashion, and close what sound like access control gaps than to add this complexity to semantic-release

jonapich:
Hi! Thanks for the time you spent on this issue.

Unfortunately, @travi, there's a misunderstanding at the core of your reply. The technical debt we're addressing today is _not_ related to `semantic-release`. On the contrary, we were expecting to use `semantic-release` everywhere as a mean to enforce the rules behind the versioning and publishing flow of our IPs, to increase velocity and reduce friction while augmenting our security posture. To be extra clear, no one is using `semantic-release` at the moment and we're not trying to override any existing setup. In fact, our initiative is totally "opt-in" so it's not like anyone's going to be surprised by a behavior change once they delete all their GitHub Workflows in favor of calling ours.

---
Some more context:
- We're a company with dozens of R&D teams working in hundreds of repositories, multiple languages.
- For the past 15+ years, each team owned 100% of their CI/CD, leading to ever-growing discrepancies for all topics involved (testing, versioning, publishing, deploying).
- This leads to varying approaches in terms of CI, CD and Security (protecting against supply chain attacks, protecting credentials, etc.)

The size and velocity of the company has reached a point where this model is difficult to audit and no longer sustainable. The direction that we are taking is to standardize the CI/CD flow, which we codenamed "Golden". The premise is that when you enroll into the Golden, you get tons of freebies that just work and that you don't have to maintain:

- Reviewing licenses and dependencies
- Building and storing artifacts
- Unit, Functional and Infrastructure testing
- Monitoring and Notifications (e.g.: test failures forwarded to a slack channel)
- Statistics (test results over time, coverage over time, integration into company's dev portal, etc)
- Opinionated Versioning (using `semantic-commits`)
- Opinionated Publishing (using `semantic-release`)

Just to be extra clear; we're not designing any 'general tooling' meant for the public. This is about enforcing company rules and standards so that we can all align and reduce the complexity of our internal code base.

So, instead of writing and maintaining several GitHub Workflows around the above, users are now able to call our "Golden" workflow(s) and everything just works as long as they followed some agreed-upon conventions around the tooling (e.g.: python must use poetry, java must use mvn, etc).

I am part of the CI/CD team and we're building these Golden Paths in order to remove the burden of CI/CD from the teams. We want them to focus more on developing features and less about maintaining or arguing about a versioning scheme or the tooling to use.

So far so good: dozens of repositories have enrolled in our workflows and they love it and we get a lot of praise! 😸 On our side, we love it because we can design and roll out improvements for everyone at the same time without having to create hundreds of PRs (the `octoherd` way, which we're trying to eliminate?). Also, since each repository's getting aligned in terms of tooling and versioning, new employees have an easier time getting around since every single repository of a given language is managed the same.

---
Of course, all of the above could be "convention-based". Except...

...there's also a "hidden agenda" in this initiative: **Security** and **Zero Trust**. Since there's an ever growing amount of cyber attacks, supply chain attacks and "spying/scamming" efforts in order to compromise our customers data, _we are simply removing the possibility for an employee to perform impactful changes_. If you need more context about this new security vector, please take a moment to get familiar with build provenance and OIDC:

- https://github.com/actions/attest-build-provenance
- https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/about-security-hardening-with-openid-connect
- https://docs.docker.com/build/metadata/attestations/slsa-provenance/
- https://blog.gitguardian.com/leaked-secrets-in-code-repositories/

Here's a quote from the last link which summarizes the problem well:

> Allan Alford - (...) if your API key or credential is exposed, that's it, an attacker has access to that system. Once they are authenticated, it can be extremely difficult to be able to distinguish between a valid user and a malicious user.

And you said:

> (...) close what sound like access control gaps

This is _exactly_ the point. Repository maintainers will _not_ have access to those credentials, so if they're impersonated, the hacker is mostly empty-handed.

With that context in mind, here's an example scenario that we want to protect ourselves against:

1. A malicious actor gains access to a developer's credentials and impersonates them.
2. The malicious actor clones a repository.
3. The malicious actor modifies an existing workflow in order to exfiltrate the content of a secret that is stored in an environment variable for the duration of the `semantic-release` action (typically, the artifactory write token)
5. The malicious actor creates a Pull Request, which triggers the action.
6. Secrets are exfiltrated (the malicious user is running a web service and pushing the environment variables over there using REST to bypass GitHub's password masking feature)
7. The malicious user uses the credentials to replace `v1.2.3` with their own compromised binary.
8. Developers slowly start pulling and using the compromised artifact.

As you see, it's not a matter of whether we trust our employees or not. Even though we fully trust our employees, we know that no one is totally safe from impersonation.

I totally understand that other blocking features exist (for instance, poetry would complain that the hash of `v1.2.3` changed compared to the locked version). However, if someone was to upgrade from `v1.2.2` that day, they would lock in the compromised version. As you can see, there are a million ways to compromise software.

The takeaway here is that `semantic-release` should not be responsible of implementing this security, but it has to offer a base where we can control the behavior. Since `semantic-release`'s primary goal is to release artifacts, you need to acknowledge the fact that you are in possession of critical credentials, unlike an opinionated unit test, linter or versioning tool which are not usually in possession of such creds. Since your configuration allows invoking the shell with arbitrary code, it is in fact "dangerous" to rely on autodetection of configuration files.


---
Hopefully now you understand our position a little better. I will follow up with some targeted replies which unfortunately often contrasts with your philosophy:


---
> a suggestion for the future, please make sure this is clear to us and the maintainers of the project agree to the path forward before creating a PR.

To be fair, the issue was opened since 2020 so I figured it was wanted / up for grabs. Anyway, this is basically a one-liner, it was easier to just push it and show you, no worries. The work I'll have to do if this PR isn't merged is a lot more involved.


---
> (...) semantic-release is an opinionated tool for simplifying releases that adhere to semantic versioning and not a general purpose workflow tool (...)
> (...) semantic-release, and cosmiconfig more generically, are not intended for use where multiple configs exist. 
> (...) the primary reason cosmiconfig has a precedence order is for performance (...)
> (...) it can check for the most likely names and stop before the less common names if something is found. 
> (...) that order is not for enabling overrides or similar.


We love opinionated tools, and we definitely love `semantic-release`. We think that these tools improve the UX for our developers since they take decisions that we used to argue about.

As mentioned above, we're not looking to override or augment someone's configuration, rather, we're creating the company's opinionated way of testing and publishing libs and services, securing credentials at the same time. We found that `semantic-release` was the perfect tool for this job (kudos on the plugin system, we can manage any language!).

However, the fact that **the user can override us** by adding a `release` key in their `package.json` is a big concern. In a sense, this is similar to allowing **shell injection** into your software, byways of cosmiconfig, without any way to prevent that from happening. It is indeed possible to override the configuration by providing one higher in the auto-detected list, even if you claim that "it doesn't support overrides". The fact is that you _can_ override if you know the mechanics, just not individual attributes. If you really want to make it impossible to override a configuration, let's implement the `--config` switch (we can rename or revisit how it works) so that you can confidently say that you can't override the configs if you're specifying which file to use.

For reference on a different implementation, [buildx bake](https://docs.docker.com/build/bake/reference/#file-format) also accepts multiple filenames, however the values are sometimes merged sometimes overwritten. But the most important part is that they also provide a switch to select which configuration to use, _which disables searching for other configurations_. I strongly believe that `semantic-release` must offer the possibility of sticking to a specific config filename if that's the will of the user. This will reinforce how to secure the config file used. I don't think `semantic-release` is getting any benefit from not allowing any kind of control here, I strongly believe that it is actually exposing itself to vulnerabilities and exploits this way.

---
> unless i overlooked it, i dont find a mention of where the release.config.mjs is coming from.
> when a config exists in the repository, it could be very confusing why that config is being ignored in favor of a config from somewhere else.

The file's origin was explained in the comment I linked: https://github.com/semantic-release/semantic-release/issues/1592#issuecomment-2471211029

The config is located in the `golden-cicd` repository that users must call from their workflow:

1. User works in repo A.
2. User creates `.github/workflows/cicd.yaml`
3. User adds `uses: github-org/golden-cicd/.github/workflows/python.yaml`

That's the only thing the user has to do. When triggered in a PR or in main, our golden python workflow:
1. Clones repo A
2. Copies `golden-cicd/release.config.mjs` to the repo's root
3. Obtains an OIDC token, which is then used to authenticate to our secret vault and inject the artifactory's write token to the env vars
4. Launches `semantic-release` to publish the artifact
5. Removes the secret from the env vars

As you see, the secret is only available during `semantic-release`, hence why we really need to ensure that the user cannot run any self-authored code in there. We design the workflows so that whenever a "sandbox" environment is present (e.g.: unit tests), there are no secrets available. Without going into details, these security features are related to the provenance of the request and are handled inside OIDC cloud policies which contain conditions, such as the branch you're coming from and which workflow you're using.

This isn't a "shared configuration" which we want to `--extends` from. In fact, none of these repos are allowed to provide or customize the semantic release config.

---
> at minimum, i think we'd need to warn and maybe even fail if multiple configs are found, which i dont think would be simple through cosmiconfig.

That's basically the title of my user story! 😆 I agree that failing would be the way to go, but as a compromise, we could bake in a different log when `--config` is used, stating that autodetection is disabled in favor of what was specified at the CLI level.

It's my plan B if I cannot get this PR merged. I will add a script that will run after cloning the source repo, and fail the job is a release config is found. It's easy to do, but I'll also have to parse the `package.json` file and check for the `release` key since that one cannot be deducted from the filename alone.


---
> using a tool like [octoherd](https://github.com/octoherd) to script cleaning

With several hundred repositories to manage, this solution simply doesn't scale. That's why we're reversing the approach.


---
> a desire to control these repos from a top-down approach, restricting the flexibility of the maintainers of these projects
> we believe in configuration being close to the projects that use it and versioned together with the project over time

I agree with you from an open-source code perspective, however the reality of corporate is something else. Our repo maintainers don't want to deal with versioning and publishing anymore. Our company doesn't want multiple repositories to have different ways of doing the same things anymore. Our repo maintainers cannot access secrets anymore, unless they opt into the Golden paths. The Golden Path limits the impacts that could occur if a repo maintainer was impersonated, making their progress more difficult since they need to "hack out" of the workflows we control. Unfortunately, it was identified that `semantic-release` allows this "hack out" by simply providing a `package.json` file with the malicious code.


---
> rather than pushing the config onto projects top-down, we'd instead suggest packaging up the config so that is easy to consume as a dependency, whether direct or installed during the pipeline execution, from the project and extending it through existing means.

In our model, the users cannot push to artifact storage from their workstation; they have to go through an action. There's no need for the configuration to be consumed from the source repository. Our model is that the golden repository consumes the source repositories. The other way doesn't scale and introduces useless discrepancies between repositories.


---

I hope this puts to light the current vulnerability that `semantic-release` is exposing itself to for unaware users. Security talks are always a bit heated, and it's easy to fall into the "if you're compromised you have bigger problems" conundrum and not do anything about it. In reality, every little step taken toward a "zero trust" and "provenance-attested" model shuts down potential exploits and makes it harder to access privileged systems.

I totally understand that this tool wasn't designed with this approach in mind,

travi:
i appreciate the thorough explanation of your goals. i dont have time at the moment to respond to all of the points that i have suggestions about, but will try to follow up further when i can find more time.

> Unfortunately, it was identified that `semantic-release` allows this "hack out" by simply providing a `package.json` file with the malicious code.

i need to share a detail for you to consider about this. i think you are overlooking an important aspect of how semantic-release works. i was considering responding with a question related to whether you are reaching out to all of your tooling providers, like eslint that use similar configuration approaches, but you've sort of covered that with your scenario highlighting the scope of when the registry token is made available to the pipeline. however, even when limited to this degree, you are overlooking an important part of how semantic-release works. semantic-release only _coordinates_ the steps of the release process, it does not do them itself. when publishing to an npm registry, for example, the npm cli is leveraged internally to do the publish. the key detail to that is that npm exposes the ability to use hooks based on the publish event. `prepack` and `prepublishOnly` are very common hooks to use for build steps before publishing. some even use hooks to run verification before publishing (bringing tools like eslint and test frameworks back in scope). 

my point here is that npm itself enables running arbitrary commands, so even if we were to take steps like you suggest in semantic-release, i could not claim that we are able to prevent a maintainer of the project repository from adding additional commands through the repo in ways that are intentionally provided by the tools involved.

> the reality of corporate is something else

i recognize that the expectations of corporate contexts can have different requirements. i am in a role that is very similar to what you describe, but with a scope of thousands of orgs and countless repos. i'm familiar with the trade-offs you highlight and my suggestions come from successful experiences, but i recognize that the choices being made in your context might not align with alternatives that i suggest. oidc, slsa, build provenance, etc are on the radar of lots of corporate teams and you can see that we leverage many of those within this oss project as well

> Sorry about this; to be fair, the issue was opened since 2020 so I figured it was wanted / up for grabs. Anyway, I checked the code and realized this was going to end up being a one-liner, it was easier to just push it and show. I had seen the warning in the `contributing` doc, but I decided to go ahead and take a chance.

we encourage this to save the effort that can go into putting a PR together. as long as you are ok with the effort and wont be upset with us if we end up not accepting it, there is no harm. the issue remains open because we are open to being convinced of the feature. in many of our open issues, we are unable to share our full thoughts because we are a small volunteer team that cannot realistically keep up with all of the requests from our large community. 

i'm still not saying that a `--config` or `--config-file` should never be added, but i remain unconvinced that this is the reason to do it, even without getting into all the additional effort that would need to go into making the difference with the existing `--extends` option.

PR #3595, opened 2025-02-08
fix(get-git-auth-url.js): Added debug for success

MrGadget1024:
Provides a timestamped log entry for this process so it's clear how long this took to complete.


travi (approved):
thanks!

PR #3610, opened 2025-02-15
Fixed typescript definition

### LINKED ISSUE #3601: How to determine if in dryRun from plugin code
I'm working on a plugin in which I want to do something conditionally in the verify step, iff we're in `dryRun`.

Based on the types definition, it looks like `Options` [has the `dryRun` attribute I'd need](https://github.com/semantic-release/semantic-release/blob/f15019c19e6b934e3dea8458b5808f2108f38705/index.d.ts#L663). And based on the implementation of `run`, it looks like the [context passed to verify](https://github.com/semantic-release/semantic-release/blob/f15019c19e6b934e3dea8458b5808f2108f38705/index.js#L106) has [an options attribute](https://github.com/semantic-release/semantic-release/blob/f15019c19e6b934e3dea8458b5808f2108f38705/index.js#L39). However, [none](https://github.com/semantic-release/semantic-release/blob/f15019c19e6b934e3dea8458b5808f2108f38705/index.d.ts#L110) [of](https://github.com/semantic-release/semantic-release/blob/f15019c19e6b934e3dea8458b5808f2108f38705/index.d.ts#L90) [the](https://github.com/semantic-release/semantic-release/blob/f15019c19e6b934e3dea8458b5808f2108f38705/index.d.ts#L49) [context](https://github.com/semantic-release/semantic-release/blob/f15019c19e6b934e3dea8458b5808f2108f38705/index.d.ts#L27) interfaces define that `options` attribute, which I assume is why my editor doesn't like me trying to use it in type script...

Do I have all that right? Is that a bug in the type definition? Or an intentional omission in the public API? What is the recommended way for a plugin to know if it's being executed in `dryRun` mode?

seebeen:
BaseContext was missing options key with the `Options` object
Fixes #3601

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -41,6 +41,11 @@ declare module "semantic-release" {
      * Has error, log, warn and success methods.
      */
     logger: Signale<"error" | "log" | "success" | "warn">;
+
+    /**
+     * Semantic release configuration
+     */
+    options: Options;
   }
 
   /**
@@ -667,6 +672,11 @@ declare module "semantic-release" {
      * This allows for making releases from a local machine.
      */
     ci?: boolean | undefined;
+
+    /**
+     * Set to true when the `ci` option is set to false.
+     */
+    noCi?: true;
 
     /**
      * Any other options supported by plugins.
--- a/package-lock.json
+++ b/package-lock.json
@@ -43,6 +43,8 @@
         "semantic-release": "bin/semantic-release.js"
       },
       "devDependencies": {
+        "@types/node": "^22.13.4",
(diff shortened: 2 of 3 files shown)


seebeen:
@travi I've added two packages to devDependencies. `@types/node` and `@types/signale` in order to validate the type definition. They are not a must-have (since they can be installed globally on the dev workstation). So if you'd like me to remove them - please let me know.

babblebey (approved):
LGTM 🚀

travi (approved):

travi:
thanks for the quick resolution on this @seebeen!

PR #3650, opened 2025-03-21
doc: default for maintenance branches doesn't match implementation

samuelcadieux:


samuelcadieux:
Nvm, I misread the docs semantic-release uses `micromatch`, not regexes.

I saw a regex in a unit test and that mislead me.

PR #3727, opened 2025-05-18
docs(issue-template): more directly ask for improved detail for bug reports

travi:


.github/ISSUE_TEMPLATE/01_bug_report.yml
```diff
@@ -19,7 +19,10 @@ body:
   - type: input
     attributes:
       label: "`semantic-release` version"
-      description: Version set in `package.json` `devDpendencies`.
+      description: |
+        The version executed in your pipeline, printed in the first line of output.
```
travi, line 23: we dont encourage defining semantic-release as a dependency, so looking in the log output is a better source to suggest finding this detail than `devDependencies`

.github/ISSUE_TEMPLATE/01_bug_report.yml
```diff
@@ -19,7 +19,10 @@ body:
...
+        If not already using the latest version, please try upgrading to the latest version and refreshing your lockfile to see if the issue persists.
```
travi, line 24: we should more officially document our willingness to support old versions, but this is a start. we also get lots of support requests that could have been resolved by refreshing a lockfile when folks do decide to depend directly on semantic-release (whether by choice or because our recommendation isnt visible enough, or is inconsistent in our docs)

.github/ISSUE_TEMPLATE/01_bug_report.yml
```diff
@@ -40,13 +43,17 @@ body:
   - type: textarea
     attributes:
       label: "`semantic-release` configuration"
-      description: Link to your repository or relevant part of the `semantic-release` config.
+      description: |
+        Link to the configuration in your repository or relevant part of the `semantic-release` config.
+        When including configuration directly, please format with triple backticks (```) to ensure it is readable.
```
travi, line 48: we get lots of issues opened that do not format config or log output, so making the request more obvious. 

there is the [`render` option](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema#attributes-for-textarea) that could do it automatically, but that would require splitting this field to separate fields for link vs directly sharing. maybe we do that at some point, but that feels like a future step

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,6 @@
+contact_links:
+  - name: Usage Support
```
travi, line 2: i've always preferred to drive config and usage questions out of issues and into discussions, but our options havent really suggested that. this adds a link to the new issue step to at least present the option

potential downside: discussions are freeform, so we provide less guidance on providing relevant details. maybe https://github.blog/news-insights/product-news/github-discussions-just-got-better-with-category-forms/ is worth considering as a follow-up

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,6 @@
...
+    url: https://github.com/semantic-release/semantic-release/discussions
+    about: |
+      For usage and configuration support, please leverage our community discussions.
```
travi, line 5: i'm wanting to encourage these topics toward discussions. please feel free to suggest improved wording to convey this.
* "how do i...?"
* "my config is this, but it isnt doing what i expect"
* "semantic-release isnt doing what i expect, so i think it is broken (but a maintainer hasnt confirmed whether that is intentional, misconfiguration, or something else)"

gr2m (approved):
fantastic improvements!

.github/ISSUE_TEMPLATE/01_bug_report.yml
babblebey: Totally agree with this... I go as far as editing issue description sometimes

babblebey (approved):
Love it 🫶🏾

.github/ISSUE_TEMPLATE/01_bug_report.yml
travi: same :) 

even if I don't have time or the mental space to respond, that's one step I'll take when reading through initially

Issue #1729, opened 2021-01-11
Create option to disable verifyAuth or allow plugins to overwrite verifyAuth


secustor:
## New feature motivation

GitlabCI tokens have no permissions to push code, therefore semnatic-release users have to use personal tokens which is not ideal from a security and maintenance standpoint.      
https://github.com/semantic-release/gitlab/issues/156

## New feature description

To enable the usage of CI tokens we need a possibility to prevent semnatic-release to fail on the `verifyAuth` stage, an alternative could be to allow plugins to overwrite the `verifyAuth` stage. 

## New feature implementation

If the first idea is implemented it could be in form of flags such as `--skip-verifyAuth` or better `--skip-stage=verifyAuth` .  

The second concept needs new implementation in the core and the Gitlab plugin, but seems for me to be the cleaner variant. 
Basically core will provide a default implementation for `verifyAuth` and allows plugins to overwrite this default. 

I'm willing give this a shot to implement this myself.

travi:
as mentioned in https://github.com/semantic-release/gitlab/issues/156#issuecomment-746562629, semantic-release needs the ability to push tags back to the repository. without authorization to do this, the core of semantic-release cannot determine the proper commit that is associated with the previous version. since auth is necessary, i don't understand how skipping verification would be helpful here.

secustor:
Thanks for the fast response!

I have looked a little bit deeper into the code base and see now the problem, but I'm not sure what do you mean that sematic needs to push to get the next release. From what I have seen it pushes only tags and their notes after sematic has decided to what the next release is. 

To better clarify my use beforehand, basically I want to use the `CI_JOB_TOKEN` from Gitlab which has no right to push code but can create Releases. This releases can be created on existing tags or implicitly generate one using a git reference.      
So basically I want to create the tags with this implicit method instead of pushing the tag using git.     

The solution could be to move the push logic into a plugin phase and provide a default as I have described in the first post for `verifyAuth`. Finally the logic can be then overwritten by the Gitlab plugin to create a release using the [Release API](https://docs.gitlab.com/ee/api/releases/#create-a-release)

travi:
> From what I have seen it pushes only tags and their notes after sematic has decided to what the next release is.

Correct, but the tag that is pushed by semantic-release is used for the _next_ release for determining the previous version to apply the next determined version bump to. The intention of semantic-release is to handle the whole release process and there are several discussions about why this behavior exists in core. It is very unlikely that we would change this behavior.

[...]

PR #3760, opened 2025-06-06
Add feature to skip git push operations

arvest-bjoneson:
This PR adds a configuration property, "skipPush", which skips git push operations (and access checks), but otherwise executes the lifecycle as normal.

Why would anyone want to do this?

My primary motivation is the ability to use short lived job tokens provided by Gitlab CI, which *do* have permissions to create repository tags, but lack permissions to push to the git remote. There is a feature flagged capability on GitLab, to allow git push permissions to job tokens: https://docs.gitlab.com/ci/jobs/ci_job_token/#allow-git-push-requests-to-your-project-repository. However... in larger enterprises and highly regulated industries, there can often be trepidation in enabling these types of features.

This has been discussed extensively here, it appears: https://github.com/semantic-release/semantic-release/issues/1729

I did consider a more significant refactor, which could potentially fully delegate the git tag creation (and verification steps) to a plugin, by introducing the git tag creation step as a plugin lifecycle phase. However, this would be a significant architectural change, difficult to do in non-breaking way, and potentially confusing for users.

In the end, it seemed a simple, well documented config flag would suffice. So users who wish to utilize this feature, are explicitly utilizing it.

There are likely other edge scenarios that this feature would be useful for. Perhaps someone wants to effectively do a "dry run", but still exercise the publish steps of various plugins. Ultimately the goal here is to provide flexibility to the user, while still maintaining the core constructs of the architecture.

I've opened a pull request for the gitlab plugin as well, which addresses the ability for that plug-in to properly accommodate the use of job tokens: https://github.com/semantic-release/gitlab/pull/859

There does not appear to be any regression of any existing functionality.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -36,6 +36,7 @@ Usage:
     .option("fail", { ...stringList, group: "Plugins" })
     .option("debug", { describe: "Output debugging information", type: "boolean", group: "Options" })
     .option("d", { alias: "dry-run", describe: "Skip publishing", type: "boolean", group: "Options" })
+    .option("s", { alias: "skip-push", describe: "Skip git push operations", type: "boolean", group: "Options" })
     .option("h", { alias: "help", group: "Options" })
     .strict(false)
     .exitProcess(false);
--- a/docs/usage/ci-configuration.md
+++ b/docs/usage/ci-configuration.md
@@ -20,7 +20,7 @@ See [CI configuration recipes](../recipes/ci-configurations/README.md) for more 
 
 ### Push access to the remote repository
 
-**semantic-release** requires push access to the project Git repository in order to create [Git tags](https://git-scm.com/book/en/v2/Git-Basics-Tagging). The Git authentication can be set with one of the following environment variables:
+**semantic-release** requires push access to the project Git repository (unless the [skipPush](configuration.md#skippush) option is enabled) in order to create [Git tags](https://git-scm.com/book/en/v2/Git-Basics-Tagging). The Git authentication can be set with one of the following environment variables:
 
 | Variable                                              | Description                                                                                                                                                                                                                  |
 | ----------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
--- a/docs/usage/configuration.md
+++ b/docs/usage/configuration.md
@@ -150,6 +150,18 @@ The objective of the dry-run mode is to get a preview of the pending release. Dr
 
 **Note**: The Dry-run mode verifies the repository push permission, even though nothing will be pushed. The verification is done to help user to figure out potential configuration issues.
 
+### skipPush
+
+Type: `Boolean`<br>
+Default: `false`<br>
+CLI arguments: `-s`, `--skip-push`
+
+Set to `true` to skip the push of the release tag to the remote repository via git, and associated push permission checks.
+This allows for the release tag to be published via other means where repository push permissions may not be required
+(i.e. GitLab Releases)
+
+**Note**: Some plug-ins (i.e. GitHub) may require that the tag exists in the remote repository before publishing. Use with caution.
+
 ### ci
 
 Type: `Boolean`<br>
--- a/docs/usage/configuration.md
+++ b/docs/usage/configuration.md
@@ -160,7 +160,10 @@ Set to `true` to skip the push of the release tag to the remote repository via g
 This allows for the release tag to be published via other means where repository push permissions may not be required
 (i.e. GitLab Releases)
 
-**Note**: Some plug-ins (i.e. GitHub) may require that the tag exists in the remote repository before publishing. Use with caution.
+Repository pull permissions are still required.
+
+**Note**: Some plug-ins (i.e. GitHub) may require that the tag exists in the remote repository before publishing.
+Use with caution.
 
 ### ci
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -668,6 +668,12 @@ declare module "semantic-release" {
     dryRun?: boolean | undefined;
 
     /**
+     * Skip pushing release tags and notes to the remote repository
+     * This allows plug-ins to manage release tags (i.e. GitLab releases)
+     */
+    skipPush?: boolean | undefined;
+
+    /**
      * Set to false to skip Continuous Integration environment verifications.
      * This allows for making releases from a local machine.
      */
--- a/index.js
+++ b/index.js
@@ -20,6 +20,7 @@ import {
   addNote,
   getGitHead,
   getTagHead,
+  isBranchInSync,
   isBranchUpToDate,
   push,
   pushNotes,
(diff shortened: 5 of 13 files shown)


travi:
> This has been discussed extensively here, it appears: #1729

please be sure to actually read the discussion, especially https://github.com/semantic-release/semantic-release/issues/1729#issuecomment-1878168159. we are not interested in making changes to how tags are added because of gitlab choosing to prevent git operations. 

also, a reminder that [we recommend avoiding commits during your release](https://semantic-release.gitbook.io/semantic-release/support/faq#why-is-the-package.jsons-version-not-updated-in-my-repository), so git operations beyond pushing tags can be avoided.

arvest-bjoneson:
I completely agree about the recommendation of not pushing to the repository beyond tags.

What I'm proposing here is a way for people to make use of this project in a way that is compatible in an out of the box way with a very large user base in GitLab, that doesn't want to manage long lived tokens, or enable repository push permissions.

I believe this solves the problem simply by providing *flexibility* to the user, without changing any of the existing architecture.

I'm trying to really hard to understand the *harm* in simply providing a clear, well documented feature that allows for more flexible workflows.

I guess my primary question is... we haven't closed #1729. One of the statements made, was "we're not saying it will never happen".

Can you provide any explicit feedback on how to improve this implementation in a way which accomplishes the original goal of the issue?

If not... I would suggest 1729 just be closed out with the declaration that there is no desire to support this functionality in the project.

Gitlab has provided a feature flagged function to provide repository push capability to the job token. And I agree that if you enable that, this becomes unnecessary in that use case. That said... I think there are likely other scenarios where a user may wish to simply skip git push operations while executing the release lifecycle.

As far as the base semantic-release project is concerned, this isn't a "delegate the tag creation to plugins" rearchitecture, but simply a flag to optionally disable a step in the workflow.

I'm willing to do the leg work. I'm a huge fan, proponent and user of this project.

travi:
> I'm trying to really hard to understand the _harm_ in simply providing a clear, well documented feature that allows for more flexible workflows.

the harm is the extra complexity that would need to be maintained by our volunteer maintainers to work around the short comings of a single git host. we already have a very complicated set of options and we are trying very hard to avoid adding additional config options when a default should be able to work across all options. as clarified in the comment that i linked directly to above, our core depends on git. tags must exist, so git is the common denominator that should exist in all scenarios, but gitlab made choices to complicate that. some of those complications are supposed to be resolved, but i'll admit that i havent fully kept up on that progress. i dont understand the purpose of this proposal based on what i thought i understood would be happening on the gitlab side. copying @fgreinacher for more insights specific to the progress from the gitlab side of things

> I believe this solves the problem simply by providing _flexibility_ to the user, without changing any of the existing architecture.

flexibility adds more code paths for us to maintain and ensure continue to work _forever_. it also adds confusion for our users, which results in more support requests for our maintainers to respond to and work through. we are trying hard to avoid flexibility. instead, we want a common core that our plugins can build on to differenciate as needed for various platforms. again, core needs tags. that is not something we are open to moving out to have plugins manage directly. we might be open to changes that account for needs we havent accounted for if they can be done in a common way in core without the need for variation for specific platforms or the addtion of additional flags that our users need to sort out. 

> Can you provide any explicit feedback on how to improve this implementation in a way which accomplishes the original goal of the issue?

in the future, i suggest starting here. one of the hardest things for us to do as maintainers is to reject a PR that someone has invested time and effort into, which often results in upsetting a contributor. however, we cannot accept proposals that do not align with the goals of the project simply because of that time and effort investment. we do appreciate your investment here. there might be a path forward, but that decision needs to be made before suggesting a code change. if we want to find a path toward a change related to this, i suggest opening a discussion thread, relating it to the issue thread and this PR and we can talk through options before investing more in actual changes

> If not... I would suggest 1729 just be closed out with the declaration that there is no desire to support this functionality in the project.

i'm confused by this interpretation. we are very interested in semantic-release working with gitlab. it does work with gitlab. we are also interested in enabling the use of the more narrowly scoped tokens, but see that as outstanding work from the gitlab side, not the semantic-release side. the issue is not closed so discussions can continue related to that

arvest-bjoneson:
All good. Understood. I disagree that this particular feature adds any overt complexity, but It's a rational perspective.

In this case the feature is not "to allow gitlab to publish without push access", but simply "to optionally skip push operations". Through that lens, I feel like this is a very straightforward and maintainable feature. It's simply a flag that allows the user to skip a step in the lifecycle (at their specific instruction).

My statement on 1729 is that the ask was to be able to allow plugins to override git tag creation. (That's the title of the issue). I think if there is not plan to support that ability, then it would make sense to close that issue, and instead focus on the enablement of the job token specifically in the gitlab plugin project.

I appreciate you taking the time to dialog. I knew going in the likelihood of this being accepted was marginal based on the conversation, so no worries on the time spent. Oftentimes, for me it is much easier to just "see what an implementation would look like", as opposed to talking about it in theory. Nothing provides more clarity of intent, than code that demonstrates that intent. I learned a lot about the inner workings of the project in the process, and totally understand it's folks like you that ultimately have to own the ramifications of any code introduced.

travi:
I appreciate your understanding. Certainly no harm to think through options through code as long as you are open to the experiment not ending up being an accepted solution in the end. We get plenty of proposals where a contributor can be fairly demanding that their contribution be accepted and quickly that we end up having to set a more realistic expectation and end up disappointing folks as a result.

> In this case the feature is not "to allow gitlab to publish without push access", but simply "to optionally skip push operations". Through that lens, I feel like this is a very straightforward and maintainable feature. It's simply a flag that allows the user to skip a step in the lifecycle (at their specific instruction).

adding a flag often seems like a simple path toward more flexibility, but we really have to consider ramifications of each that we consider adding. since tags are so core to the functionality of semantic-release, we dont want to leave that behavior up to making sure other details are aligned just right. we would be asking for lots of support requests about semantic-release not remembering the previous version if the tags arent getting pushed. semantic-release currently works without including either the github or gitlab plugins and we intentionally want to keep it that way

> My statement on 1729 is that the ask was to be able to allow plugins to override git tag creation. (That's the title of the issue). I think if there is not plan to support that ability, then it would make sense to close that issue, and instead focus on the enablement of the job token specifically in the gitlab plugin project.

this is fair feedback. we have a hard time keeping up with all of our issues, so we dont always get around to reframing issues when they diverge from the original request. i'll give some consideration if we should rename that issue since the conversation has gone in a different direction, or if we should close and capture the ongoing conversation around gitlab tokens in a different discussion

arvest-bjoneson:
Totally understand the sentiment on "demanding" engineers. We can often have a propensity to believe our idea is the greatest thing ever, and should just be quickly accepted and intuitively obvious to anyone. :)

> we dont want to leave that behavior up to making sure other details are aligned just right. we would be asking for lots of support requests about semantic-release not remembering the previous version if the tags arent getting pushed. semantic-release currently works without including either the github or gitlab plugins and we intentionally want to keep it that way

The point around users not using github or gitlab plugins is admittedly not something I'd considered, and makes a lot of sense. That said.... I do feel like, when you have to *explicitly* state in the configuration that you are skipping git push operations, the answer to "why isn't my release creating a release tag"? should be hopefully, quite obvious. It would seem like asking "why didn't my release get published"? when running in dry run mode.

That said... users are not always the sharpest tools in the shed, and perhaps I'm underestimating the ability for a user to understand their own configuration, or read what may be right in front of them in the output.

travi:
the potentially confusing situation with the tag getting pushed is that it wouldnt cause a failure for the current version being published, but would for the next release because it could try to publish the same version again (depending on the level of version bump) because the previous tag wasnt pushed. we already get plenty of support questions about that situation, so we'd like to avoid making it more likely to make it happen

PR #3861, opened 2025-09-12
Update ci-configuration.md Gitlab URL about Tokens

torrespro:
https://docs.gitlab.com/ce/user/profile/personal_access_tokens/ is a 404 we need to remove the ce/ part and works -> https://docs.gitlab.com/user/profile/personal_access_tokens/


travi (approved):
thanks!

travi:
it looks like there is a formatting issue. could you run `npm run lint:prettier:fix` to fix that up?

PR #3905, opened 2025-10-16
promote to beta

travi:
for semantic-release/npm#958for semantic-release/npm#958


travi:
like https://github.com/semantic-release/npm/pull/1015#issuecomment-3408998374, i'm going to promote this to beta status since it is mostly working at production quality. we can do a proper round of reviews before promoting to stable

PR #3913, opened 2025-10-20
docs: update GitHub Actions documentation for Trusted Publisher

Th3S4mur41:
Clarified the setup process for Trusted Publisher on npmjs for GitHub Actions, emphasizing the importance of specifying the correct workflow file.

See https://github.com/semantic-release/npm/issues/958#issuecomment-3418617666

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,7 +29,7 @@ jobs:
       - run: corepack npm audit signatures
       # pinned version updated automatically by Renovate.
       # details at https://semantic-release.gitbook.io/semantic-release/usage/installation#global-installation
-      - run: npx semantic-release@24.2.9
+      - run: npx semantic-release@25.0.1
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.SEMANTIC_RELEASE_BOT_NPM_TOKEN }}
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -25,12 +25,12 @@ jobs:
           results_format: sarif
           publish_results: true
       - name: Upload artifact
-        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
         with:
           name: SARIF file
           path: results.sarif
           retention-days: 5
       - name: Upload to code-scanning
-        uses: github/codeql-action/upload-sarif@16140ae1a102900babc80a33c44059580f687047 # v4.30.9
+        uses: github/codeql-action/upload-sarif@4e94bd11f71e507f7f87df81788dff88d1dacbfb # v4.31.0
         with:
           sarif_file: results.sarif
--- a/docs/extending/plugins-list.md
+++ b/docs/extending/plugins-list.md
@@ -213,3 +213,11 @@
 - [sr-uv-plugin](https://github.com/Artessan-Devs/sr-uv-plugin)
   - `verifyConditions`: Ensures `pyproject.toml` exists and contains a `[project]` section.
   - `prepare`: Updates the `[project].version` field in `pyproject.toml` to match the release version.
+- [@jno21/semantic-release-github-commit](https://github.com/Jno21/semantic-release-github-commit)
+  - **Notes**: This plugin creates a commit on GitHub using the GitHub API, enabling signed commits via a GitHub App.
+  - `verifyConditions`: Verify GitHub authentication and configuration.
+  - `prepare`: Create a commit with the specified files using the GitHub API.
+- [semantic-release-minecraft](https://github.com/pynickle/semantic-release)
+  - `verifyConditions`: Verify that all needed configuration is present.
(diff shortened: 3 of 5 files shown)


Th3S4mur41:
@travi is this the kind of documentation change you had in mind?

docs/recipes/ci-configurations/github-actions.md
```diff
@@ -13,6 +13,11 @@ GitHub Actions is a [trusted identity provider](https://docs.npmjs.com/trusted-p
 The npm registry [recently increased restrictions for use of long-lived access tokens](https://github.blog/changelog/2025-09-29-strengthening-npm-security-important-changes-to-authentication-and-token-management/), further encouraging trusted publishing as the preferred approach for publishing to npm from GitHub Actions.
 Enabling trusted publishing requires granting the `id-token: write` permission to the job performing the publish step and [configuring a trust relationship](https://docs.npmjs.com/trusted-publishers#step-1-add-a-trusted-publisher-on-npmjscom) between your GitHub repository and npm.
 
+> [!NOTE]
```
travi: i'm not sure whether this will render as expected in the context of gitbook, like on https://semantic-release.gitbook.io/semantic-release/recipes/ci-configurations/github-actions#trusted-publishing-and-npm-provenance

it is probably safer to match simpler instances of calling out a note on this page than to use the github admonition, unless you could confirm that it would work in the gitbook context too

Th3S4mur41: Good hint @travi ... Indeed, Gitbook does not support GitHub Alerts, but has its own version of it:
```md
{% hint style="info" %}
**Info** — general tips or helpful context.
{% endhint %}
```

Would you prefer I switched to the Gitbook version or rather turn this into something that would work in both GitHub and Gitbook like e.g. a sub-chapter?

34j:
@travi @Th3S4mur41
I apologize for interrupting, but I believe the current workflow example does not work for Trusted Publishing.
After several attempts, I've found that using Trusted Publisher requires **BOTH** of the following:

- npm >= 11.5.1, which is **not included**  in `setup-node@v4 (lts)` (without manual update: `An error occurred while running semantic-release: ExecaError: Command failed with exit code 1: npm publish ... npm error code ENEEDAUTH` https://github.com/34j/create-minimal-package/actions/runs/18815189225/job/53682456467)
- semantic-release>=25, not 24 (with 24: `AggregateError:  SemanticReleaseError: No npm token specified.` https://github.com/34j/react-ag-psd-psdtool/actions/runs/18813783923/job/53679906899)

It would be greatly appreciated if
1. the above two requirements could be clearly stated in the documentation
2. add the npm version update to the sample workflow as described in the npm documentation. (https://docs.npmjs.com/trusted-publishers)

```yaml
      - name: Update npm
        run: npm install -g npm@latest
```

https://github.com/semantic-release/semantic-release/blob/b2d127191c02cf213b4b6c9316fa9b047e6dcea1/docs/recipes/ci-configurations/github-actions.md?plain=1#L55-L66

Also I believe this sentence is obsolete and should be fixed as well

https://github.com/semantic-release/semantic-release/blob/b2d127191c02cf213b4b6c9316fa9b047e6dcea1/docs/recipes/ci-configurations/github-actions.md?plain=1#L7

Th3S4mur41:
I'd rather suggest to update node to 24, which already includes the correct npm version... Especially considering that node 24 should become LTS any day now...

34j:
> I'd rather suggest to update node to 24, which already includes the correct npm version... Especially considering that node 24 should become LTS any day now...

I didn't know that; thank you for pointing it out. In any case, it would be helpful if something were documented.

travi:
We do need to update the semantic-release version, but there is no need to consider node or npm for the trusted publishing requirement. The npm plug-in for semantic-release has a direct dependency on npm specifically to ensure a supported version is used. We do not use the otherwise installed version from the environment for the release process

34j:
@travi

> has a direct dependency on npm specifically to ensure a supported version is used

This workflow used semantic-release 25.0.1, but trusted publishing failed due to too old npm.
https://github.com/34j/create-minimal-package/actions/runs/18815189225/job/53682456467
If recent version is installed, a warning stating something like “unless using OIDC...” should appear, but in the logs it does not appear. After adding "Update npm" it started to work correctly.

travi:
@34j What you are seeing is unrelated to this pr. Please open a separate issue. 

I see you have semantic-release and plugins under devDependencies and installing with pnpm, which might be complicating things. Please try removing from devDependencies and just use npx to both install and run

docs/recipes/ci-configurations/github-actions.md
travi: good find. it is helpful to know the difference so we can decide at a higher level how we should be marking these up. i think to change, we'd likely want to change strategies consistently in other areas too. for now, lets avoid the larger change and go with something simpler, https://github.com/semantic-release/semantic-release/blob/b2d127191c02cf213b4b6c9316fa9b047e6dcea1/docs/recipes/ci-configurations/github-actions.md#L24.

travi:
> Please try removing from devDependencies and just use npx to both install and run

i fully admit that our documentation is currently not clear on this advice either. i've opened https://github.com/semantic-release/semantic-release/pull/3920 to finally update that advice in our documentation, even though we've been consistent with that advice in support for a while now

Th3S4mur41:
@travi I've updated the not format as discussed and formatter with prettier

PR #3920, opened 2025-10-26
docs(installation): update recommendations to align with advice provided as support for some time

travi:
this update is long overdue and gives us a place to point to from support as well as hopefully guiding folks to better initial success

### DIFF
--- a/docs/usage/installation.md
+++ b/docs/usage/installation.md
@@ -1,44 +1,35 @@
 # Installation
 
-## Local installation
-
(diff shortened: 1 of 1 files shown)


34j:
@travi Sorry for asking so many questions, but do you mean that the plugins should be installed using npx as well like `npx --package=semantic-release --package=@semantic-release/git semantic-release`? The current plugins documentation seems to recommend installing plugins as devDependencies: 
https://github.com/semantic-release/semantic-release/blob/c6be2f2136a3d642c42f94ea7075b6939e9ac26f/docs/usage/plugins.md?plain=1#L40-L42

travi:
Yes, that documentation needs an update as well to align with this guidance

PR #3926, opened 2025-10-31
docs(social): removed the link to the social network Twitter

Olexandr88:


travi:
thanks for bringing this to our attention, @Olexandr88. in reality, we have not been active on Twitter for some time and no longer consider it a safe space for our community. rather than updating the branding, it would actually be better for us to remove the link entirely. would you be open to modifying your PR to take care of that for us?

travi (approved):
Thanks!

PR #3980, opened 2025-12-16
refactor(deps): remove deprecated semver-diff

### LINKED ISSUE #3887: Deprecated dependency: semantic-release still uses deprecated semver-diff package
### Current behavior

When running `npm test` with semantic-release@24.2.9, npm shows the following deprecation warning: 

```
npm WARN deprecated semver-diff@5.0.0: Deprecated as the semver package now supports this built-in.
```

The root cause is that semantic-release@24.2.9 continues to depend on the deprecated semver-diff package, which cannot be resolved by simply uninstalling semver-diff or upgrading semver locally.

### Expected behavior

semantic-release should remove the dependency on semver-diff and rely on the built-in functionality provided by the semver package. This would eliminate the warning and improve maintainability.

24.2.9

### CI environment

npm

### Plugins used

"@semantic-release/changelog": "^6.0.3",
    "@semantic-release/commit-analyzer": "^13.0.1",
    "@semantic-release/exec": "^7.1.0",
    "@semantic-release/git": "^10.0.1",
    "@semantic-release/github": "^11.0.6",
    "@semantic-release/release-notes-generator": "^14.1.0",

nd

### CI logs

npm warn deprecated semver-diff@5.0.0: Deprecated as the semver package now supports this built-in.

LuisUrrutia:
## Summary

- Replace deprecated `semver-diff` package with built-in `semver.diff()` functionality
- Add `getVersionDiff` helper function in `lib/utils.js` that preserves `semver-diff` behavior (returns `undefined` when v1 >= v2, indicating no upgrade path)
- Add unit tests for the new helper function

Fixes #3887

## Key difference

  The critical difference between `semver-diff` and `semver.diff` is how they handle cases where v1 >= v2:

  | Comparison | `semver-diff` | `semver.diff` |
  |------------|---------------|---------------|
  | `('1.0.0', '2.0.0')` | `'major'` | `'major'` |
  | `('2.0.0', '1.0.0')` | `undefined` | `'major'` |
  | `('1.1.0', '1.0.0')` | `undefined` | `'minor'` |
  | `('1.0.0', '1.0.0')` | `undefined` | `null` |

  `semver-diff` returns `undefined` when there's no upgrade path (v1 >= v2), while `semver.diff` returns the diff type regardless of direction.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1214,7 +1214,6 @@
       "resolved": "https://registry.npmjs.org/@octokit/core/-/core-7.0.6.tgz",
       "integrity": "sha512-DhGl4xMVFGVIyMwswXeyzdL4uXD5OGILGX5N8Y+f6W7LhC1Ze2poSNrkF/fedpVDHEEZ+PHFW0vL14I+mm8K3Q==",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@octokit/auth-token": "^6.0.0",
         "@octokit/graphql": "^9.0.3",
@@ -1987,7 +1986,6 @@
       "integrity": "sha512-ne4A0IpG3+2ETuREInjPNhUGis1SFjv1d5asp8MzEAGtOZeTeHVDOYqOgqfhvseqg/iXty2hjBf1zAOb7RNiNw==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "undici-types": "~7.16.0"
       }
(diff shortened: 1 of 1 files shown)


MikeMcC399:
@LuisUrrutia 

It's great to see this PR submitted! I wonder if maintainers will be able to review it and run the workflow soon?

I would have expected this to be `fix(deps):` rather than `refactor(deps):` since it removes a deprecation notice that is visible to users. I guess this is open to interpretation though.

travi:
> I would have expected this to be `fix(deps):` rather than `refactor(deps):` since it removes a deprecation notice that is visible to users

agreed. updated in the pending merge

PR #4024, opened 2026-01-29
docs: warn against using registry-url in setup-node

### LINKED ISSUE #3842: documentation: highlight that defining `registry` as part of `setup-node` can conflict with semantic-release behaviors
https://github.com/semantic-release/semantic-release/discussions/3837#discussioncomment-14164801 highlighted that we dont make this clear enough. we should at least add a warning to https://semantic-release.gitbook.io/semantic-release/recipes/ci-configurations/github-actions to help folks discover this when referencing our recipe

taronsung:
## Summary

Fixes #3842

Adds a warning section to the GitHub Actions documentation explaining that using the `registry-url` option in `actions/setup-node` can conflict with semantic-release's npm authentication mechanism, leading to `EINVALIDNPMTOKEN` errors even when the npm token is valid.

### Problem

As discussed in #3837, users who configure `registry-url` in their `setup-node` step may encounter npm authentication failures because:
1. `setup-node` creates an `.npmrc` file when `registry-url` is set
2. This `.npmrc` can conflict with semantic-release's npm authentication
3. The resulting error (`EINVALIDNPMTOKEN`) is misleading since the token itself is valid

### Solution

Added a dedicated warning section with:
- Clear explanation of why `registry-url` should be avoided
- Code examples showing incorrect vs correct configuration
- Recommendation to use project `.npmrc` if custom registry is needed

### DIFF
--- a/docs/recipes/ci-configurations/github-actions.md
+++ b/docs/recipes/ci-configurations/github-actions.md
@@ -21,6 +21,27 @@ This is because npm's Trusted Publisher mechanism authorizes the workflow that i
 Before trusted publishing was available, generating provenance attestations required configuring your project to enable publishing with provenance.
 With trusted publishing, npm provenance is automatically generated for packages published to npm from GitHub Actions without any additional configuration.
 
+## Important: Avoid `registry-url` in `setup-node`
+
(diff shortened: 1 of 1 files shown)


travi (approved):
thanks!

PR #4057, opened 2026-03-01
test(integration): push local git notes to "remote" test repo

travi:
to avoid them being pruned by personal config when unshallowing the test repo

### DIFF
--- a/test/helpers/git-utils.js
+++ b/test/helpers/git-utils.js
@@ -319,6 +319,16 @@ export async function gitAddNote(note, ref, execaOptions) {
 }
 
 /**
+ * Push all notes refs to the remote repository.
+ *
+ * @param {String} repositoryUrl The remote repository URL.
+ * @param {Object} [execaOpts] Options to pass to `execa`.
+ */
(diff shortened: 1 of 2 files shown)


travi:
i've had trouble running the `test:integration` suite locally for quite some time, even though that suite has had no trouble running successfully on CI and no one else has reported the suite failing locally for them. this meant there was clearly something different about my local environment causing the problem. 

i've previously tracked problems like this down to differences in ANSI colors that i use locally and have worked around those in some of these suites, but this one took longer for me to track down.

finally tracked it down to my local `~/.gitconfig` being configured to [prune refs that no longer exist on the remote](https://github.com/travi/dotfiles/blob/bca62352c45458ad4e3a41466c5dfe0fd448f3d6/git/.gitconfig). since the tests that were failing trigger an unshallow, that triggers the prune step, removing the notes added locally for the tests because they didnt exist in the remote for the test repo. pushing the tags to the remote of the test repo after creating them avoids the prune behavior from removing the tags since they now also exist on remote.

PR #4068, opened 2026-03-04
fix: get-last-release matches prerelease tags with dots in config

### LINKED ISSUE #4067: get-last-release fails to match prerelease tags when prerelease config value contains dots
### Current behavior

When the `prerelease` config resolves to a value containing dots (e.g., `AS2024.10`), semantic-release fails to find the correct last prerelease tag. It falls back to the base release version and attempts to create a prerelease version that already exists.

**Example scenario:**
> With 15 existing prerelease tags (`v1.705.3-AS2024.10.1` through `v1.705.3-AS2024.10.15`), semantic-release ignores all of them, finds only the base version `1.705.2`, and tries to create `v1.705.3-AS2024.10.1` — which already exists. This causes downstream failures (e.g., Docker registry rejecting duplicate tags).

### Root cause

The `.includes()` check added in [PR #2416](https://github.com/semantic-release/semantic-release/pull/2416) (`lib/get-last-release.js`):

```js
semver.parse(tag.version).prerelease.includes(
  branch.prerelease === true ? branch.name : branch.prerelease
)
```

`semver.parse()` splits the prerelease portion on dots into separate components per the [semver spec](https://semver.org/#spec-item-9):

```js
semver.parse('1.705.3-AS2024.10.15').prerelease
// → ['AS2024', 10, 15]

['AS2024', 10, 15].includes('AS2024.10')
// → false 
```

`'AS2024.10'` is not a single element in the array — it was split into `'AS2024'` and `10`. This means **any** `prerelease` config value containing a dot will silently fail to match its own tags.

### Why this is a bug, not user error

| # | Reason |
|---|--------|
| 1 | **No validation** — semantic-release silently accepts dots in the `prerelease` config. Unlike `/` or `_` which throw `EPRERELEASEBRANCH`, dots produce no error and no warning. There is no documentation explicitly prohibiting them. |
| 2 | **Valid semver output** — The resulting versions (e.g., `1.705.3-AS2024.10.15`) are fully valid per the [semver spec §9](https://semver.org/#spec-item-9). Dots are defined as separators between prerelease identifiers, and `['AS2024', 10, 15]` is a valid prerelease sequence. |
| 3 | **Regression** — The same configuration worked correctly in **v17–v20**, where `get-last-release.js` only checked channel matching without the `.includes()` filter. |

### Questioning the necessity of the `.includes()` check

Prior to [PR #2416](https://github.com/semantic-release/semantic-release/pull/2416), semantic-release relied solely on **git notes channels** to match tags to branches — and that was sufficient:

```js
// v17–v20: Only checked channels 
tag.channels.some((channel) => isSameChannel(branch.channel, channel))
```

The `.includes()` check was added to handle an edge case where multiple prerelease branches share the same commit ([#2415](https://github.com/semantic-release/semantic-release/issues/2415)). However, this check introduces an **implicit requirement** that the `prerelease` config value must map to a **single** semver prerelease component — a constraint that is neither documented nor validated.

### Expected behavior

Either of the following would be a valid fix:

### Option A: Reject dots in the `prerelease` config

Add explicit validation that rejects `.` in the resolved prerelease identifier (similar to how `/` and `_` trigger `EPRERELEASEBRANCH`). This makes the constraint explicit rather than silently producing broken behavior.

### Option B: Fix the matching logic

The `.includes()` check assumes the prerelease config maps to a single semver component. Instead, join the parsed components back and check the prefix:

```diff
- semver.parse(tag.version).prerelease.includes(branch.prerelease)
+ semver.parse(tag.version).prerelease.join('.').startsWith(
+   branch.prerelease === true ? branch.name : branch.prerelease
+ )
```

This preserves the original intent of PR #2416 (preventing cross-contamination between prerelease branches like `alpha` and `beta`) while correctly handling multi-component prerelease identifiers.

22.0.12 (also affects 21.0.6+)

### CI environment

GitHub Actions (ubuntu-22)

### Plugins used

```
@semantic-release/commit-analyzer
@semantic-release/changelog
@semantic-release/release-notes-generator
@semantic-release/npm
@semantic-release/exec
@semantic-release/github
@semantic-release/git
```

> Orchestrated via `@qiwi/multi-semantic-release@7.1.2`, but the bug is in core `semantic-release`.

```yaml
branches:
  - name: "main"
  - name: "alpha"
    prerelease: true
  - name: "beta"
    prerelease: true
  - name: "lts/AS+([0-9])(.+([0-9]))"
    prerelease: '${name.replace(/^lts\//, "")}'
    # For branch lts/AS2024.10, resolves to: AS2024.10
plugins:
  - - "@semantic-release/commit-analyzer"
    - preset: "angular"
  - "@semantic-release/changelog"
  - "@semantic-release/release-notes-generator"
  - - "@semantic-release/npm"
    - npmPublish: false
  - - "@semantic-release/exec"
    - prepareCmd: "cargo make build"
  - - "@semantic-release/exec"
    - prepareCmd: "cargo make package"
  - - "@semantic-release/exec"
    - prepareCmd: "cargo make publish"
  - - "@semantic-release/github"
    - successComment: false
  - - "@semantic-release/git"
    - assets: ["package.json", "package-lock.json", "CHANGELOG.md"]
```

### CI logs

Ran with `DEBUG=semantic-release:*`. Relevant excerpt:

```log
[10:13:27 AM] [semantic-release] › ℹ  Found git tag connector-builder-ui@1.705.2 associated with version 1.705.2 on branch lts/AS2024.10
[10:13:27 AM] [semantic-release] › ℹ  Found 18 commits since last release
[10:13:27 AM] [semantic-release] › ℹ  Start step "analyzeCommits" of plugin "@semantic-release/commit-analyzer"
[10:13:27 AM] [semantic-release] › ℹ  Analyzing 18 commits
[10:13:27 AM] [semantic-release] › ℹ  The release type for the commit is patch
[10:13:27 AM] [semantic-release] › ℹ  Analysis of 18 commits complete: patch release
[10:13:27 AM] [semantic-release] › ℹ  The next release version is 1.705.3-AS2024.10.1
```

> All 15 prerelease tags (`connector-builder-ui@1.705.3-AS2024.10.1` through `.15`) exist and are fetched, yet semantic-release finds only the base version `1.705.2` instead of the latest prerelease `1.705.3-AS2024.10.15`

theluckystrike:
Fixes #4067

The .includes() check fails when prerelease config contains dots (e.g., AS2024.10) because semver.parse() splits the prerelease portion on dots per semver spec. For example: semver.parse('1.705.3-AS2024.10.15').prerelease returns ['AS2024', 10, 15] and ['AS2024', 10, 15].includes('AS2024.10') returns false.

This fix joins the prerelease components and uses startsWith() instead, which correctly handles multi-component prerelease identifiers.

---
*Contributed by [theluckystrike](https://github.com/theluckystrike) | [Zovo](https://zovo.one)*

### DIFF
--- a/lib/get-last-release.js
+++ b/lib/get-last-release.js
@@ -34,7 +34,8 @@ export default ({ branch, options: { tagFormat } }, { before } = {}) => {
           tag.channels.some((channel) => isSameChannel(branch.channel, channel)) &&
           semver
             .parse(tag.version)
-            .prerelease.includes(branch.prerelease === true ? branch.name : branch.prerelease)) ||
+            .prerelease.join(".")
+            .startsWith(branch.prerelease === true ? branch.name : branch.prerelease)) ||
           !semver.prerelease(tag.version)) &&
         (isUndefined(before) || semver.lt(tag.version, before))
     )


theluckystrike:
Closing — the open-issue count on this repo exceeds our threshold. Thank you for maintaining semantic-release!

PR #3732, opened 2025-05-19
perf(get-tags.js): bulk get for tags notes

vojty:
I’ve identified that retrieving tags and their notes is a performance issue, especially in large repositories with hundreds of tags. 

The current implementation calls git notes twice (due to backward compatibility for some older semantic-release versions) for each tag, sequentially. Here is the problematic part https://github.com/semantic-release/semantic-release/blob/master/lib/branches/get-tags.js#L24

I found out that notes for each tag can be obtained in a single command, which takes only a few milliseconds. Here is the command I used (should cover the older notes refs as well):
```sh
git log --tags="*" --decorate-refs="refs/tags/*" --no-walk --format="%d%x09%N" --notes="refs/notes/semantic-release*"
```

And the example output for this repository:
```
 (tag: v24.2.5)	
 (tag: v24.2.4)	
 (tag: v24.2.3)	{"channels":[null]}

 (tag: v24.2.2)	{"channels":[null]}

 (tag: v24.2.1)	{"channels":[null]}

 (tag: v24.2.0)	{"channels":[null]}
```

The retrieved tags and their notes are fetched all at once (takes only a few ms) and then parsed and converted to a map.

This change shaves off a few minutes from our CI job.

Related https://github.com/semantic-release/semantic-release/issues/2827, https://github.com/semantic-release/semantic-release/pull/3689, https://github.com/semantic-release/semantic-release/issues/3594

PR #2085, opened 2021-08-23
fix: use one note reference per tag to prevent conflicts

KillianHmyd:
## Current behavior

All notes are pushed under the reference `semantic-release` because of that, if from two different branches and channels a release is performed at the same time the release would fail on one of them with this error : 
```
 ! [rejected]        refs/notes/semantic-release -> refs/notes/semantic-release (fetch first)
```

## New behavior

Instead of using only one note reference we would use one per tag, this way we could not have any conflict between multiple channels.
This change should be retro-compatible and doesn't need a migration.

resolves #1613

PR #4074, opened 2026-03-06
fix: handle multiple notes refs on same commit in getTagsNotes

### LINKED ISSUE #4073: getTagsNotes fails when multiple tags on same commit have different notes refs
## Description

When two tags point to the same commit and each has a git note under a different `refs/notes/semantic-release-*` ref, `getTagsNotes()` produces unparseable JSON, causing semantic-release to silently ignore the tag's channel information.

### Setup
- Branch config: `staging` (prerelease: `beta`) and `develop` (prerelease: `alpha`)
- A commit is released on **both** branches (e.g., via merge or backmerge), producing two tags on the same commit:
  - `v2.9.0-beta.13` → note in `refs/notes/semantic-release-v2.9.0-beta.13`: `{"channels":["staging"]}`
  - `v2.9.0-alpha.1` → note in `refs/notes/semantic-release-v2.9.0-alpha.1`: `{"channels":["develop"]}`

### What happens

`getTagsNotes()` in `lib/git.js` runs:

```
git log --tags="*" --decorate-refs="refs/tags/*" --no-walk --format="%d%x09%N" --notes="refs/notes/semantic-release*"
```

When multiple `--notes` refs match for the same commit, git **concatenates** the notes with newlines in `%N`:

```
(tag: v2.9.0-beta.13, tag: v2.9.0-alpha.1)	{"channels":["develop"]}
{"channels":["staging"]}
```

The tab split gives `notePart = '{"channels":["develop"]}\n{"channels":["staging"]}'`, which is **not valid JSON**. `JSON.parse()` throws, the catch block silently drops the error, and both tags get no entry in the map → default to `channels: [null]`.

### Result

- The prerelease tag (e.g., `v2.9.0-beta.13`) is invisible to its channel because `isSameChannel('staging', null)` returns `false`
- semantic-release falls back to an older tag (e.g., `beta.12`) and tries to recreate `beta.13`
- Fails with: `fatal: tag 'v2.9.0-beta.13' already exists`
- **The release is permanently stuck** — retrying doesn't help since the tag conflict persists

## Root cause

`getTagsNotes()` assumes `%N` produces a single-line JSON value per commit, but git concatenates notes from multiple matching refs with newlines.

## Suggested fix

In `getTagsNotes()`, split `notePart` by newlines and parse each line as a separate JSON object, merging the `channels` arrays:

```js
// Instead of:
const parsed = JSON.parse(notePart);

// Do:
const noteLines = notePart.split("\n").filter(Boolean);
const mergedChannels = new Set();
for (const line of noteLines) {
  const parsed = JSON.parse(line.trim());
  if (parsed.channels) {
    parsed.channels.forEach((ch) => mergedChannels.add(ch));
  }
}
const parsed = { channels: [...mergedChannels] };
```

This ensures that when multiple notes refs exist for the same commit, all channels are preserved.

## Environment

- semantic-release version: 25.0.3
- Git version: 2.47.1
- CI: GitLab CI

## Workaround

Manually remove one of the conflicting git notes:
```bash
git fetch origin "+refs/notes/*:refs/notes/*"
git notes --ref "semantic-release-<conflicting-version>" remove <commit-sha>
git push origin refs/notes/semantic-release-<conflicting-version>
```

jean-humann:
## Summary

Fixes #4073

When two tags point to the same commit (e.g., `v2.9.0-beta.13` and `v2.9.0-alpha.1` from releases on different prerelease branches), each release creates a git note under a separate `refs/notes/semantic-release-*` ref. The `getTagsNotes()` function uses `git log --notes="refs/notes/semantic-release*"` which causes git to **concatenate** notes from multiple matching refs with newlines in `%N`.

### Before this fix

The concatenated output produces lines like:
```
(tag: v2.9.0-beta.13, tag: v2.9.0-alpha.1)	{"channels":["develop"]}
{"channels":["staging"]}
```

- Line 1 is parsed correctly, but assigns `channels: ["develop"]` to **both** tags (using whichever notes ref comes first alphabetically)
- Line 2 has no tag decoration, so `extractGitLogTags()` returns `[]` and the line is silently dropped

Result: `v2.9.0-beta.13` gets `channels: ["develop"]` instead of `["staging"]`, making it invisible to the staging branch. semantic-release falls back to an older tag and tries to recreate the existing one → `fatal: tag already exists`.

### After this fix

Continuation lines (without tag decorations) are detected as additional notes from other `refs/notes/semantic-release-*` refs on the same commit. Their channels are merged with the preceding tags' entries:

```js
// Both tags now get: { channels: ["develop", "staging"] }
```

This ensures tags are visible to all channels they were released on.

## Changes

- **`lib/git.js`**: Modified `getTagsNotes()` to track the last processed tags and merge channels from continuation note lines
- **`test/git.test.js`**: Added test case that creates two tags on the same commit with different notes refs and verifies channels are merged

## Test plan

- [x] New test: "Merge notes when multiple tags on same commit have different notes refs"
- [x] All 33 existing git tests pass
- [x] Verified fix resolves the exact scenario from #4073

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/git.js
+++ b/lib/git.js
@@ -334,21 +334,69 @@ export async function getTagsNotes(execaOptions) {
   // parse and create a map of tags to notes
   const tagNotesMap = new Map();
 
-  for (const line of lines) {
-    const [tagPart, notePart] = line.trim().split("\t"); // tab separator is defined in the git command above (%x09)
-    const tags = extractGitLogTags(tagPart);
-    if (tags.length === 0) {
-      debug(`Cannot parse tags from line: ${line}`);
-      continue;
-    }
+  // A single commit can carry more than one `semantic-release` notes ref - for example a legacy
+  // `refs/notes/semantic-release` note written before v23 (#2085) alongside the current per-tag
+  // `refs/notes/semantic-release-<tag>` note, or several tags that point at the same commit.
+  // `git log` concatenates every matching ref into %N (newline-separated), so a commit's notes
+  // can span several lines with nothing indicating which note came from which ref.
+  //
+  // When a commit has a single tag we merge those lines (e.g. the legacy and per-tag notes of one
+  // tag, as left behind by upgrading across the v23 change), unioning channels so none are
+  // dropped. When several tags share a commit the concatenation is ambiguous, so we instead read
+  // each tag's note from its own ref to attribute the channels to the correct tag.
+  let lastTag;
 
-    try {
-      const parsed = JSON.parse(notePart);
-      tags.forEach((tag) => {
-        tagNotesMap.set(tag, parsed);
-      });
-    } catch (error) {
-      debug(error);
+  for (const line of lines) {
+    const [tagPart, ...rest] = line.trim().split("\t"); // tab separator is defined in the git command above (%x09)
+    const notePart = rest.join("\t");
+    const tags = extractGitLogTags(tagPart);
+
+    if (tags.length === 1) {
+      const [tag] = tags;
+      lastTag = tag;
+      if (notePart) {
+        try {
+          tagNotesMap.set(tag, JSON.parse(notePart));
+        } catch (error) {
+          debug(error);
+        }
+      }
+    } else if (tags.length > 1) {
+      // Read each tag's note from its dedicated ref, as the concatenated %N cannot be attributed.
+      lastTag = undefined;
+      for (const tag of tags) {
+        try {
+          const { stdout: note } = await execa(
+            "git",
+            ["notes", "--ref", `${GIT_NOTE_REF}-${tag}`, "show", tag],
+            execaOptions
+          );
+          tagNotesMap.set(tag, JSON.parse(note.trim()));
+        } catch (error) {
+          debug(error);
+        }
+      }
+    } else if (lastTag && line.trim()) {
+      // Continuation line: an additional notes ref on the same single-tag commit.
+      // Merge it into the existing entry, unioning the `channels` so none are dropped.
+      try {
+        const parsed = JSON.parse(line.trim());
+        const existing = tagNotesMap.get(lastTag);
+        if (existing) {
+          const merged = { ...existing, ...parsed };
+          if (existing.channels || parsed.channels) {
+            merged.channels = [...new Set([...(existing.channels ?? []), ...(parsed.channels ?? [])])];
+          }
+
+          tagNotesMap.set(lastTag, merged);
+        } else {
+          tagNotesMap.set(lastTag, parsed);
+        }
+      } catch (error) {
+        debug(error);
+      }
+    } else {
+      debug(`Cannot parse tags from line: ${line}`);
     }
   }
 
--- a/lib/git.js
+++ b/lib/git.js
@@ -340,21 +340,69 @@ export async function getTagsNotes(execaOptions) {
   // parse and create a map of tags to notes
   const tagNotesMap = new Map();
 
-  for (const line of lines) {
-    const [tagPart, notePart] = line.trim().split("\t"); // tab separator is defined in the git command above (%x09)
-    const tags = extractGitLogTags(tagPart);
-    if (tags.length === 0) {
-      debug(`Cannot parse tags from line: ${line}`);
-      continue;
-    }
+  // A single commit can carry more than one `semantic-release` notes ref - for example a legacy
+  // `refs/notes/semantic-release` note written before v23 (#2085) alongside the current per-tag
+  // `refs/notes/semantic-release-<tag>` note, or several tags that point at the same commit.
+  // `git log` concatenates every matching ref into %N (newline-separated), so a commit's notes
+  // can span several lines with nothing indicating which note came from which ref.
+  //
+  // When a commit has a single tag we merge those lines (e.g. the legacy and per-tag notes of one
+  // tag, as left behind by upgrading across the v23 change), unioning channels so none are
+  // dropped. When several tags share a commit the concatenation is ambiguous, so we instead read
+  // each tag's note from its own ref to attribute the channels to the correct tag.
+  let lastTag;
 
-    try {
-      const parsed = JSON.parse(notePart);
-      tags.forEach((tag) => {
-        tagNotesMap.set(tag, parsed);
-      });
-    } catch (error) {
-      debug(error);
+  for (const line of lines) {
+    const [tagPart, ...rest] = line.trim().split("\t"); // tab separator is defined in the git command above (%x09)
+    const notePart = rest.join("\t");
+    const tags = extractGitLogTags(tagPart);
+
+    if (tags.length === 1) {
+      const [tag] = tags;
+      lastTag = tag;
+      if (notePart) {
+        try {
+          tagNotesMap.set(tag, JSON.parse(notePart));
+        } catch (error) {
+          debug(error);
+        }
+      }
+    } else if (tags.length > 1) {
+      // Read each tag's note from its dedicated ref, as the concatenated %N cannot be attributed.
+      lastTag = undefined;
+      for (const tag of tags) {
+        try {
+          const { stdout: note } = await execa(
+            "git",
+            ["notes", "--ref", `${GIT_NOTE_REF}-${tag}`, "show", tag],
+            execaOptions
+          );
+          tagNotesMap.set(tag, JSON.parse(note.trim()));
+        } catch (error) {
+          debug(error);
+        }
+      }
+    } else if (lastTag && line.trim()) {
+      // Continuation line: an additional notes ref on the same single-tag commit.
+      // Merge it into the existing entry, unioning the `channels` so none are dropped.
+      try {
+        const parsed = JSON.parse(line.trim());
+        const existing = tagNotesMap.get(lastTag);
+        if (existing) {
+          const merged = { ...existing, ...parsed };
+          if (existing.channels || parsed.channels) {
+            merged.channels = [...new Set([...(existing.channels ?? []), ...(parsed.channels ?? [])])];
+          }
+
+          tagNotesMap.set(lastTag, merged);
+        } else {
+          tagNotesMap.set(lastTag, parsed);
+        }
+      } catch (error) {
+        debug(error);
+      }
+    } else {
+      debug(`Cannot parse tags from line: ${line}`);
     }
   }
 
--- a/test/git.test.js
+++ b/test/git.test.js
@@ -19,6 +19,7 @@ import {
 } from "../lib/git.js";
 import {
   gitAddConfig,
+  gitAddLegacyNote,
   gitAddNote,
   gitCheckout,
   gitCommits,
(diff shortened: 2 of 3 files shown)


jean-humann:
@travi Would you be able to review this? It fixes a bug in getTagsNotes() where multiple notes refs on the same commit cause JSON parsing to fail, breaking prerelease channels. I noticed you recently worked on the notes-related tests (#4057), so this is right in that area.

travi:
your PR has been on my radar and i agree that this needs attention to likely handle this better. so far, our available time as volunteer maintainers has been dedicated to other things, so this hasnt risen to the point where we've been able to give it priority yet. however, the workflow you describe highlights that you are following gitflow, which is officially unsupported by semantic-release. this makes it even harder for us to consider this a priority. it could potentially help if you could show us how this shows up in supported workflows.

> breaking prerelease channels

prerelease channels are not broken in general, when used in supported ways. 

our prerelease workflow is not provided as a testing workflow. semantic-release specifically does not provide a testing workflow and likely never will because it is expected that you are verifying the behaviors of your package _before_ release and dont need to consume it to verify it is stable enough to release. instead of existing to confirm that your package behaves as intended, the prerelease workflow exists for scenarios where you are trying to define what the behavior should be. things like confirming that the breaking interface works for consumers so that it is not more breaking than intended, or that a new interface is ergonomic enough for consumers to accomplish the intended goal. these situations should be rare and infrequent. when promoting a prerelease to be stable, the prerelease branch is intended to be deleted and not exist until the next rare situation where the branch would be created fresh from trunk again

if you can help us see where the problem you describe shows up in supported situations, that would help us make it a higher priority

alisson276:
I have the same fix on my fork. I have used `%x00` to better organize the output. Here is my commit: https://github.com/semantic-release/semantic-release/compare/master...alisson276:semantic-release:master

I opened a PR a long time ago, but I don't know why it was not merged.

rQcAnE4pk15VDHVa:
@travi 
My team is also going for a gitflowy process and this PR would resolve a big pain. Whether gitflow is supported or not:
I see the ignored channels after linebreak as a kind of bug. It just feels like it output has been overlooked.
I would so much appreciate it, if you could look into this PR.

I can confirm: The code in the PR detects the linebreak'ed notes correctly:
``` 
semantic-release:get-tags found tags for branch develop: [ { gitTag: 'v1.0.0', version: '1.0.0', channels: [ null, 'staging', 'develop' ] }, ...
```

Previous:
``` 
semantic-release:get-tags found tags for branch develop: [ { gitTag: 'v1.0.0', version: '1.0.0', channels: [ null ] }
```

travi:
> Whether gitflow is supported or not:
> I see the ignored channels after linebreak as a kind of bug.

heres the thing. we dont disagree with this. we think this is likely a bug as well. however, supported workflows is one of the decision criteria we have to use to decide where we invest our time. we are volunteer maintainers with limited time and attention. we appreciate the effort that has gone into identifying and suggesting a fix for this, but we havent been able to give it priority yet, largely because it mostly, if not exclusively, is needed for workflows that are not officially supported. 

> if you can help us see where the problem you describe shows up in supported situations, that would help us make it a higher priority

i stated this above and it continues to be true. helping us see where this applies to officially supported workflows is the best way for us to give more priority to investigating further

jean-humann:
Thanks @travi — that's fair, and you're right the original report leaned on a gitflow setup. I went back to find where this surfaces in a **supported** workflow, since that's the deciding factor.

**Supported repro: upgrading across the v23 notes-ref change + distribution channels.** In #2085 (v23) notes moved from the single `refs/notes/semantic-release` ref to per-tag `refs/notes/semantic-release-<tag>` refs, and #3732 switched `getTagsNotes` to one bulk `git log --notes=refs/notes/semantic-release*`. So for any repo first released before v23 and still released after it, a tag can legitimately carry **both** its legacy note and its new per-tag note, and `git log` concatenates them into `%N`:

```
 (tag: v1.0.0)	{"channels":[null]}
{"channels":[null,"next"]}
```

The current parser `JSON.parse`s only the first line, so the second note is dropped and the tag's channels collapse to `[null]`:

```
master:  v1.0.0 => {"channels":[null]}        # "next" silently dropped
fixed:   v1.0.0 => {"channels":[null,"next"]} # correct
```

Minimal repro (no gitflow — just the legacy + per-tag refs the upgrade leaves behind):

```bash
git notes --ref semantic-release        add -m '{"channels":[null]}'        v1.0.0
git notes --ref semantic-release-v1.0.0 add -m '{"channels":[null,"next"]}' v1.0.0
git log --tags='*' --decorate-refs='refs/tags/*' --no-walk --format='%d%x09%N' --notes='refs/notes/semantic-release*'
```

To be precise about scope: I'm not claiming channels or prereleases are broken in general — they work as documented. It's specifically this parse path mishandling the documented case where one commit carries notes from more than one `semantic-release-*` ref — exactly the back-compat the `--notes` glob in #3732 was added for.

I've reworked the PR accordingly: it merges notes only when the commit has a **single tag** (the supported case), unioning channels; when several distinct tags share a commit, `%N` gives no way to attribute each note to a tag, so it deliberately doesn't guess (that only comes up in unsupported workflows). Added a test reproducing the v23 upgrade scenario plus a guard against cross-tag contamination. Rebased on master, prettier-clean, tests green.

@alisson276 landed the same fix independently — happy to converge on whichever you'd prefer. Totally understand the volunteer-time constraints; I've tried to keep this cheap to review. Thanks for taking another look.

jean-humann:
I spent some time tracing the root cause, and I think this might be a correctness regression from #3732 rather than a workflow-specific issue — but I'd value your read before going further, since you co-authored it.

Before #3732, `get-tags` read each tag's note with `getNote(tag)`, which read **both** of that tag's refs — the legacy `refs/notes/semantic-release` note (kept, per its own comment, "for retro-compatibility" after #2085 moved to one ref per tag) and the per-tag `refs/notes/semantic-release-<tag>` note — and `merge`d them, so channels were read from the right ref and combined correctly.

#3732 replaced that with a single bulk `git log --notes=refs/notes/semantic-release*` for performance. Because `%N` concatenates the notes from every matching ref with no indication of which came from which, `getTagsNotes` parses only the first line and drops the rest. As far as I can tell that lost two behaviours `getNote` had:

1. the legacy + per-tag **merge** — so a tag released before the v23 ref change loses channels after an upgrade (the single-tag case in my earlier comment); and
2. correct **per-tag attribution** — when a commit carries more than one tag, a tag can be handed another tag's channels, or none.

Same repo, old vs current logic:

```
# commit with two tags, each with its own per-tag note
getNote (pre-#3732):  beta.13 => {channels:["staging"]}   alpha.1 => {channels:["develop"]}   correct
getTagsNotes (today): beta.13 => {channels:["develop"]}   alpha.1 => {channels:["develop"]}   wrong
```

No test covered the multiple-refs case, so it seems to have gone unnoticed.

Does that match your understanding? If so, my thought was to restore the old correctness while keeping #3732's performance for the common case: keep the single bulk `git log`, merge the legacy + per-tag notes for single-tag commits, and only fall back to reading each tag's own ref when a commit has multiple tags — so repos without multi-tag commits still do exactly one `git log`, unchanged. I've drafted that on the branch, but I'm very happy to adjust the shape — e.g. scope it strictly to the single-tag/upgrade case — depending on how you'd prefer to handle it.

Blackclaws:
> > Whether gitflow is supported or not:
> > I see the ignored channels after linebreak as a kind of bug.
> 
> heres the thing. we dont disagree with this. we think this is likely a bug as well. however, supported workflows is one of the decision criteria we have to use to decide where we invest our time. we are volunteer maintainers with limited time and attention. we appreciate the effort that has gone into identifying and suggesting a fix for this, but we havent been able to give it priority yet, largely because it mostly, if not exclusively, is needed for workflows that are not officially supported.
> 
> > if you can help us see where the problem you describe shows up in supported situations, that would help us make it a higher priority
> 
> i stated this above and it continues to be true. helping us see where this applies to officially supported workflows is the best way for us to give more priority to investigating further

I'm not sure about supported or unsupported workflows. We've been using the same workflow forever which is that we only do fast forward merges from dev -> staging -> main. Now this breaks suddenly and there is talk about this being an unsupported workflow for semantic-release? Since when?

jean-humann:
Hi @travi — I’ve rebased this PR onto the current `master` (v25.0.9) and revalidated the change locally:

- focused Git test suite: 36 passing
- full unit suite: 280 passing
- Prettier check: passing

The implementation keeps the bulk-read path for the common single-tag case and falls back to each tag’s dedicated notes ref only when multiple tags share a commit. The regression coverage includes both the legacy + per-tag notes upgrade case and correct attribution for shared-commit tags.

Would you be able to take another look when you have a chance? Thanks.

travi:
Please stop tagging maintainers directly about this. How we are prioritizing this has already been explained. You are welcome to continue commenting on this thread, but tagging us will not change our priority

Moumouls:
My team encountered the issue also, we applied a fix using "yarn patch"

here the complet fix applied, with some performance improvement.


```patch
diff --git a/lib/git.js b/lib/git.js
index 8766eda168ff9b4bb6edf5acf71f4ce4077b62a7..01c7b914a970c28cb6baf1ebce110a353593e49e 100644
--- a/lib/git.js
+++ b/lib/git.js
@@ -314,7 +314,7 @@ export async function isBranchUpToDate(repositoryUrl, branch, execaOptions) {
  */
 export async function getTagsNotes(execaOptions) {
   /**
-   * git log --tags="*" --decorate-refs="refs/tags/*" --no-walk --format="%d%x09%N" --notes="refs/notes/semantic-release*"
+   * git log --tags="*" --decorate-refs="refs/tags/*" --no-walk --format="%H%x09%d%x09%N" --notes="refs/notes/semantic-release*"
    *
    *  (tag: v1.2.3)
    *  (tag: v2.0.0)  {"channels":[null]}
@@ -328,7 +328,7 @@ export async function getTagsNotes(execaOptions) {
       "--tags=*",
       "--decorate-refs=refs/tags/*", // This filters the refs shown in the %d format specifier to only include tags matching refs/tags/*.
       "--no-walk", // This ensures that only the commits directly pointed to by the tags are shown, not their historical parents.
-      "--format=%d%x09%N", // <refName><tab><notes> eg. (tag: v24.2.3)  {"channels":[null]}
+      "--format=%H%x09%d%x09%N", // <sha><tab><refName><tab><notes>
       `--notes=refs/notes/${GIT_NOTE_REF}*`, // handles both patterns for notes - `semantic-release` (old) and `semantic-release-<version>` (current)
     ],
     execaOptions
@@ -339,15 +339,20 @@ export async function getTagsNotes(execaOptions) {
 
   // parse and create a map of tags to notes
   const tagNotesMap = new Map();
+  const sharedCommitTags = [];
 
   for (const line of lines) {
-    const [tagPart, notePart] = line.trim().split("\t"); // tab separator is defined in the git command above (%x09)
+    const [hash, tagPart, notePart] = line.trim().split("\t"); // tab separator is defined in the git command above (%x09)
     const tags = extractGitLogTags(tagPart);
     if (tags.length === 0) {
       debug(`Cannot parse tags from line: ${line}`);
       continue;
     }
 
+    if (tags.length > 1 && hash) {
+      tags.forEach((tag) => sharedCommitTags.push({ tag, hash }));
+    }
+
     try {
       const parsed = JSON.parse(notePart);
       tags.forEach((tag) => {
@@ -358,9 +363,55 @@ export async function getTagsNotes(execaOptions) {
     }
   }
 
+  // Per-tag notes must win over the commit-level %N when several tags share a
+  // SHA. One `git cat-file --batch` avoids thousands of `git notes show` spawns.
+  if (sharedCommitTags.length > 0) {
+    const { stdout: batchStdout } = await execa("git", ["cat-file", "--batch"], {
+      ...execaOptions,
+      input: `${sharedCommitTags.map(({ tag, hash }) => `refs/notes/${GIT_NOTE_REF}-${tag}:${hash}`).join("\n")}\n`,
+    });
+    applyPerTagNotesFromCatFileBatch(batchStdout, sharedCommitTags, tagNotesMap);
+  }
+
   return tagNotesMap;
 }
 
+function applyPerTagNotesFromCatFileBatch(stdout, sharedCommitTags, tagNotesMap) {
+  let pos = 0;
+  let i = 0;
+  while (i < sharedCommitTags.length && pos < stdout.length) {
+    const nl = stdout.indexOf("\n", pos);
+    if (nl === -1) {
+      break;
+    }
+
+    const header = stdout.slice(pos, nl);
+    pos = nl + 1;
+    if (header.endsWith(" missing")) {
+      i += 1;
+      continue;
+    }
+
+    const size = Number(header.split(" ")[2]);
+    if (!Number.isFinite(size) || size < 0) {
+      break;
+    }
+
+    const content = stdout.slice(pos, pos + size);
+    pos += size;
+    if (stdout[pos] === "\n") {
+      pos += 1;
+    }
+
+    try {
+      tagNotesMap.set(sharedCommitTags[i].tag, JSON.parse(content));
+    } catch (error) {
+      debug(error);
+    }
+    i += 1;
+  }
+}
+
 /**
  * Add JSON note to a given reference.
  *
  ```

PR #4149, opened 2026-05-03
fix: anchor tagFormat regex with end-of-string marker (#4138)

### LINKED ISSUE #4138: tagFormat collisions
### Current behavior

Either a release doesn't occur or occurs with an incorrect version (as in the linked case here). In this case, the release should have been tagged `v1.0.0-azurekeycloaktester-container-longer` as no prior release with that exact tag existed but it was instead tagged as [v2.0.0-azurekeycloaktester-container-longer](https://github.com/SwanseaUniversityMedical/workflows-test/releases/tag/v2.0.0-azurekeycloaktester-container-longer). This occurred because semantic release picked up the existence of the slightly longer `v1.1.0-azurekeycloaktester-container-longer-tag` tag, took into account a major bump was requested then based it's output off that version number rather than matching the exact but shorter string.

In summary, a shorter tag format that is a prefix of a longer tag can consume the longer tag and get an incorrect version number or report that the tag already exists, thereby erroneously halting the release process.

All this occurs because the tagRegexp is not end anchored: https://github.com/semantic-release/semantic-release/blob/afb5b7afb75a2a2305e9f3ee6925d11a4f2212cf/lib/branches/get-tags.js#L14 and needs a $ at the end

More example input tagFormat values:
- v${version}-sonar-dotnet-10
- v${version}-sonar-dotnet-10-action

How semantic-release currently builds regex:

1. It renders ${version} as a space first.

2. It escapes regex chars.

3. It replaces that space with (.+).

4. It prefixes with ^ but does not suffix with $.

So you get:

v${version}-sonar-dotnet-10 → ^v(.+)-sonar-dotnet-10

v${version}-sonar-dotnet-10-action → ^v(.+)-sonar-dotnet-10-action

What those match

For regex ^v(.+)-sonar-dotnet-10:

- Matches v1.0.0-sonar-dotnet-10

- Also matches v1.0.0-sonar-dotnet-10-action ← unwanted partial match

The captured version in both cases: 1.0.0

For regex ^v(.+)-sonar-dotnet-10-action:

- Matches v1.0.0-sonar-dotnet-10-action

- Does not match v1.0.0-sonar-dotnet-10

That is why collisions are one-way: the shorter pattern can consume the longer tag, because there is no end anchor.

If it were end-anchored, it would be:

^v(.+)-sonar-dotnet-10$

^v(.+)-sonar-dotnet-10-action$

### Expected behavior

No collisions should occur - tags like `v1.0.0-sonar-dotnet-10` shouldn't collide with `v1.0.0-sonar-dotnet-10-action`

25.0.3

### CI environment

GitHub Actions

### Plugins used

exec
- verifyConditions
- analyzeCommits
- verifyRelease
- generateNotes
- prepare
- publish
- addChannel
- success
- fail

https://github.com/SwanseaUniversityMedical/workflows/blob/a6a860a0597c55e8d7336fa268a98fd2d2537a2f/.github/actions/determine-bump/action.yaml#L94

### CI logs

https://github.com/SwanseaUniversityMedical/workflows-test/actions/runs/24657930402/job/72098278307

wahajahmed010:
## Description

Fixes #4138

The `tagRegexp` in `lib/branches/get-tags.js` was not end-anchored with `$`, causing shorter `tagFormat` patterns to incorrectly match longer tags that share the same prefix.

### Problem Example

Given these two `tagFormat` values:
- `v${{version}}-sonar-dotnet-10`
- `v${{version}}-sonar-dotnet-10-action`

The shorter regex `^v(.+)-sonar-dotnet-10` would incorrectly match the longer tag `v1.0.0-sonar-dotnet-10-action`, producing an incorrect version number or causing releases to be halted.

### Solution

Add `$` (end-of-string anchor) to the tagRegexp pattern, ensuring exact tagFormat matching.

**Before:**
```js
const tagRegexp = `^${escapeRegExp(template(tagFormat)({ version: " " })).replace(" ", "(.+)")}`;
```

**After:**
```js
const tagRegexp = `^${escapeRegExp(template(tagFormat)({ version: " " })).replace(" ", "(.+)")}$`;
```

### Testing

- [x] Existing tests continue to pass (249 unit + 35 integration)
- [x] Manually verified the regex now correctly rejects partial matches

*This contribution was AI-assisted.*


babblebey:
Could you kindly add tests to this!?

And my only curious ask is why this was the case originally.... @semantic-release/maintainers what do you think about this?

gr2m:
I don't know. If it was intentional then our tests should cover this case. Can you check what our documentation says?

PR #4189, opened 2026-05-30
chore(deps): remove `aggregate-error` polyfill

gameroman:
Removes `aggregate-error` dependency

`AggregateError` is available natively since Node 15.0.0


gameroman:
Superseded by #4292

PR #4221, opened 2026-06-28
test(e2e): migrate from raw use of dockerode to use testcontainers instead

travi:
which uses dockerode internally. this is a follow up to https://github.com/semantic-release/semantic-release/pull/4220 to simplify teardown and avoid managing the complexity as directly


test/helpers/e2e.compose.yml
travi, line 1: in addition to the overall simplification that testcontainers enables, pulling the images out to the compose file here should enable renovate to pin the versions and keep them up to date for us

gr2m (approved):
nice!

PR #4256, opened 2026-07-17
feat: integrate @semantic-release/core engine

babblebey:
This pull request refactors the core logic of the package to use the `@semantic-release/core` engine directly, removes legacy and duplicated code, and centralizes configuration. The changes simplify the codebase by delegating most of the release logic to the upstream core package, making the CLI wrapper thinner and easier to maintain. Several internal utility files for branch and tag handling are also removed, as their logic is now handled by `@semantic-release/core`.

### Migration to Core Engine

* The CLI now delegates release logic to `@semantic-release/core`, using its `resolveConfig`, `getLogger`, and other helpers, instead of maintaining a custom implementation. This reduces maintenance overhead and ensures alignment with the core release logic.
* The documentation is updated to explain that `semantic-release` uses `@semantic-release/core` as its engine, and provides guidance for advanced use cases.

### Configuration and Defaults

* A new `SEMANTIC_RELEASE_DEFAULT_CONFIG` constant is introduced, centralizing the default plugin stack for releases.

### Codebase Simplification

* Removes internal implementations for branch expansion, tag retrieval, and normalization, as well as related validation and error handling logic. These are now handled by the core engine.


travi:
i know this should, in theory, be non-breaking, but since we are already likely to have a beta for the next major soon, i'm wondering if this should target that release line instead. i even wonder if this should be an alpha release line since there might be some open decisions that we might not be ready to include in the beta line. thoughts?

babblebey:
I get the `beta` for sure, and I just changed base.

I'm not sure I'm down the complication of the alpha channel to be amongst our worries next 😆, I'd prefer we stick to beta

travi:
the reason i'm thinking about the potential of alpha is the need to start putting together a beta to handle the recent concentional-changelog releases. as we work through open decisions around the core extraction, i think we should consider whether we want that to block the ability to move forward with the more known changes

babblebey:
I get it now.... I'll appropriate... I was thinking `alpha` with `next` in mind... Considering the alpha does prereleases... then by all means YEA

PR #4269, opened 2026-07-29
chore(ci): update Node.js engine version and test matrix

babblebey:
This pull request updates the supported Node.js versions for both development and CI workflows to ensure compatibility with more recent releases.

**Node.js version updates:**

* Updated the Node.js versions in the GitHub Actions workflow (`.github/workflows/test.yml`) to use `22.22.2` and `24.15` instead of the older `22.14.0` and `24.10.0`.
* Updated the `engines` field in `package.json` to require Node.js `^22.22.2` or `>=24.15`, aligning local development requirements with CI.

BREAKING CHANGE: Now require Node.js `^22.22.2` or `>=24.15`


travi:
This is a breaking change, so we should coordinate it with the other breaking changes targeting the beta channel

(It also needs the breaking change trailer in the commit message, at least for the change to the package.json)

gr2m (approved):

PR #4287, opened 2026-08-07
chore(deps): update @semantic-release/commit-analyzer to v14.0.0-beta.4

babblebey:
This pull request updates the version of a dependency in the `package.json` file to ensure compatibility with the latest features and bug fixes.

Dependency updates:

* Upgraded `@semantic-release/commit-analyzer` from version `^13.0.1` to `^14.0.0-beta.4` to incorporate the latest improvements and potential breaking changes from the new major version.


gr2m (approved):

README.md
```diff
@@ -11,7 +11,7 @@
     <img alt="OpenSSF Scorecard" src="https://api.securityscorecards.dev/projects/github.com/semantic-release/semantic-release/badge">
   </a>
   <a href="#badge">
-    <img alt="semantic-release: angular" src="https://img.shields.io/badge/semantic--release-angular-e10079?logo=semantic-release">
```
travi, line 14: this isnt true anymore, correct?

babblebey: Not right away.

We'll have to release a 14.0.0-beta.4 (which reverts the default in commit-analyzer) to make that the case.

README.md
```diff
@@ -147,10 +147,10 @@ In order to use **semantic-release** you need:
 
 Let people know that your package is published using **semantic-release** and which [commit-convention](#commit-message-format) is followed by including this badge in your readme.
 
-[![semantic-release: angular](https://img.shields.io/badge/semantic--release-angular-e10079?logo=semantic-release)](https://github.com/semantic-release/semantic-release)
+[![semantic-release: COMMIT-CONVENTION](https://img.shields.io/badge/semantic--release-COMMIT--CONVENTION-e10079?logo=semantic-release)](https://github.com/semantic-release/semantic-release)
```
travi, line 150: i understand the goal here, but i'm not sure this remains clear enough that "COMMIT-CONVENTION" should be replaced with the convention they've chosen. thats why it was previously aligned with the current default. we could either leave it aligned with the default, or i think we need to make it far more clear that this shouldnt just be used as-is

travi (approved):
approving, but consider my comment about the badge before merging. if nothing else, maybe the badge part of this should be it's own pr so that can be discussed separately from the actual upgrade?

README.md
babblebey: I suppose the text right above it states that clearly

> Let people know that your package is published using **semantic-release** and which [commit-convention](#commit-message-format) is followed by including this badge in your readme.

🤔

travi: fair enough. i still question whether folks will pay enough attention, but nothing is broken if they overlook it 👍🏼

travi: fair enough. i still question whether folks will pay enough attention, but nothing is broken if they overlook it 👍🏼

